PR #3, opened 2022-06-13
[containers/*] Create a new workflow for verification

fmulero:

**Description of the change**

Add new github workflow using new Verification Engine actions for containers

**Benefits**
* Better code quality
* Reduce risk of including breaking changes
* Shorter review time

**Possible drawbacks**
N/A

**Applicable issues**
N/A

**Additional information**


.github/workflows/verify.yaml
```diff
@@ -1,11 +1,6 @@
 name: 'VIB'
```
recena, line 1: Why did you change the name? We should follow the same approach as [here](https://github.com/bitnami/charts/tree/master/.github/workflows):

- `.github/workflows/vib-verify.yaml`: GitHub Pipeline to integrate with Content Platform and perform a verification.
- `.github/workflows/vib-publish.yaml`: GitHub Pipeline to integrate with Content Platform and perform a publishing.

fmulero: There was no reason to change it. Good catch

.vib/vib-verify.json
```diff
@@ -0,0 +1,28 @@
+{
```
recena, line 1: @fmulero Could you review the indentation of this file? 2-spaces indentation.

fmulero: It should be fixed now

fmulero:
Things that can be improved and known TODOs:
* Container tags are already set to `latest`, we should retrieve the real version in some way
* In the `vib-verify` jobs we are checking out the repository (around 200Mb) for every "flavor".  We only need the vib-verify.json file to run the container-lint and container-package action, so we could reduce the execution time downloading only that file and not the whole repository. We agreed to keep it simple at this point because we could need more resources from the repository in a near future

recena (approved):

.github/workflows/vib-verify.yaml
```diff
@@ -0,0 +1,88 @@
...
+          fi
+      # Using actions/github-scripts because using exit 1 in the script above would not provide any output
+      # Source: https://github.community/t/no-output-on-process-completed-with-exit-code-1/123821/3
+      - id: show-error
+        name: Show error
+        if: ${{ steps.get-container.outputs.result == 'fail' }}
+        uses: actions/github-script@v3
+        with:
+          script: |
+              core.setFailed('${{ steps.get-container.outputs.error }}')
```
joancafom, line 63: This will never be executed, as none of the cases above set `result::fail`. You may safely remove this block of code.

PR #7, opened 2022-06-15
Add trivy verification phase to the pipeline

fmulero:

**Description of the change**
Add trivy verification phase to the pipeline

**Benefits**

Run CVE scan during verification phase

### DIFF
--- a/.vib/vib-scheduled-verify.json
+++ b/.vib/vib-scheduled-verify.json
@@ -23,6 +23,23 @@
           "action_id": "container-image-lint"
         }
       ]
+    },
+    "verify": {
+      "context": {
+        "resources": {
+          "url": "{VIB_ENV_URL}",
+          "path": "{VIB_ENV_PATH}"
+        }
+      },
+      "actions": [
+        {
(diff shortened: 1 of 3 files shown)


.vib/vib-scheduled-verify.json
```diff
@@ -23,6 +23,23 @@
           "action_id": "container-image-lint"
         }
       ]
+    },
+    "verify": {
+      "context": {
```
recena, line 28: Do you need this section here?

fmulero:
I am going to close this PR because the workflow is not triggered, I have to check why

.vib/vib-scheduled-verify.json
fmulero: Yes, you are right. I need some additional parameterization

recena: I doubt if we need the `target_platform` since we are not deploying...

fmulero: `target_platform` is not required, Trivy analyse directly the image

PR #9, opened 2022-06-21
[containers/*] Enable scheduled workflow

fmulero:
**Description of the change**

Issue with [vmware-image-builder-action](https://github.com/vmware-labs/vmware-image-builder-action/pull/49) is already fixed. We can enable again the scheduled workflow

**Additional information**

`fail-fast` was disabled to avoid the eviction of any execution. By default if one of the jobs in a loop fails, the rest of the jobs will be aborted


rafariossaa:
LGTM

PR #16, opened 2022-07-06
Fix sync process

fmulero:

**Description of the change**

`bitnami/containers` repository contains all commits from each container in a plain structure (it doesn't have merge commits) Getting the list of commits in topological order helps to plain the tree ignoring the merge commits.

**Benefits**

Current sync errors with pgbouncer will be fixed

**Possible drawbacks**

Not found

### DIFF
--- a/scripts/sync.sh
+++ b/scripts/sync.sh
@@ -64,12 +64,12 @@ findCommitsToSync() {
     local max=100 # If we need to sync more than 100 commits there must be something wrong since we run the job on a daily basis
     # Get all commits IDs on the origin
     local commits=()
-    while IFS='' read -r line; do commits+=("$line"); done < <(git rev-list "${origin_name}/master" -- .)
+    # Get the revision list in topological order ignoring merge commits
+    while IFS='' read -r line; do commits+=("$line"); done < <(git rev-list --topo-order --no-merges "${origin_name}/master" -- .)
     for commit in "${commits[@]}"; do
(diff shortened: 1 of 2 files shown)


scripts/sync.sh
```diff
@@ -64,12 +64,12 @@ findCommitsToSync() {
...
-    while IFS='' read -r line; do commits+=("$line"); done < <(git rev-list "${origin_name}/master" -- .)
+    # Get the revision list in topological order ignoring merge commits
+    while IFS='' read -r line; do commits+=("$line"); done < <(git rev-list --topo-order --no-merges "${origin_name}/master" -- .)
     for commit in "${commits[@]}"; do
         if [[ "$commit" != "$last_synced_commit_id" ]] && [[ "$max" -gt "0" ]]; then
-            actual_commit_id="$(plainCommit "${commit}")"
-            if [[ "$commit" != "$SKIP_COMMIT_ID" ]] && [[ ! $commits_to_sync =~ (^|[[:space:]])$actual_commit_id($|[[:space:]]) ]]; then
-              commits_to_sync="${actual_commit_id} ${commits_to_sync}"
+            if [[ "$commit" != "$SKIP_COMMIT_ID" ]]; then
+              commits_to_sync="${commit} ${commits_to_sync}"
```
alemorcuq, line 72: Why this change?

fmulero: Now we don't need to resolve merge commits, so `plainCommit` is not needed and there is no risk to add twice the same commit

alemorcuq (approved):

PR #18, opened 2022-07-08
[containers/*] Fix issue with consecutive commits with the same subject

fmulero:
Signed-off-by: Bitnami Containers <bitnami-bot@vmware.com>

**Description of the change**

Use author date also to distinguish several consecutive commits with the same subject.

**Benefits**

Don't miss commits during synchronization process


migruiz4 (approved):
LGTM

PR #22, opened 2022-07-18
[bitnami/*] Remove sync workflow

jotadrilo:
**Description of the change**

This PR removes the "sync" workflow, which behavior is being replaced by automatic commits from an external CI system.

**Benefits**

The "sync" workflow is fully removed, and this opens the door to automatic updates with PRs that validate, build, and publish the published changes.

**Possible drawbacks**

The automatic updates do not fully work and we have to bring the "sync" workflow back.


carrodher (approved):

carrodher:
You should also remove the `sync.sh` (well, the whole `scripts` folder) since it is not going to be used anymore

fmulero (approved):
LGTM

PR #191, opened 2022-07-20
[bitnami/*] Change pipelines to get to build branch from readme

fmulero:
### Description of the change

Get branches to publish from readme changes. 

### Benefits

Detect releases done without code changes. For instance, forced releases with system package upgrades


.github/workflows/cd-pipeline.yaml
```diff
@@ -33,6 +33,17 @@ jobs:
               files_changed="$(echo $files_changed_data | jq -r '.files[] | .filename')"
               # Adding || true to avoid "Process exited with code 1" errors
               containers+=($(echo "$files_changed" | xargs dirname | grep -o "bitnami/[^/]*/[^/]*/[^/]*" | sort | uniq || true))
+              # Changes done in readme files?
+              readme_object="$(echo $files_changed_data | jq -r '[ .files[] | select(.filename | test("bitnami/[^/]+/README.md"))] | first')"
+              if [[ "${readme_object}" != "null" ]]; then
+                # Look for modified branch in Readme content
+                container_name="$(echo ${readme_object} | jq -r '.filename' | sed -nr "s|bitnami/([^/]+)/README.md|\1|p")"
+                branch="$(echo ${readme_object} | jq -r '.patch' | grep -o "\`latest\` ([^)]*)"  | sort -u | sed -nr "s|\`latest\` \((.+)/Dockerfile\)|\1|p")"
```
FraPazGal, line 41: This will only take into account branches used as `latest` right? What would happen if the changed branch is not latest. E.g.: branch `7` of elasticsearch.

Another concern is what will happen in cases where more than one branch is updated in the same PR, see https://github.com/bitnami/containers/pull/186

fmulero: Good catch!

To be perfectly blunt, I have no idea at the moment about how to solve the that PR (#186) automatically. There is no easy way in github (as far as I know) to run several executions and include a final job (like in this case) waiting for the result from all of them.

I'll work in the first point.

PR #342, opened 2022-07-22
Ignoring automated PR (Release activity) from the support flow.

agomezmoron:
Signed-off-by: Alejandro Gómez <morona@vmware.com>

### Description of the change

Ignoring automated PR (Release activity) from the support flow.

### Benefits

Ensuring the support flow only handles the support and not any flow related to the Release one.

### Possible drawbacks

None detected.

### DIFF
--- a/.github/workflows/comments.yml
+++ b/.github/workflows/comments.yml
@@ -7,12 +7,14 @@ permissions:
   repository-projects: write
   issues: write
 
+# To fix the concurrency when for example more than one label is added
 concurrency:
-  group: ${{ github.event_name != 'issues' && github.event.number || github.event.issue.number }}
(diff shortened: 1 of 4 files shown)


fmulero (commented):
Shouldn't we add  same condition to `clossing-issues.yml` workflow?

agomezmoron:
@fmulero no because that action will be triggered only **if has the `solved` label**. Automated PRs are excluded from the support process so will be out of the board and then the `solved` label will not exist (it's assigned when the card is moved into the `Solved` column OR when it's closed - but we are excluding that in the  `move-closed-issues.yml`

fmulero (approved):
You right! LGTM

PR #1055, opened 2022-07-28
[bitnami/wordpress-nginx] see #1051: fix wp-config.php path

### LINKED ISSUE #1051: [bitnami/wordpress-nginx] The wrong wp-config.php is checked for existence
### Name and Version

bitnami/wordpress-nginx:6

1. Map a local wp-content folder to /bitnami/wordpress/wp-content
2. Spin up a container
3. WordPress is not initialized

The startup script should configure WordPress because the /bitnami/wordpress/wp-config.php is not existing.

### What do you see instead?

The startup script is considering the app already initialized.

### Additional information

The error is at https://github.com/bitnami/bitnami-docker-wordpress-nginx/blob/master/6/debian-11/rootfs/opt/bitnami/scripts/libwordpress.sh#L231 since we're looking for /opt/bitnami/wordpress/wp-config.php instead of /bitnami/wordpress/wp-config.php.

/opt/bitnami/wordpress/wp-config.php is always present since it's part of the image.

ziodave:
### Description of the change

Fix the wp-config.php path.

### Benefits

The script will be able to correctly detect whether WordPress is already installed or not.

### Possible drawbacks

None.

### Applicable issues

  - fixes #1051 

### Additional information

https://github.com/bitnami/bitnami-docker-wordpress-nginx/pull/88

### DIFF
--- a/bitnami/wordpress-nginx/6/debian-11/rootfs/opt/bitnami/scripts/libwordpress.sh
+++ b/bitnami/wordpress-nginx/6/debian-11/rootfs/opt/bitnami/scripts/libwordpress.sh
@@ -228,7 +228,7 @@ wordpress_initialize() {
 
     # Check if WordPress has already been initialized and persisted in a previous run
     local -r app_name="wordpress"
-    if ! is_app_initialized "$app_name" || [[ ! -f "$WORDPRESS_CONF_FILE" ]]; then
+    if ! is_app_initialized "$app_name" || [[ ! -f "${BITNAMI_VOLUME_DIR}/${app_name}/wp-config.php" ]]; then
         # Ensure WordPress persisted directories exist (i.e. when a volume has been mounted to /bitnami)
         info "Ensuring WordPress directories exist"
         ensure_dir_exists "$WORDPRESS_VOLUME_DIR"


FraPazGal:
Hello @ziodave,

Thanks for the PR! I have a doubt regarding your use case just in case I'm missing something. When you are mapping a local `wp-content` folder, is it empty? The [is_app_initialized](https://github.com/bitnami/containers/blob/c6d8801222f2df4273ec1507a354ca2fba8f97de/bitnami/wordpress/6/debian-11/prebuildfs/opt/bitnami/scripts/libpersistence.sh#L114) checks so and it isn't it won't initialize the asset. The file path `$WORDPRESS_CONF_FILE` is a symlink of the one at `/bitnami/wordpress/wp-config.php`.

ziodave:
Hello,

> When you are mapping a local `wp-content` folder, is it empty?

No, it's not empty, which causes the script to believe that the app is already initialized, hence the check for wp-config.php

ziodave:
@FraPazGal do you need anything else from my side?

siddjellali:
up here (thanks)

FraPazGal:
Hello @ziodave,

Sorry about the delay, I was on PTO and our system didn't reassign this PR. Let me test and review this PR before merging.

bitnami/wordpress-nginx/6/debian-11/rootfs/opt/bitnami/scripts/libwordpress.sh
```diff
@@ -228,7 +228,7 @@ wordpress_initialize() {
 
     # Check if WordPress has already been initialized and persisted in a previous run
     local -r app_name="wordpress"
-    if ! is_app_initialized "$app_name" || [[ ! -f "$WORDPRESS_CONF_FILE" ]]; then
+    if ! is_app_initialized "$app_name" || [[ ! -f "${BITNAMI_VOLUME_DIR}/${app_name}/wp-config.php" ]]; then
```
FraPazGal, line 231: Hi, sorry about the delay! 

I agree that checking `$WORDPRESS_CONF_FILE` is not helpful as we are generating it in the postunpack stage and the correct check is the one you added. That said, I think we could remove the first condition of this `if` as it is only checking if `${BITNAMI_VOLUME_DIR}/${app_name}` is empty (which is implicitly done by the second condition):

https://github.com/bitnami/containers/blob/b3ea0b2b21484e5184c0acee0fa61228d591367a/bitnami/wordpress-intel/6/debian-11/prebuildfs/opt/bitnami/scripts/libpersistence.sh#L114

PR #1154, opened 2022-07-29
[bitnami/*] Changing token to avoid rate limiting on GITHUB_TOKEN

agomezmoron:
### Description of the change

After fixing the rate-limit of the bitnami-bot, I change the token to avoid a rate-limit of the GITHUB_TOKEN.

### Benefits

Ensuring we have no failing actions due to rate-limiting because of the GITHUB_TOKEN.

### Possible drawbacks

None detected


fmulero (approved):
LGTM

PR #1158, opened 2022-07-29
[bitnami/*] Using bitnami-bot to retrieve org team members

agomezmoron:
### Description of the change

After checking with @fmulero, the GITHUB_TOKEN cannot be use for this.

### Benefits

Ensuring we have no 403 errors on that workflow.

### Possible drawbacks

None


fmulero (approved):
LGTM

PR #1461, opened 2022-08-01
[bitnami/openldap] feature: make it possible to run some additional custom init scripts after the first run.

hacmao:
**Description of the change**

+ Change function `ldap_custom_init_scripts` to check whether the custom init script was executed or not.
+ If not, execute it.
+ If the file was executed before, skip executing.

**Benefits**
+ Make it possible to run some additional custom init scripts after the first run.

**Possible drawbacks**
+ No, I don't see one.


rafariossaa:
Hi,
  I am not very sure about this.
  What if an user want to execute this in each start ? I would move this logic to the provide scripts, so the user can control if they want it to be re-executed or not.
  What do you think ?

PR #1492, opened 2022-08-01
[bitnami/postgresql-repmgr] - Postgresql HA - solution for slip brain - multi master 

martinfar:
### Description of the change

This change incorporate the case for a failing primary Node with persistent volume attached to the Node. That create a problem .
Because the primary cannot move to another node. So, it can occur that the replicas don't have a upstream available.
So I incorporate the possibility for a standby node to self promote it self in case the Primary is more than **REPMGR_WAIT_PRIMARY_TIME** unavailable.

I also put random alterations to some waiting times to the standby replicas came at different point in time.

##### The benefits:

- more resilient cluster in case of definite failure of a primary node
- more control over the amount of time we wait for the primary. In case the node is busy restarting. This prevent the famous split brain Reference:   https://github.com/bitnami/charts/pull/9035

#### Possible drawbacks

- too much waiting time for a new primary.

### Applicable issues
 https://github.com/bitnami/charts/pull/9035

### Additional  information

Its important to apply this improvement with a **Liveness probe** with at least a bigger period for REPMGR_WAIT_PRIMARY_TIME , plus the time of the node setup.


martinfar:
Hi Team, 
Was it possible for you to review this merge ?

bitnami/postgresql-repmgr/10/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -222,13 +222,28 @@ repmgr_get_primary_node() {
...
+            repmgr_wait_primary_node || repmgr_switchover_standby
+
         else
+
+            info "Upstream detected: primary_host: '${upstream_host}:${upstream_port}'"
+
             primary_host="$upstream_host"
             primary_port="$upstream_port"
+
+
```
migruiz4, line 246: Please remove extra whitelines

```suggestion
            info "No upstream detected: primary_host: '${REPMGR_PRIMARY_HOST}:${REPMGR_PRIMARY_PORT}'"
            info "No upstream detected: node_network: '${REPMGR_NODE_NETWORK_NAME}:${REPMGR_PORT_NUMBER}'"
            if [[ "${REPMGR_PRIMARY_HOST}:${REPMGR_PRIMARY_PORT}" != "${REPMGR_NODE_NETWORK_NAME}:${REPMGR_PORT_NUMBER}" ]]; then
                debug "Primary hosts differences"
                primary_host="$REPMGR_PRIMARY_HOST"
                primary_port="$REPMGR_PRIMARY_PORT"
            fi
            repmgr_wait_primary_node || repmgr_switchover_standby
        else
            info "Upstream detected: primary_host: '${upstream_host}:${upstream_port}'"
            primary_host="$upstream_host"
            primary_port="$upstream_port"
```

bitnami/postgresql-repmgr/10/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -237,6 +252,60 @@ repmgr_get_primary_node() {
...
+repmgr_switchover_standby() {
+    local -i timeout="${REPMGR_WAIT_PRIMARY_TIME:-60}"
+
+    primary_host=""
+    primary_port="${REPMGR_PORT_NUMBER}"
+
+    info "Switchover from standby to primary . After unavailable primary for '$timeout' seconds . Assuming the primary role..."
+
+    echo "promote_me" >> "$REPMGR_PROMOTE_STANDBY_LOCK_FILE_NAME"
```
migruiz4, line 276: You can create an empty file using `touch`:
```suggestion
    touch "$REPMGR_PROMOTE_STANDBY_LOCK_FILE_NAME"
```

bitnami/postgresql-repmgr/10/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -237,6 +252,60 @@ repmgr_get_primary_node() {
...
+
+    debug "Marking node to be promoted from standby to primary by writing the following file: '$REPMGR_PROMOTE_STANDBY_LOCK_FILE_NAME' "
+
+    [[ -n "$primary_host" ]] && debug "Primary node: '${primary_host}:${primary_port}'"
+    echo "$primary_host"
+    echo "$primary_port"
```
migruiz4, line 282: I don't see any circumstances where $primary_host won't be empty. 

What is the purpose of the two `echo`?

bitnami/postgresql-repmgr/10/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -237,6 +252,60 @@ repmgr_get_primary_node() {
...
+}
+
+
+
```
migruiz4, line 286: Please use 1 single blank line instead of 3

bitnami/postgresql-repmgr/10/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -237,6 +252,60 @@ repmgr_get_primary_node() {
...
+#   None
+#########################
+
+repmgr_switchover_standby() {
+    local -i timeout="${REPMGR_WAIT_PRIMARY_TIME:-60}"
+
+    primary_host=""
+    primary_port="${REPMGR_PORT_NUMBER}"
+
+    info "Switchover from standby to primary . After unavailable primary for '$timeout' seconds . Assuming the primary role..."
```
migruiz4, line 274: No need to display the time, I prefer using static info messages as they make troubleshooting easier.

```suggestion
info "Primary node is unavailable. Assuming primary role..."
```

bitnami/postgresql-repmgr/10/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -469,6 +538,7 @@ repmgr_generate_repmgr_config() {
     # set the "--waldir" option accordingly
     local -r waldir=$(postgresql_get_waldir)
     local -r waldir_option=$([[ -n "$waldir" ]] && echo "--waldir=$waldir")
+    local -r randomize=$(( ( RANDOM % 10 )  + 5 ))
```
migruiz4, line 541: randomize is not needed, our libraries include the function `generate_random_string` which can be used to generate numeric values:
```
generate_random_string --type numeric --count 1
```

bitnami/postgresql-repmgr/10/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -485,7 +555,7 @@ failover='automatic'
 promote_command='$(repmgr_get_env_password) repmgr standby promote -f "${REPMGR_CONF_FILE}" --log-level DEBUG --verbose'
 follow_command='$(repmgr_get_env_password) repmgr standby follow -f "${REPMGR_CONF_FILE}" -W --log-level DEBUG --verbose'
 reconnect_attempts='${REPMGR_RECONNECT_ATTEMPTS}'
-reconnect_interval='${REPMGR_RECONNECT_INTERVAL}'
+reconnect_interval=$(expr $randomize + $REPMGR_RECONNECT_INTERVAL 2>/dev/null  || echo "$REPMGR_RECONNECT_INTERVAL")
```
migruiz4, line 558:
```suggestion
reconnect_interval=$(( 5 + REPMGR_RECONNECT_INTERVAL + $(generate_random_string --type numeric --count 1) ))
```

bitnami/postgresql-repmgr/13/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -560,13 +630,35 @@ EOF
 #########################
 repmgr_wait_primary_node() {
     local return_value=1
-    local -i timeout=60
-    local -i step=10
-    local -i max_tries=$((timeout / step))
+    local -i timeout="${REPMGR_WAIT_PRIMARY_TIME:-60}"
+    local -i randomize=$(( ( RANDOM % 20 )  + 1 ))
```
migruiz4, line 634: What is the point of randomizing the `max_retries`?

Randomizing the reconnection interval could make sense to avoid collisions, but I don't see the point of setting a random number of retries.

In addition, `max_retries` is not used for anything except for the debug message below, the loop condition will still retry $(timeout / step) times.

bitnami/postgresql-repmgr/13/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -583,6 +675,7 @@ repmgr_wait_primary_node() {
     return $return_value
 }
 
+
```
migruiz4, line 678:
```suggestion
```

migruiz4 (changes requested):
Hi @martinfar,

Thank you for your contribution! 
I would appreciate it if you could take a look at my comments.

bitnami/postgresql-repmgr/10/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
martinfar: I will get rid of it

bitnami/postgresql-repmgr/10/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
martinfar: this was for verification purposes I will get rid of it

bitnami/postgresql-repmgr/13/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
martinfar: You are right at this point, I will do a new commit with this corrected.
I will use your native method to randomize the step variable and remove additional default values. 
REPMGR_WAIT_PRIMARY_TIME and REPMGR_WAIT_PRIMARY_STEP are already declared in the postgresql-env.sh file.

martinfar (commented):
@migruiz4 thanks for your comments, I will commit again with all your corrections so we can review again for the final observations.

bitnami/postgresql-repmgr/10/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
martinfar: Hi @migruiz4, 
Thanks for taking the time to check. 
I already made the changes and review your comments, Can you take a look?

martinfar:
Hi, I wanted to know if the team was able to review this pull request ?

martinfar:
Hi Bitnami team. Where you able to review the proposed changes ?

bitnami/postgresql-repmgr/12/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -237,6 +249,56 @@ repmgr_get_primary_node() {
...
+    primary_host=""
+    primary_port="${REPMGR_PORT_NUMBER}"
+
+info "Primary node is unavailable. Assuming primary role..."
+
+    touch "$REPMGR_PROMOTE_STANDBY_LOCK_FILE_NAME"
+
+    debug "Marking node to be promoted from standby to primary by writing the following file: '$REPMGR_PROMOTE_STANDBY_LOCK_FILE_NAME' "
+
+    [[ -n "$primary_host" ]] && debug "Primary node: '${primary_host}:${primary_port}'"
```
migruiz4, line 277: Please keep an eye on the coding style. Be careful with indentation and additional white lines.
```suggestion
    primary_host=""
    primary_port="${REPMGR_PORT_NUMBER}"
    
    info "Primary node is unavailable. Assuming primary role..."
    touch "$REPMGR_PROMOTE_STANDBY_LOCK_FILE_NAME"
    debug "Marking node to be promoted from standby to primary by writing the following file: '$REPMGR_PROMOTE_STANDBY_LOCK_FILE_NAME' "
    [[ -n "$primary_host" ]] && debug "Primary node: '${primary_host}:${primary_port}'"
```

bitnami/postgresql-repmgr/12/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -237,6 +249,56 @@ repmgr_get_primary_node() {
...
+# Arguments:
+#   None
+# Returns:
+#   None
+#########################
+
+repmgr_switchover_standby() {
+    local -i timeout="${REPMGR_WAIT_PRIMARY_TIME:-60}"
+
+    primary_host=""
```
migruiz4, line 268: What is this variable used for? It is never assigned a value and referenced in the conditional at line 277.

bitnami/postgresql-repmgr/12/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -222,11 +222,23 @@ repmgr_get_primary_node() {
...
+
                 primary_host="$REPMGR_PRIMARY_HOST"
                 primary_port="$REPMGR_PRIMARY_PORT"
             fi
+
+            repmgr_wait_primary_node || repmgr_switchover_standby
+
         else
+            info "Upstream detected: primary_host: '${upstream_host}:${upstream_port}'"
+
```
migruiz4, line 241: Please remove extra white lines:
```suggestion
            info "No upstream detected: primary_host: '${REPMGR_PRIMARY_HOST}:${REPMGR_PRIMARY_PORT}'"
            info "No upstream detected: node_network: '${REPMGR_NODE_NETWORK_NAME}:${REPMGR_PORT_NUMBER}'"
            if [[ "${REPMGR_PRIMARY_HOST}:${REPMGR_PRIMARY_PORT}" != "${REPMGR_NODE_NETWORK_NAME}:${REPMGR_PORT_NUMBER}" ]]; then
                debug "Primary hosts differences"
                primary_host="$REPMGR_PRIMARY_HOST"
                primary_port="$REPMGR_PRIMARY_PORT"
            fi
            repmgr_wait_primary_node || repmgr_switchover_standby
        else
            info "Upstream detected: primary_host: '${upstream_host}:${upstream_port}'"
```

bitnami/postgresql-repmgr/12/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -560,13 +622,33 @@ EOF
...
+    local -i step=$(( REPMGR_WAIT_PRIMARY_STEP + $(generate_random_string --type numeric --count 1) ))
     local -i max_tries=$((timeout / step))
     local schemata
     info "Waiting for primary node..."
     debug "Wait for schema $REPMGR_DATABASE.repmgr on '${REPMGR_CURRENT_PRIMARY_HOST}:${REPMGR_CURRENT_PRIMARY_PORT}', will try $max_tries times with $step delay seconds (TIMEOUT=$timeout)"
     for ((i = 0; i <= timeout; i += step)); do
+
+        if [[ -z "$REPMGR_CURRENT_PRIMARY_HOST" ]] ; then
+          # because primary could be down while standby turn on, so the primary host could not be get from upstream at that momement
+          # asking on each step help to give the primary the opportunity to come back
```
migruiz4, line 635: Please fix typos in these comment lines.

Suggestion:
```
         # The primary node may not be available at the time the standby node turns on
         # Asking partner nodes who the primary node is to give it an opportunity to come back
```

bitnami/postgresql-repmgr/12/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
```diff
@@ -796,6 +878,14 @@ repmgr_initialize() {
             repmgr_upgrade_extension
         else
             debug "Skipping repmgr configuration..."
+            debug "Checking if node is marked for promotion "
```
migruiz4, line 881: Please remove additional whitespace at the end:
```suggestion
            debug "Checking if node is marked for promotion"
```

migruiz4 (changes requested):
Hi @martinfar, 

I'm sorry for the late response.
I have left some comments on the files corresponding to branch 12, but they need to be applied in every branch.

seriouz:
Why don't you just add the possibility to add nodes of the type witness? I thought witness nodes are just made for this purpose (https://repmgr.org/docs/4.2/using-witness-server.html)?
Currently only primary and standby nodes are possible to configure.

bitnami/postgresql-repmgr/12/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
martinfar: that variable it is not used any more, so i will delete that line

bitnami/postgresql-repmgr/12/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh
martinfar: suggestion accepted

martinfar:
@seriouz adding the witness is a major surgery to this development. However with the changes in this commit I was able to support a down datacenter without downtime. I am running the postgres cluster in production.

martinfar:
@migruiz4 I've already made the changes. Can you let me know if you need something else ?

martinfar:
Hi bitnami team, is it posible to go along with this merge?

martinfar:
Hi Bitnami team, any news about this request ?

migruiz4:
Hi @martinfar, I'm sorry for the late response.

After some in-depth evaluation of this proposal, I'm afraid we can not merge these changes as they are because of the following reasons:
1. We can not change the `reconnect_interval` to a random value as it is not deterministic (different deployments could behave differently).
  https://github.com/bitnami/containers/blob/beb5b0b050c0cb93ac756708339d99aea8bd1f21/bitnami/postgresql-repmgr/10/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh#L537-L539
2. The same applies to the changes proposed on the `repmgr_wait_primary_node` function. https://github.com/bitnami/containers/blob/beb5b0b050c0cb93ac756708339d99aea8bd1f21/bitnami/postgresql-repmgr/10/debian-11/rootfs/opt/bitnami/scripts/librepmgr.sh#L611-L615
3. The new REPMGR_PRIMARY_ROLE_LOCK_FILE_NAME could introduce additional corner cases, and its functionality is unclear.
  I would like to share my insights on it. 
  Considering this is the container setup.sh:
    ```bash
    # Ensure PostgreSQL & repmgr environment variables settings are valid
    repmgr_validate
    postgresql_validate

    # Set the environment variables for the node's role
    eval "$(repmgr_set_role)"

    # Ensure PostgreSQL is stopped when this script ends.
    trap "postgresql_stop" EXIT
    # Ensure 'daemon' user exists when running as 'root'
    am_i_root && ensure_user_exists "$POSTGRESQL_DAEMON_USER" --group "$POSTGRESQL_DAEMON_GROUP"
    # Prepare PostgreSQL default configuration
    repmgr_postgresql_configuration
    # Prepare repmgr configuration
    repmgr_generate_repmgr_config
    # Initialize PostgreSQL & repmgr
    repmgr_initialize
    ```
   What the proposed changes consist of:
    - The function `repmgr_wait_primary_node`  has been modified to retry the upstream connection to obtain a primary node.
    - At `eval "$(repmgr_set_role)"`, the container will try to connect to the upstream primary, and if a connection can't be established, then the node will start as primary if it is marked to, otherwise will start as standby.
    The proposed change is to run the modified `repmgr_wait_primary_node`, and if it fails, create the 'promote standby' lock file. This is executed either if the node will be assigned the primary or standby role, as far as it wasn't a primary before. If no upstream nodes can be reached, the lock file will exist.
    - During the `repmgr_initialize`, if the node has role 'primary' and it is not a first boot, the command `standby promote` will be executed.
  I'm afraid this logic is too complicated and would cause the troubleshooting to be more complicated. Additionally, the objective of the promotion lock file is unclear, in what scenario would it be helpful?

With the current proposed changes, looks like the objective is to run the `promote` command if the following conditions are met:
- Node wasn't a primary node in the previous execution.
- Upstream couldn't be reached.
- Node is marked as initial primary during current execution.

But that raises two additional questions:
- Is a lock file really necessary? Do we need to persist it between executions?
- Considering we would be promoting Standby nodes if they are isolated from upstream, wouldn't that cause even more split-brain scenarios?

It would be helpful if you could provide more information about in which scenarios would this feature would be beneficial, otherwise, we may have to close this PR.

martinfar:
Hi @migruiz4 . Thanks for the feedback. I will analyse your suggestion

martinfar:
> @martinfar If there is anything we could do to help you please let us know.

@migruiz4 In fact, if you can help me, I would like to know if have a better idea to remove the random time periods for a way where each will verify in a different time after the primary is unavailable. 
The idea is to avoid that two independent nodes decide to upgrade themselves to primary at the same. That will cause a split brain situation.

migruiz4:
> The idea is to avoid that two independent nodes decide to upgrade themselves to primary at the same. That will cause a split brain situation.

The best way to prevent split-brain is through quorum, not sure if this is the case, but I would always recommend deploying an odd number of nodes.

Normally, deploying an odd number of nodes (n), should be enough to prevent split-brain scenarios as (n / 2 + 1) nodes need to agree for a new primary to be promoted.

The witness node function is to force quorum in a 2-node deployment, where the witness node votes on the primary node election, but does not participate in the replication.

martinfar:
> The witness node function is to force quorum in a 2-node deployment, where the witness node votes on the primary node election, but does not participate in the replication.

I really think that a witness node is the solution to all of this... where is this feature in the roadmap ? I mean in this bitnami project [Postgres HA](https://github.com/bitnami/charts/tree/main/bitnami/postgresql-ha)

seriouz:
> > The witness node function is to force quorum in a 2-node deployment, where the witness node votes on the primary node election, but does not participate in the replication.
> 
> I really think that a witness node is the solution to all of this... where is this feature in the roadmap ? I mean in this bitnami project [Postgres HA](https://github.com/bitnami/charts/tree/main/bitnami/postgresql-ha)

Are thinking of something like this? https://github.com/bitnami/containers/issues/6663

martinfar:
> > > The witness node function is to force quorum in a 2-node deployment, where the witness node votes on the primary node election, but does not participate in the replication.
> > 
> > 
> > I really think that a witness node is the solution to all of this... where is this feature in the roadmap ? I mean in this bitnami project [Postgres HA](https://github.com/bitnami/charts/tree/main/bitnami/postgresql-ha)
> 
> Are thinking of something like this? #6663

Yes the idea is to implement a witness as you suggested in that issue

PR #2061, opened 2022-08-05
[bitnami/rabbitmq] Support additional virtual host via RABBITMQ_VHOSTS

nobiit:
### Description of the change
Support additional virtual host via RABBITMQ_VHOSTS

### Benefits
Can create additional vhosts quickly

### Possible drawbacks
None

### Applicable issues
None


Mauraza:
Hi @nobiit 

Thanks for the contribution! I will do some tests to check that everything works correctly and when I have more information I will update this thread.

Mauraza (approved):
Hi @nobiit,
The test went well so LGTM!
The change will be in the next revision of the container.
Thanks again for the contribution.

PR #2355, opened 2022-08-08
[bitnami/discourse] Use unicorn instead of passenger

wzrdtales:
Original https://github.com/bitnami/bitnami-docker-discourse/pull/234

discourse lately removed the setting from the UI, this image doesn't really work properly without disabling this option currently.

Signed-off-by: Tobias Gurtzick magic@wizardtales.com

---

right now requires to set serve_static_assets = true

---

important info from previous ticket:

the docker-compose seems to be outdated and deviates from the helm chart.
The helm chart sets already

```
  - DISCOURSE_PORT_NUMBER=8080
  - DISCOURSE_EXTERNAL_HTTP_PORT_NUMBER=80
```

These two variables, and adding those to the docker-compose lets it run just fine.

### DIFF
--- a/bitnami/discourse/2/debian-11/Dockerfile
+++ b/bitnami/discourse/2/debian-11/Dockerfile
@@ -17,7 +17,7 @@ RUN . /opt/bitnami/scripts/libcomponent.sh && component_unpack "node" "14.20.0-1
 RUN . /opt/bitnami/scripts/libcomponent.sh && component_unpack "git" "2.37.1-1" --checksum 39ea3040baa552b4760c1100a3713f86493620c0a74121c1ceba50fe17341878
 RUN . /opt/bitnami/scripts/libcomponent.sh && component_unpack "brotli" "1.0.9-151" --checksum ba4c4d26d232f8d91d1143078a633ba7c5ae33f5c4182684400ffb7bbc0ed262
 RUN . /opt/bitnami/scripts/libcomponent.sh && component_unpack "discourse" "2.8.7-1" --checksum bd120eab1a7403e691f608f57b574a9782391fe1c4a46ce40752cfa560f922c2
-RUN apt-get update && apt-get upgrade -y && \
+RUN apt-get update && apt-get upgrade -y && apt-get install sudo runit -y && \
     rm -r /var/lib/apt/lists /var/cache/apt/archives
 RUN chmod g+rwX /opt/bitnami
 RUN /opt/bitnami/ruby/bin/gem install --force bundler -v '< 2'
--- a/bitnami/discourse/2/debian-11/rootfs/opt/bitnami/scripts/discourse-env.sh
+++ b/bitnami/discourse/2/debian-11/rootfs/opt/bitnami/scripts/discourse-env.sh
@@ -22,6 +22,7 @@ export BITNAMI_DEBUG="${BITNAMI_DEBUG:-false}"
 # By setting an environment variable matching *_FILE to a file path, the prefixed environment
 # variable will be overridden with the value specified in that file
 discourse_env_vars=(
+    DISCOURSE_SERVE_STATIC_ASSETS
     DISCOURSE_DATA_TO_PERSIST
     DISCOURSE_ENABLE_HTTPS
     DISCOURSE_EXTERNAL_HTTP_PORT_NUMBER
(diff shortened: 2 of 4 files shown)


gongomgra:
@migruiz4, as you have more context than me with this PR and it looks like you have already started checking the changes in the associated issue, I'm handling it over to you so you can take the best decision.

bitnami/discourse/2/debian-11/rootfs/opt/bitnami/scripts/discourse-env.sh
```diff
@@ -92,6 +93,8 @@ export DISCOURSE_CONF_FILE="${DISCOURSE_BASE_DIR}/config/discourse.conf"
 export PATH="${BITNAMI_ROOT_DIR}/common/bin:${BITNAMI_ROOT_DIR}/brotli/bin:${BITNAMI_ROOT_DIR}/git/bin:${PATH}"
 
 # Discourse persistence configuration
+DISCOURSE_SERVE_STATIC_ASSETS="${DISCOURSE_SERVE_STATIC_ASSETS:-true}"
+export DISCOURSE_SERVE_STATIC_ASSETS="${DISCOURSE_SERVE_STATIC_ASSETS:-true}"
```
migruiz4, line 97: Please move these lines under the `# Discourse configuration` comment

bitnami/discourse/2/debian-11/rootfs/opt/bitnami/scripts/discourse-env.sh
```diff
@@ -92,6 +93,8 @@ export DISCOURSE_CONF_FILE="${DISCOURSE_BASE_DIR}/config/discourse.conf"
...
 export DISCOURSE_VOLUME_DIR="${BITNAMI_VOLUME_DIR}/discourse"
 export DISCOURSE_DATA_TO_PERSIST="${DISCOURSE_DATA_TO_PERSIST:-plugins public/backups public/uploads}"
```
migruiz4, line 100: Unicorn will fail to start if we use the current default value for DISCOURSE_PORT_NUMBER, please change the value from 3000 to 8080.

```
export DISCOURSE_PORT_NUMBER="${DISCOURSE_PORT_NUMBER:-8080}"
```

In addition, please replace the `ports` value at `docker-compose.yaml`.

migruiz4 (changes requested):
Hi @wzrdtales,

Please take a look at my comments below.

I have suggested several changes, including the removal of `runit` as it does not seem necessary to run unicorn.

prathamesh-gharat:
Will this be PR be merged soon?

bitnami/discourse/2/debian-11/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN . /opt/bitnami/scripts/libcomponent.sh && component_unpack "node" "14.20.0-1
 RUN . /opt/bitnami/scripts/libcomponent.sh && component_unpack "git" "2.37.1-1" --checksum 39ea3040baa552b4760c1100a3713f86493620c0a74121c1ceba50fe17341878
 RUN . /opt/bitnami/scripts/libcomponent.sh && component_unpack "brotli" "1.0.9-151" --checksum ba4c4d26d232f8d91d1143078a633ba7c5ae33f5c4182684400ffb7bbc0ed262
 RUN . /opt/bitnami/scripts/libcomponent.sh && component_unpack "discourse" "2.8.7-1" --checksum bd120eab1a7403e691f608f57b574a9782391fe1c4a46ce40752cfa560f922c2
-RUN apt-get update && apt-get upgrade -y && \
+RUN apt-get update && apt-get upgrade -y && apt-get install sudo runit -y && \
```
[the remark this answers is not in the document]
wzrdtales: if you change something that supposedly is working, then it would be nice if you just attach it. I don't have time to test a foreign theory. When we tested it didn't work without runit to run as non-root

carrodher:
The PR was reviewed but there is pending to implement some of the suggestions

wzrdtales:
there were no valid suggestions @carrodher 

Please review again, runit is needed, without the container is crashing

wzrdtales:
and please reopen

wzrdtales:
not stale

wzrdtales:
bla bla

mcules:
How are you doing with the PR? Can it be merged?
There are also massive performance problems here, which are probably caused by passenger.

wzrdtales:
yeah, we're maintaining it until today ourselves in a forked image... .

jacobowitz:
@migruiz4 
Can I assume that you will pick up this PR eventually? I need this PR merged as well. If nobody is picking this one up, I'd try, but I am lacking a lot of context here so I would rather not do it.

migruiz4:
Hi @jacobowitz,

We have an internal task to study if migrating `bitnami/unicorn` from passenger to unicorn is viable or not, but as we have to deal with other priorities it is currently in our Backlog.

As I mentioned in my review, the usage of `runit` in our images is a strong no from my side.

PR #3331, opened 2022-08-16
[bitnami/*] Change workflow to fix issues retrieving tag information

fmulero:

### Description of the change
* cd-pipeline is triggered on push events coming from bitnami-bot user
* Add new workflow for automated PRs. This change should reduce complexity in the rest of the workflows
* Retrieve tag information from OCI Label `org.opencontainers.image.ref.name` in the Dockerfile instead of using the README file.
* Depth in clone actions is set to 1, now there is no need to get full history

### Benefits

* Have an explicit workflow for automated PRs.
* Fix issues parsing changes from README file.
* Reduce the traffic and time consumed by checkout operations

### Possible drawbacks

None identified


fmulero:
Some execution examples:
* fmulero#21
* fmulero#22
* fmulero#23
* fmulero#24

.github/workflows/ci-pipeline.yaml
```diff
@@ -63,136 +35,67 @@ jobs:
...
-          curl --request POST \
-          --url https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/requested_reviewers \
-          --header 'authorization: Bearer ${{ secrets.BITNAMI_BOT_TOKEN }}' \
-          --header 'content-type: application/json' \
-          --data '{
-            "reviewers": ["fmulero"]
-            }' \
-          --fail
+      - name: Verify
+        if: ${{ needs.vib-verify.result != 'success' && needs.vib-verify.result != 'skipped'}}
```
FraPazGal, line 100: Can `needs.vib-verify.result` status be different from `success/skipped/failure`? If not, we may simplify this condition to:

 ```suggestion
        if: ${{ needs.vib-verify.result == 'failure'}}
```

File: .github/workflows/ci-automated-pipeline.yaml [hunk 2]
```diff
@@ -0,0 +1,155 @@
+name: '[CI/CD] CI Pipeline for automated PRs'
+on: # Triggered only on open PR events done by bitnami-bot on containers folders
+  pull_request_target:
+    types:
+      - opened
+    branches:
+      - main
+      - bitnami:main
+    paths:
```
FraPazGal, line 9: Should we use this filter for the non-automated CI? After all, we are also filtering it for the CD.

.github/workflows/ci-automated-pipeline.yaml
```diff
@@ -0,0 +1,155 @@
...
+
+          if [[ "${#flavors[@]}" -eq "1" ]]; then
+            # Changes done in only one container -> OK
+            echo "::set-output name=result::ok"
+            echo "::set-output name=container::${container_name}"
+            echo "::set-output name=path::${flavors[0]}"
+          else
+            # No changes detected or there are changes done in more than one container -> SKIP
+            echo -e "::set-output name=error::No changes detected or there are changes done in more than one container:\n${containers_dirs_changed}\n. The rest of the tests will be skipped."
+            echo "::set-output name=result::skip"
```
FraPazGal, line 70: I'm not sure we need this validation for automated PRs; they are expected to only modify 1 flavor.

Having said that, we could always leave this validation to check for a faulty PR created by our internal tools, but in that case, we need to modify the `auto-pr-review` logic. As it is now, setting the `result=skip` will make all the following jobs skip, including the `auto-pr-review` and no reviewer would be assigned to the PR.

.github/workflows/ci-pipeline.yaml
```diff
@@ -63,136 +35,67 @@ jobs:
...
-          # Now EAM commits has the following format "[bitnami/<container>] Release <tag>"
-          tag="$(git log --pretty=tformat:"%s" -n 1 --grep="[R|r]elease" --author bitnami-bot@vmware.com --author containers@bitnami.com --author containers-bot@bitnami.com -- ${{ needs.get-container.outputs.path }} | sed "s|\[.*\]||" | sed "s|[R|r]elease||" | awk '{print $1}')"
-          if [[ -z "${tag}" ]]; then
-            echo "No tag found for: ${{ needs.get-container.outputs.path }}"
-            exit 1
-          else
+          if [[ -d "${{ matrix.container }}" ]]; then
+            name="$(echo "${{ matrix.container }}" | awk -F '/' '{print $2}')"
+            # Build a tag based on current SHA
+            tag="$(echo "${{ matrix.container }}" | awk -F '/' -v run_number="${GITHUB_RUN_NUMBER}" '{printf "%s-rc.%s", $3, run_number}')"
```
FraPazGal, line 73: Why can't we retrieve the tag from the Dockerfile label as we do in the CD and instead create our custom one?

FraPazGal (commented):
Thanks a lot for this PR @fmulero, it looks really good! I have added several comments to it, please check them out when you can.

.github/workflows/ci-automated-pipeline.yaml
fmulero: Good question! The `verify` action is mandatory for all PR. If we add the filter and we change for example this file, we couldn't merge the PR because the verify action is not triggered. Automated PR are a little bit special because the scope of them is very focused on the realease of particular branch

.github/workflows/ci-automated-pipeline.yaml
fmulero: You right!, I'll revisit this

.github/workflows/ci-pipeline.yaml
```diff
@@ -63,136 +35,67 @@ jobs:
...
-              fi
-            fi
-          else
-            # Changes done in more than container -> SKIP
-            echo -e "::set-output name=error::Changes detected in more than one container directory:\n${containers_dirs_changed}\nIt is strongly advised to change only one container in a PR. The rest of the tests will be skipped."
+          if [[ "${#flavors[@]}" -le "0" ]]; then
+            echo "No changes detected in containers. The rest of the steps will be skipped."
             echo "::set-output name=result::skip"
+          else
+            containers_json=$(printf "%s\n" "${flavors[@]}" | jq -R . | jq -cs .)
```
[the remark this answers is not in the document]
fmulero: Good catch!. We can also restrict the number of [parallel jobs in the matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#defining-the-maximum-number-of-concurrent-jobs)

.github/workflows/ci-pipeline.yaml
fmulero: Because at the moment is not mandatory for the users to increase that label in the Dockerfile and it could overlaps with an existing version.

.github/workflows/ci-pipeline.yaml
```diff
@@ -63,136 +35,67 @@ jobs:
...
           # Container name
-          VIB_ENV_CONTAINER: ${{ needs.get-container.outputs.container }}
-          VIB_ENV_TAG: ${{ steps.get-tag.outputs.tag }}
-  auto-pr-review:
+          VIB_ENV_CONTAINER: ${{ steps.get-container-metadata.outputs.name  }}
+          VIB_ENV_TAG: ${{ steps.get-container-metadata.outputs.tag }}
+  verify-all:
+    # Ensure all containers passed the verification
     runs-on: ubuntu-latest
+    name: Verify
```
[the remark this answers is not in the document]
fmulero: I had the same internal discussion and I left it  as you can see here due to the GitHub configuration. PRs can be merged if there is a `Verify` job succeded (or skipped). We are running the 'VIB verify' in the matrix job, so this means a lot of jobs with the name `Verify XXXXX` can be triggered but they don't match the branch protection rule, so we need another 'Verify' job retrieving the result from all previous executions

.github/workflows/ci-pipeline.yaml
fmulero: `success/skipped/failure/cancelled` -> https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context

.github/workflows/ci-pipeline.yaml
```diff
@@ -63,136 +35,67 @@ jobs:
...
-            # Changes done in only one container -> OK
-            echo "::set-output name=result::ok"
-            echo "::set-output name=container::${container_name}"
-            echo "::set-output name=path::${flavors[0]}"
-          elif [[ "${#flavors[@]}" -le "0" ]]; then
-            # Changes done in the containers/ folder but not inside a container subfolder -> Look for readme changes
-            readme_object="$(echo $files_changed_data | jq -r '[ .[] | select(.filename | test("^bitnami/[^/]+/README.md"))] | first')"
-            if [[ "${readme_object}" == "null" ]]; then
-              echo "::set-output name=error::No changes detected in containers. The rest of the tests will be skipped."
-              echo "::set-output name=result::skip"
```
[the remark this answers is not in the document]
fmulero: I'll have a look

.github/workflows/ci-pipeline.yaml
FraPazGal: My bad, I forgot about the `canceled` one introduced by the matrix logic. 😄

.github/workflows/ci-automated-pipeline.yaml
FraPazGal: I agree with you. The required status for the `Verify` job is key and it's not worthwhile to explore a workaround for this.

.github/workflows/ci-automated-pipeline.yaml
fmulero: Updated, if there are more than one "flavor" changed the PR will fail and manual intervention will be required

.github/workflows/ci-pipeline.yaml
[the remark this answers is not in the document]
fmulero: Done!

.github/workflows/cd-pipeline.yaml
```diff
@@ -13,6 +13,9 @@ jobs:
   get-containers:
     runs-on: ubuntu-latest
     name: Get modified containers path
+    if: |
```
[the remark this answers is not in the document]
fmulero: That's correct. Currently our releases, our delivery pipeline, is managed externally, out of github. So to run the CD workflow we can only trust on PRs created by bitnami-bot.

.github/workflows/ci-automated-pipeline.yaml
```diff
@@ -0,0 +1,155 @@
...
+
+          if [[ "${#flavors[@]}" -eq "1" ]]; then
+            # Changes done in only one container -> OK
+            echo "::set-output name=result::ok"
+            echo "::set-output name=container::${container_name}"
+            echo "::set-output name=path::${flavors[0]}"
+          else
+            # No changes detected or there are changes done in more than one container -> FAIL. It doesn't make sense in automated PRs.
+            echo -e "::set-output name=error::No changes detected or there are changes done in more than one container:\n${containers_dirs_changed}\n. The rest of the tests will be skipped."
+            echo "::set-output name=result::fail"
```
FraPazGal, line 70: This is a minor thing but is there a need to use the `output.result` here? Given that `get-container` can only succeed or fail, I think we can just make the job finish (when it succeeds) or fail (maybe a `exit 1`). That way we avoid both the output param and the `show-error` job.

fmulero: You right!

.github/workflows/ci-pipeline.yaml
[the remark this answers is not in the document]
fmulero: I think you are talking about the feature in charts to avoid running vib-verify when only `.md` files are changed. The code above do a different thing, but I've added added the feature to skip the verify phase in changes to `.md` files only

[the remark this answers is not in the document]
FraPazGal: Indeed I selected the wrong lines 😄 The change to skip the CI lgtm!

FraPazGal (approved):
LGTM, thanks a lot for the PR @fmulero!

fmulero:
Woop Woop

PR #3479, opened 2022-08-18
[bitnami/kibana] Fix is_kibana_running function

day1118:
### Description of the change

Fix the `is_kibana_running` function used to check kibana status when running in background mode.
The previous check (`.status.overall.state`) works in Kibana 7.x, but the json structure looks to have changed in 8.x. 
I cannot find the `/api/status` endpoint documented to confirm/reference.

### Benefits

The container no longer hangs when running in background mode.

### Applicable issues

  - fixes #

### DIFF
--- a/bitnami/kibana/7/debian-11/rootfs/opt/bitnami/scripts/libkibana.sh
+++ b/bitnami/kibana/7/debian-11/rootfs/opt/bitnami/scripts/libkibana.sh
@@ -310,9 +310,9 @@ is_kibana_ready() {
     # Therefore, we must check the value is not 'true'
     [[ ! "$rewriteBasePath" = "false" ]] && basePath=$(kibana_conf_get "server.basePath")
     if is_kibana_running; then
-        local -r status="$(yq eval ".status.overall" - <<<"$(curl -s "127.0.0.1:${KIBANA_PORT_NUMBER}${basePath:-}/api/status")")"
-        [[ $(echo $status | yq eval ".state") = "green" ]] && return # Kibana 7
-        [[ $(echo $status | yq eval ".level") = "available" ]] && return # Kibana 8
+        # Kibana 7 expects .status.overall.state to be 'green', while 8 expects .status.overall.level to be 'available'
+        local -r status="$(yq eval '.status.overall | pick(["state", "level"]) | .[]' - <<<"$(curl -s "127.0.0.1:${KIBANA_PORT_NUMBER}${basePath:-}/api/status")")"
+        [[ "$status" = "green" || "$status" = "available" ]] && return
     else
         false
     fi
--- a/bitnami/kibana/7/debian-11/rootfs/opt/bitnami/scripts/libkibana.sh
+++ b/bitnami/kibana/7/debian-11/rootfs/opt/bitnami/scripts/libkibana.sh
@@ -310,8 +310,9 @@ is_kibana_ready() {
     # Therefore, we must check the value is not 'true'
     [[ ! "$rewriteBasePath" = "false" ]] && basePath=$(kibana_conf_get "server.basePath")
     if is_kibana_running; then
-        local -r state="$(yq eval ".status.overall.state" - <<<"$(curl -s "127.0.0.1:${KIBANA_PORT_NUMBER}${basePath:-}/api/status")")"
-        [[ "$state" = "green" ]]
+        local -r status="$(yq eval ".status.overall" - <<<"$(curl -s "127.0.0.1:${KIBANA_PORT_NUMBER}${basePath:-}/api/status")")"
+        [[ $(echo $status | yq eval ".state") = "green" ]] && return # Kibana 7
+        [[ $(echo $status | yq eval ".level") = "available" ]] && return # Kibana 8
     else
         false
     fi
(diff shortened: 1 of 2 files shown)


gongomgra:
Hi @day1118,

Thank you for your contribution! Unfortunately, your solution, although valid, only works with Kibana 8, but Kibana 7 still needs to check the `.status.overall.state` path. The `libkibana.sh` library is the same for both branches, and merging your changes will end-up breaking Kibana 7 in the next container release. Can you update your PR having this into account? You will also have to implement the same changes to both files (under the 8 and 7 directories).

day1118:
> Hi @day1118,
> 
> Thank you for your contribution! Unfortunately, your solution, although valid, only works with Kibana 8, but Kibana 7 still needs to check the `.status.overall.state` path. The `libkibana.sh` library is the same for both branches, and merging your changes will end-up breaking Kibana 7 in the next container release. Can you update your PR having this into account? You will also have to implement the same changes to both files (under the 8 and 7 directories).

Hi @gongomgra - I don't understand why there are two directories with separate `Dockerfiles` and two `libkibana.sh` if they are required to be identical. Given there are two `Dockerfiles`, do these files need to match?
Thanks

day1118:
Updated to handle this case @gongomgra

day1118:
Is there anything else you need for this @gongomgra ?

day1118:
> Is there anything else you need for this @gongomgra ?

Following up if there is anything else required to progress this @gongomgra? Kibana 8 does not work when you try to load dashboards due to this bug.
Thanks

gongomgra:
Hi @day1118,

Sorry for the delay on getting back to you on this. I'm reviewing and testing your changes internally to verify everything is correct and I'll get back to you as soon as possible.

gongomgra:
Hi @day1118,

After checking your changes, I think they can be simplified a bit by using the `pick` function of `yq`. Can you please update your changes using the snippet below?

```
# Kibana 7 expects .status.overall.state to be 'green', while 8 expects .status.overall.level to be 'available'
local -r status="$(yq eval '.status.overall | pick(["state", "level"]) | .[]' - <<<"$(curl -s "127.0.0.1:${KIBANA_PORT_NUMBER}${basePath:-}/api/status")")"
[[ "$status" = "green" || "$status" = "available" ]] && return
```

bitnami/kibana/7/debian-11/rootfs/opt/bitnami/scripts/libkibana.sh
```diff
@@ -310,8 +310,9 @@ is_kibana_ready() {
     # Therefore, we must check the value is not 'true'
     [[ ! "$rewriteBasePath" = "false" ]] && basePath=$(kibana_conf_get "server.basePath")
     if is_kibana_running; then
-        local -r state="$(yq eval ".status.overall.state" - <<<"$(curl -s "127.0.0.1:${KIBANA_PORT_NUMBER}${basePath:-}/api/status")")"
-        [[ "$state" = "green" ]]
+        local -r status="$(yq eval ".status.overall" - <<<"$(curl -s "127.0.0.1:${KIBANA_PORT_NUMBER}${basePath:-}/api/status")")"
+        [[ $(echo $status | yq eval ".state") = "green" ]] && return # Kibana 7
+        [[ $(echo $status | yq eval ".level") = "available" ]] && return # Kibana 8
```
gongomgra, line 315: Can you simplify this logic a bit?

```
# Kibana 7 expects .status.overall.state to be 'green', while 8 expects .status.overall.level to be 'available'
local -r status="$(yq eval '.status.overall | pick(["state", "level"]) | .[]' - <<<"$(curl -s "127.0.0.1:${KIBANA_PORT_NUMBER}${basePath:-}/api/status")")"
[[ "$status" = "green" || "$status" = "available" ]] && return
```

gongomgra (changes requested):

bitnami/kibana/7/debian-11/rootfs/opt/bitnami/scripts/libkibana.sh
day1118: Updated @gongomgra

gongomgra:
Hi @day1118,

Thanks for applying the changes. Unfortunately, it looks like your branch is missing some files in the `bitnami:main` branch and the verification jobs can't be run. Can you update your branch with main?

day1118:
> Can you update your branch with main?

Rebased @gongomgra

gongomgra (approved):
LGTM!

gongomgra:
Hi @day1118,

Thanks for your contribution! I'm merging your changes now!

day1118:
Any idea why the build was skipped @gongomgra ?
https://github.com/bitnami/containers/actions/runs/3044658485

gongomgra:
Hi @day1118,

Yes, it is expected that the action is skipped here in GitHub. Don't worry. The new image is released from our custom pipeline. I can confirm it's been already published in the different registries including your changes.

```
$ docker rmi bitnami/kibana:8
$ docker pull bitnami/kibana:8
(...)
$ docker-compose up -d
$ docker exec -it -u root 783605336efb grep "status" /opt/bitnami/scripts/libkibana.sh
        # Kibana 7 expects .status.overall.state to be 'green', while 8 expects .status.overall.level to be 'available'
        local -r status="$(yq eval '.status.overall | pick(["state", "level"]) | .[]' - <<<"$(curl -s "127.0.0.1:${KIBANA_PORT_NUMBER}${basePath:-}/api/status")")"
        [[ "$status" = "green" || "$status" = "available" ]] && return
(...)
```

PR #4027, opened 2022-08-23
[bitnami/wordpress] allow only one source of truth

wibed:
remove redundant configuration.


rafariossaa:
Hi,
  Could you provide more background about this change ?

PR #4234, opened 2022-08-24
[bitnami/apache] Allow to know when setup is done bitnami/charts#11770

### LINKED ISSUE #11770: [bitnami/apache] [Question] Why initialDelaySeconds is needed?
### Name and Version

bitnami/apache

Why the `readinessProbe.initialDelaySeconds` is needed?

https://github.com/bitnami/charts/blob/6cb91fc2558d21754357e5ef2b1dee6fdb93e272/bitnami/apache/values.yaml#L298

The apache chart uses 30 seconds

If the `readinessProbe` is doing a `http` request to `/`, why the `initialDelaySeconds` is needed? If the call is success, why not just to use 0 delay?

Is there a problem in kubernetes services or apache and that's why the chart uses a high initialDelaySeconds? Your [nginx chart](https://github.com/bitnami/charts/blob/b786e6e7e7d2431c61aa6bd2dacef14e29c24768/bitnami/nginx/values.yaml#L317) uses 5 seconds.

set `initialDelaySeconds=0` in `startupProbe`, `livenessProbe` and `readinessProbe`

If the container needs to do a migration or an [initial setup](https://github.com/bitnami/containers/blob/main/bitnami/apache/2.4/debian-11/rootfs/opt/bitnami/scripts/apache/entrypoint.sh#L22) maybe the chart should enable `startupProbe`. So the entry point migrates or do the initial setup and then starts apache? `startupProbe.failureThreshold` could be very high to let the migration/setup ends, but with an initial daily = 0

JoniJnm:
### Description of the change

Creates a file in /tmp directory to allow to know other applications when the setup is done.

### Benefits

Allow to set a better startupProbe when using kubernetes. Or create a docker [healthCheck](https://docs.docker.com/engine/reference/builder/#healthcheck)

### Possible drawbacks

Users with a mounted /tmp directory with custom permissions?

### Applicable issues

  - fixes bitnami/charts#11770

### Additional information

Related: bitnami/charts#12144

### DIFF
--- a/bitnami/apache/2.4/debian-11/rootfs/opt/bitnami/scripts/apache/setup.sh
+++ b/bitnami/apache/2.4/debian-11/rootfs/opt/bitnami/scripts/apache/setup.sh
@@ -61,3 +61,6 @@ fi
 
 # Fix logging issue when running as root
 ! am_i_root || chmod o+w "$(readlink /dev/stdout)" "$(readlink /dev/stderr)"
+
+# Allow to know when setup is done (used by startupProbe)
+date > /tmp/bitnami-setup-done


rafariossaa:
Hi,
  Could yo provide more background on this ? I am not sure why you need to check for a file instead of waiting for apache listening to its port ?

JoniJnm:
The PR in [bitnami/charts/apache](https://github.com/bitnami/charts/pull/12144) says why:

Why the new startupProbe is using a custom command instead of checking if http port is open? Because sometimes the setup needs to restart the service to apply some changes. For example:

https://github.com/bitnami/containers/blob/main/bitnami/mysql/8.0/debian-11/rootfs/opt/bitnami/scripts/mysql/setup.sh#L38
https://github.com/bitnami/containers/blob/main/bitnami/rabbitmq/3.10/debian-11/rootfs/opt/bitnami/scripts/rabbitmq/setup.sh#L20
So the probe could be success (port open) but the service is not ready because needs a restart.

fmulero:
Thanks Jónatan for this PR. I've just open an internal discussion because I think this is a valuable feature that can be included in other containers. I'll come back with any news

fmulero:
Please bear with me. I hope take it a glance soon.

bitnami/apache/2.4/debian-11/rootfs/opt/bitnami/scripts/apache/setup.sh
```diff
@@ -61,3 +61,6 @@ fi
 
 # Fix logging issue when running as root
 ! am_i_root || chmod o+w "$(readlink /dev/stdout)" "$(readlink /dev/stderr)"
+
+# Allow to know when setup is done (used by startupProbe)
+date > /tmp/bitnami-setup-done
```
JoniJnm, line 66: Another option (instead of create the file `/tmp/bitnami-setup-done`) is to start mysql in a different port (so the health-check will fail) and after the setup and the stop, start it again with the right port

JoniJnm: I released how you fix this problem in other containers. For example in mysql, during the setup, you start the process with `--bind-address=127.0.0.1`, so the external connection should fails.

This can be improved using a different port. In apache there isn't a bg process during the setup so this change is no necessary.

Issue #1482, opened 2022-08-01
Error 80 during startup with TLS configuration


christiansicari:
### Name and Version

bitnami/openldap

### What steps will reproduce the bug?

I am currently using the following compose 

```yaml
# for certs https://deliciousbrains.com/ssl-certificate-authority-for-local-https-development/#becoming-certificate-authority
version: '2'

services:
  openldap:
    image: bitnami/openldap:2.5.6
    ports:
      - '1389:1389'
      - '1636:1636'
    environment:
      - LDAP_ADMIN_USERNAME=admin
      - LDAP_ADMIN_PASSWORD=adminpassword
      - LDAP_USERS=user01,user02
      - LDAP_PASSWORDS=password1,password2
      - LDAP_ENABLE_TLS=yes
      - BITNAMI_DEBUG=true
      - LDAP_TLS_CERT_FILE=/opt/bitnami/openldap/certs/openldap.crt
      - LDAP_TLS_KEY_FILE=/opt/bitnami/openldap/certs/openldap.key
      - LDAP_TLS_CA_FILE=/opt/bitnami/openldap/certs/openldapCA.crt
    volumes:
      - /home/ubuntu/ldap/certs/myCA.pem:/opt/bitnami/openldap/certs/openldapCA.crt
      - /home/ubuntu/ldap/certs/ldap.key:/opt/bitnami/openldap/certs/openldap.key
      - /home/ubuntu/ldap/certs/ldap.crt:/opt/bitnami/openldap/certs/openldap.crt
      #- $PWD/openldap:/bitnami/openldap
```

For generating the certificates I used this guide [https://deliciousbrains.com/ssl-certificate-authority-for-local-https-development/](url)

### What do you see instead?

At the bootstrap I get this error:

```console
$ docker-compose logs -f
Attaching to ldap_openldap_1
openldap_1  |  16:07:08.03 INFO  ==> ** Starting LDAP setup **
openldap_1  |  16:07:08.06 INFO  ==> Validating settings in LDAP_* env vars
openldap_1  |  16:07:08.07 INFO  ==> Initializing OpenLDAP...
openldap_1  |  16:07:08.07 DEBUG ==> Ensuring expected directories/files exist...
openldap_1  |  16:07:08.08 INFO  ==> Creating LDAP online configuration
openldap_1  |  16:07:08.11 INFO  ==> Starting OpenLDAP server in background
openldap_1  |  16:07:08.16 INFO  ==> Configure LDAP credentials for admin user
[...]

PR #4318, opened 2022-08-26
[bitnami/openldap] Follow LDAP_LOGLEVEL in setup phase

tatankat:
### Description of the change

Follow the LDAP_LOGLEVEL environment variable already in the setup phase.
The mechanisms are copied from ldap_stop and run.sh

### Benefits

More logging will be generated on first startup of the slapd daemon. The startup may fail, for example when permissions on the key file do not permit reading by the user running the slapd daemon. Without this patch, it will fail with a generic error and quit.
Using this patch, with BITNAMI_DEBUG set to true, the logs will tell you slapd can't read the key file. The logging can be tuned with the LDAP_LOGLEVEL variable.

### Possible drawbacks

None I see. Output can be tuned with BITNAMI_DEBUG and LDAP_LOGLEVEL

### DIFF
--- a/bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
+++ b/bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
@@ -196,12 +196,19 @@ is_ldap_not_running() {
 #   None
 #########################
 ldap_start_bg() {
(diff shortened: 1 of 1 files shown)


tatankat:
Using this patch would also show what the issue is in #1482
(and it was most probably the permission issue I mentioned)

migruiz4 (approved):
Thank you very much for your contribution @tatankat! Looks good to me!

PR #4453, opened 2022-08-29
[bitnami/*]Use github token to reduce the number of labeled events

fmulero:

### Description of the change

Follow up #4452. Use explicitly `GITHUB_TOKEN` as `repo-token`  in the configuration of the action

### Benefits

Avoid [hidden 403 errors](https://github.com/bitnami/containers/runs/8076184162?check_suite_focus=true)

### Possible drawbacks

None


FraPazGal (approved):
LGTM

PR #4766, opened 2022-09-01
[bitnami/*] Containers CI Automated PRs should follow same matrix approach as regular PRs

fmulero:

### Description of the change

CI for Automated PRs should follow same matrix approach as regular PRs. To achieve that goal the following changes has been implemented:
* Workflow for automated PRs has been removed.
* If an Automated PR fails, it will be handled by support workflows to include it in the support dashboard and manage its lifecycle as a regular PR.
* `Verification Summary` job was included to ensure all modified containers are passing the VIB verify.
* Support workflows were simplified removing the 'bitnami-bot' conditions.
* CI workflow is not listening for reopened events. That event creates a lot of noise, in most of the cases the branches has been removed and all jobs will fail.

### Benefits

* Clearer workflows
* No big differences between automated and non automated PRs
* Labelling logic and support dashboard management is responsibility of support workflows.

### Possible drawbacks

None identified


.github/workflows/ci-pipeline.yaml
```diff
@@ -95,13 +107,63 @@ jobs:
...
+                echo "Well done! Everything looks good. Please wait for the Bitnami Team review."
+              fi
+            fi
+          fi
+  auto-pr-review:
+    runs-on: ubuntu-latest
+    name: Reviewal for automated PRs
+    if: |
+      always() &&
+      github.actor == 'bitnami-bot'
```
FraPazGal, line 143: If I'm not mistaken, this will launch the job for every labeled and synchronized event triggered by a PR created by `bitnami-bot`. 

I don't think it is necessary to execute this once the CI has failed once. If an agent needs to check it, I would avoid automerging anything in case the process requires the PR not to be merged immediately. We would also avoid additional comments if the CI keeps failing.

I think we could add another condition so the CI should have been triggered by the labeling of `verify`, though that is not a perfect solution.

.github/workflows/ci-pipeline.yaml
```diff
@@ -95,13 +107,63 @@ jobs:
...
+          token: ${{ secrets.BITNAMI_BOT_TOKEN }}
+          merge-method: squash
+          pull-request: ${{ github.event.number }}
+      - name: PR Approval
+        if: ${{ needs.verification-summary.result == 'success' }}
+        uses: hmarr/auto-approve-action@v2.4.0
+        with:
+          pull-request-number: ${{ github.event.number }}
+      - name: Manual review required
+        if: ${{ needs.vib-verify.result != 'success' }}
```
FraPazGal, line 163: Shouldn't we use `needs.verification-summary` here as well?

FraPazGal (commented):
Thanks for the PR @fmulero, looking really good! 

I just have a concern about how our changes could inflate our support stats. I'm mainly doubtful of what entails to add the automated PRs to the Support board and add the `solved` label to closed ones. This is probably a matter of filtering the PRs from `bitnami-bot` in our stats, but I'm missing some context here.

Besides that, I added a couple of comments to the code. Please check them when possible.

.github/workflows/ci-pipeline.yaml
fmulero: You right! Good Catch

.github/workflows/ci-pipeline.yaml
fmulero: I am not sure if I got it well. If the PR fails and some user review it, resolving conflicts or adding other labels the action is triggered by that user not bitnami-bot.

.github/workflows/ci-pipeline.yaml
```diff
@@ -95,13 +107,64 @@ jobs:
...
     if: ${{ always() }}
     steps:
-      - name: Verify
-        if: ${{ needs.vib-verify.result != 'success' && needs.vib-verify.result != 'skipped'}}
-        run: exit 1
+      - name: Verification Summary
+        run: |
+          if [[ "${{ needs.get-containers.result }}" != "success" ]]; then
+            echo "Verification is required."
+            echo " If you've just created this PR, don't worry about this message. Bitnami Team has to review it and make possible the verification."
```
joancafom, line 123:
```suggestion
            echo "If you've just created this PR, don't worry about this message. Bitnami Team has to review it and make the verification possible."
```

.github/workflows/ci-pipeline.yaml
joancafom: Maybe I am missing something, but `bitnami-bot` will no longer label the PRs & I don't expect it to push more commits (and trigger synchronized) either, so this condition might be enough.

joancafom (commented):
Thanks for the PR! 

As @FraPazGal commented, I am doubtful whether including automated PRs in the Dashboard is a good idea. Metrics aside, most of these PRs are expected to succeed, so it will directly translate to many moving cards that, IMHO, won't add any real value.

joancafom (approved):
LGTM! Thanks for your changes 🚀

PR #5007, opened 2022-09-02
Fix running openldap as $LDAP_DAEMON_USER

tatankat:
Running openldap as the default $LDAP_DAEMON_USER (who is not member of the root group) failed.
This happens for example when starting the container as root or in appveyor Windows image (and probably on any other Windows as Administrator), as can be seen here: https://ci.appveyor.com/project/release-joomla/joomla-cms/builds/44665987

### Description of the change

The changes make sure that:
- slapadd is called as the same non-privileged user who runs the slapd daemon (which it should anyway), so the slapd daemon has permission to read the created slapd configuration ldif files.
- the slapd daemon has write permission for the pid file in its directory

### Benefits

Running the slapd daemon as the non-privileged $LDAP_DAEMON_USER will work in (appveyor) Windows and when the container started as root

### Possible drawbacks

None


carrodher (approved):
Thanks for the changes! Everything LGTM

PR #5477, opened 2022-09-07
[bitnami/*] Add individual VIB pipelines and ARM support

FraPazGal:
Signed-off-by: FraPazGal <fdepaz@vmware.com>

### Description of the change

This PR adds support for ARM verification and publishing and modifies the container registry where images are published. 

It also replaces the general `vib-verify.json` and `vib-publish.json` pipeline definitions used by every asset with individual verify/publish pipelines for each asset. With per-asset pipelines, we can define the supported architectures for each asset. 

### Benefits

- Use of a proper container registry instead of a mock-up one.
- Created the file structure for Cypress/Goss tests addition.
- ARM support

### Possible drawbacks

Although the use of individual pipelines has its benefits, it also restricts us to divide the supported architectures per asset and not per `asset/branch`. As such, for assets with only some branches supporting ARM, it was decided to not publish ARM to any of them.

The assets' branches affected by this are:

- envoy/1.21
- envoy/1.22
- envoy/1.23
- jupyter-base-notebook/1
- mariadb/10.5
- mariadb/10.6
- mariadb/10.7
- mariadb/10.8
- mariadb-galera/10.5
- mariadb-galera/10.6
- mariadb-galera/10.7
- mariadb-galera/10.8
- odoo/15
- parse/5

As there is no testing done as part of VIB right now and the used container registry is a private one this is not critical right now, though we'll have to look for a long-term solution.


beltran-rubo:
I see three patterns we are following for the `architectures`, can we use only two? e.g. do not add the option to the pipeline so it is `amd64` by default and specify the architectures whenever they are supported.

FraPazGal:
Hi @beltran-rubo, there should be only two patterns:

1. The `container-image-package` action adds both `amd64` and `arm64` as architectures
2. Only `amd64` is added as architecture. 

I may have missed a pattern, could you please give me an example? I decided to add everything directly to the pipeline definition as that is the approach we are using in `bitnami/charts` with those assets that have Cypress/Goss tests.

beltran-rubo:
Sorry I was mixing `verify` and `publish` pipelines.

beltran-rubo (approved):

PR #6418, opened 2022-09-14
[bitnami/*] Split workflow responsibilites

fmulero:
### Description of the change

With these changes we are trying to keep under control the responsibility of each workflow:
* `triage.yml`: Listening for new Issues or pull request. Its goal is to include them in the support project assigning the right column.
* `moving-cards.yml`: Listening for card movements. Its goal is to assign the right labels for each movement to keep the columns tracked and assign a bitnami team member to review the issue or the PR associated.
* `move-closed-issues.yml`: Listening for close events. Its goal is to move the card associated with that issue/PR to the "solved" column.
* `comments.yml`: Listening for new comments. It will move the card associated to the right column based on the author of the comment and card status.
* `ci-pipeline.yml`: Listening for code changes and label events in pull requests. It will run the verification of the changes. If the author is `bitnami-bot` and changes goes well the pull request will be merged automatically, if there is a failure a comment will be added with the error (this event will trigger the `comments.yml` workflow).

Other changes applied:
* Permissions: The permissions were reviewed according to the tokens used in each workflow.
* Remove unnecessary `concurrency` blocks.

### Benefits

Split responsibilities between support and CI github workflows. Keep clear where an issue or PR is labeled, moved or assigned.

### Possible drawbacks

None identified, but several changes were done without testing opportunity.


.github/workflows/ci-pipeline.yaml
```diff
@@ -95,7 +81,7 @@ jobs:
           if [[ -d "${{ matrix.container }}" ]]; then
             name="$(echo "${{ matrix.container }}" | awk -F '/' '{print $2}')"
             tag=""
-            if [[ "${{ github.actor }}" == "bitnami-bot" ]]; then
+            if [[ "${{ github.event.pull_request.user.login }}" == "bitnami-bot" ]]; then
```
agomezmoron, line 84: I don't want to be squeamish but I'd use the ' instead of " for the _bitnami-bot_ string like you did before.

.github/workflows/comments.yml
```diff
@@ -26,32 +20,31 @@ jobs:
...
+        if: ${{ github.actor != 'bitnami-bot' && (!contains(github.event.issue.labels.*.name, 'bitnami')) && contains(fromJson(env.BITNAMI_TEAM), github.actor) }}
         with:
           project-name: Support
           column-name: Pending
-          token: "${{ secrets.GITHUB_TOKEN }}"
+          token: "${{ secrets.BITNAMI_BOT_TOKEN }}"
       - name: Move into In Progress
         uses: peter-evans/create-or-update-project-card@v2
-        if: ${{ contains(github.event.issue.labels.*.name, 'in-progress') && (!contains(fromJson(env.BITNAMI_TEAM), github.event.comment.user.login)) }}
+        if: ${{ github.actor != 'bitnami-bot' && contains(github.event.issue.labels.*.name, 'in-progress') && (!contains(fromJson(env.BITNAMI_TEAM), github.actor)) }}
```
agomezmoron, line 39: Although I think is fine, careful, it's not the same `github.actor` than `github.event.comment.user.login`. The first one is the user who triggered the initial workflow run. The other one, is the user who wrote the comment. In this case we are based on comments event, is the same but if this event is triggered by any other event, the `github.actor`can be different than the author of the comment.

.github/workflows/comments.yml
```diff
@@ -26,32 +20,31 @@ jobs:
...
         with:
           project-name: Support
           column-name: In progress
-          token: "${{ secrets.GITHUB_TOKEN }}"
+          token: "${{ secrets.BITNAMI_BOT_TOKEN }}"
```
agomezmoron, line 43: When you move into In Progress, do you really need to trigger the `moving-cards.yml` workflow?

.github/workflows/triage.yml
```diff
@@ -45,14 +39,23 @@ jobs:
...
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
-          add-labels: ${{ (!contains(fromJson(env.BITNAMI_TEAM), github.actor)) && 'triage' || 'bitnami' }}
+          add-labels: ${{ github.actor == 'bitnami-bot' && 'auto-merge' || ((!contains(fromJson(env.BITNAMI_TEAM), github.actor)) && 'triage' || 'bitnami') }}
           # For reopened issues
-          remove-labels: "solved"
\ No newline at end of file
+          remove-labels: "solved"
+      - name: Verify labeling
+        # Only if the actor is bitnami-bot
+        if: ${{ github.actor == 'bitnami-bot' }}
```
agomezmoron, line 56: Will we trigger always this although is an issue? I'd add to the if condition that should be a PR (thinking in the future, if an action fails for example, we decide `bitnami-bot`to open an issue, then this step will be triggered).

fmulero (commented):
Thanks @agomezmoron Could you have a look to my comments?

.github/workflows/triage.yml
fmulero: This code was removed, but you right I am not distinguishing between PRs and issues

.github/workflows/moving-cards.yml
```diff
@@ -42,20 +43,26 @@ jobs:
...
+          add-labels: ${{ needs.get-issue.outputs.creator == 'bitnami-bot' && 'auto-merge' || 'triage' }}
           remove-labels: "on-hold, in-progress, solved"
+      - name: Triage labeling - verify
+        # Only if moved into triage and the creator is bitnami-bot
+        if: ${{ github.event.project_card.column_id == env.TRIAGE_COLUMN_ID && needs.get-issue.outputs.creator == 'bitnami-bot' }}
+        uses: andymckay/labeler@1.0.4
+        with:
+          add-labels: verify
+          # Required to trigger CI workflow
+          repo-token: "${{ secrets.BITNAMI_BOT_TOKEN }}"
```
agomezmoron, line 67: Why do we need to trigger the `labeled` events?

.github/workflows/moving-cards.yml
```diff
@@ -42,20 +43,26 @@ jobs:
...
         # NOTE: This step has to be modified when a new column for automated PR is added
-        if: |
-          github.event.project_card.column_id == env.BITNAMI_COLUMN_ID &&
-          (needs.get-issue.outputs.creator == 'bitnami-bot')
+        if: ${{ github.event.project_card.column_id == env.BITNAMI_COLUMN_ID }}
         uses: andymckay/labeler@1.0.4
         with:
-          add-labels: "review-required"
-          remove-labels: "auto-merge"
+          add-labels: ${{ needs.get-issue.outputs.creator == 'bitnami-bot' && 'review-required' || 'bitnami' }}
```
agomezmoron, line 73: Careful with this. We said to add the `bitnami` label **only** when we (Bitnami team) creates the issue/PR, not when the bot does it.

.github/workflows/moving-cards.yml
```diff
@@ -92,16 +100,24 @@ jobs:
         uses: xom9ikk/dotenv@v1.0.2
         with:
           path: .github/workflows/
+      # NOTE: The assignment rules has to be changed when the new column for automated PR is added
```
agomezmoron, line 112: I'd create the column now, I'd add the ID to the .env file and will start leaving the "pending changes" commented to ensure in the future it's covered.

.github/workflows/moving-cards.yml
fmulero: `ci-pipeline` is listening for labeled or synchronize events

.github/workflows/moving-cards.yml
```diff
@@ -92,16 +100,24 @@ jobs:
...
+        with:
+          numOfAssignee: 1
+          teams: ${{ github.event.project_card.column_id == env.BITNAMI_COLUMN_ID && env.SUPPORT_TEAM_NAME || env.TRIAGE_TEAM_NAME }}
+          repo-token: "${{ secrets.BITNAMI_BOT_TOKEN }}"
+      - name: Reassign when moved into 'In progress' from 'Triage'
+        # Reassigned when moved into In progress FROM Triage
         if: |
-          needs.get-issue.outputs.assignees == '[]' ||
-          github.event.project_card.column_id == env.IN_PROGRESS_COLUMN_ID && github.event.changes != null && github.event.changes.column_id &&
+          github.event.action == 'moved' && needs.get-issue.outputs.assignees != '[]' &&
```
agomezmoron, line 124: If the issue was in triage and you move it, the assignees are not empty, not? Or am I missing anything? :/

.github/workflows/moving-cards.yml
fmulero: We can reach `From Bitnami` column in these cases:
* Bitnami Team member creates a PR, `triage.yml` workflow will create the card directly in in 'From Bitnami'.
* Automated PR fails, `ci-workflow.yml` will add a comment and `comments.yml` workflow will move the card to `From Bitnami`, in this case the label added will be `review-required`. NOTE: This will be changed when the new column for failed PRs is added.
* A Bitnami Team member moves the card to `From Bitnami`, I think it makes sense to have the `bitnami` label because we are using the label to track the column where the issue/PR belongs to

.github/workflows/moving-cards.yml
fmulero: I didn't do it because this PR is complex and I didn't want to add more new features

.github/workflows/moving-cards.yml
fmulero: It shouldn't be empty. If assignees if empty the step `Assign to a person to work on it` will be executed and this one will be skipped

agomezmoron (approved):
Thanks for this awesome work. LGTM

PR #6969, opened 2022-09-18
[bitnami/matomo] Proxy host headers

JoostVanDerSluis:
### Description of the change

It adds the option to set the proxy_host_headers option in Matomo's option.ini.php configuration file by setting the MATOMO_PROXY_HOST_HEADER environment variable.

### Benefits

This [setting is necessary when Matomo is behind a proxy](https://matomo.org/faq/how-to-install/faq_98/). It could be that it was not needed in the past. But in my setup it was not possible to login to Matomo when it was behind a proxy due to a [origin/nonce-check](https://github.com/matomo-org/matomo/blob/b183ce903f3e24c97ad929c77aff99e1ef93ab11/core/Nonce.php#L136) during login (The error message is about `form security failed`). Without this setting Matomo can not know what the original origin/host is, so the check fails.

### Possible drawbacks

Can not think of any.


aoterolorenzo:
Hey @JoostVanDerSluis, thanks for your contribution. I'm removing `stale` label and marking and marking it as `on-hold` meanwhile reviewing the case.

aoterolorenzo (approved):
LGTM

PR #6978, opened 2022-09-19
[bitnami/mongodb] Fix hardcoded root username for metrics user creation

chess-knight:
### Description of the change

Hardcoded root username `'root'` is used during *MONGODB_METRICS_USERNAME* user creation, so it is impossible to create a metric user when the root username is different. This PR changes this behaviour.

### Benefits

From now it is possible to create a metric user when there is a different root username than `'root'`.
 
### Possible drawbacks

n/a


chess-knight:
> could split this PR into two, one for mongodb and another for mongodb-sharded?

@fmulero it is no problem, thank you for the review. See #6996 for mongodb-sharded.

CeliaGMqrz (approved):

PR #6994, opened 2022-09-19
[bitnami/cassandra] honor for broadcast-address in nodetool-output

stefan-work:
### Description of the change

If CASSANDRA_BROADCAST_ADDRESS was set, nodetool reports that address only. Thus wait_for_nodetool_up() must grep for broadcast-ip else container startup hangs.

### Benefits

You can use config-option CASSANDRA_BROADCAST_ADDRESS

### Possible drawbacks

None (if CASSANDRA_BROADCAST_ADDRESS is unset, behavior will not change)

### DIFF
--- a/bitnami/cassandra/4.0/debian-11/rootfs/opt/bitnami/scripts/libcassandra.sh
+++ b/bitnami/cassandra/4.0/debian-11/rootfs/opt/bitnami/scripts/libcassandra.sh
@@ -839,7 +839,7 @@ wait_for_nodetool_up() {
         # https://community.datastax.com/questions/13764/java-version-for-cassandra-3113.html
         local -r check_cmd=("${CASSANDRA_BIN_DIR}/nodetool" "-Dcom.sun.jndi.rmiURLParsing=legacy")
         local -r check_args=("status" "--port" "$CASSANDRA_JMX_PORT_NUMBER")
-        local -r machine_ip="$(dns_lookup "$CASSANDRA_HOST" "v4")"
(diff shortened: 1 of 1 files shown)


bitnami/cassandra/4.0/debian-11/rootfs/opt/bitnami/scripts/libcassandra.sh
```diff
@@ -855,7 +855,7 @@ wait_for_nodetool_up() {
         # https://community.datastax.com/questions/13764/java-version-for-cassandra-3113.html
         local -r check_cmd=("${CASSANDRA_BIN_DIR}/nodetool" "-Dcom.sun.jndi.rmiURLParsing=legacy")
         local -r check_args=("status" "--port" "$CASSANDRA_JMX_PORT_NUMBER")
-        local -r machine_ip="$(dns_lookup "$CASSANDRA_HOST" "v4")"
+        local -r machine_ip="$(dns_lookup "${CASSANDRA_BROADCAST_ADDRESS:-$CASSANDRA_HOST}" "v4")"
```
aoterolorenzo, line 858: Hey! Thanks for your contribution!

I think that this would also apply to [L842](https://github.com/bitnami/containers/blob/16398787d2b648d2799a8ad4122b3686e7c8e97d/bitnami/cassandra/4.0/debian-11/rootfs/opt/bitnami/scripts/libcassandra.sh#L842)

stefan-work: Makes absolutely sense... ;)
Fixed in  0a6baed

aoterolorenzo (approved):

PR #7074, opened 2022-09-20
[bitnami/*] Fix label issue when verify is already set

fmulero:

### Description of the change

Triggers `get-containers` action on any label event with `verify` or when `verify` is already set

### Benefits

Avoid CI error with label events when verify is already set

### Possible drawbacks

VIB verify could be triggered several times if we set several labels at the same time and `verify` is already set.


agomezmoron (approved):
LGTM

PR #7326, opened 2022-09-21
[bitnami/spark] Update python version from 3.8 to 3.10

rich-java-dev:
update python version from 3.8.14-1 to 3.10.7-1.

- This still uses bitnami's python package, however the upgrade from 3.8 to 3.10 ensures compatibility with pyspark notebooks, which is based on python 3.10.

- https://github.com/jupyter/docker-stacks/tree/main/pyspark-notebook

- This prevents errors associated with spark driver and spark workers on conflicting python minor versions:

roughly:
"Spark Exception: Python in worker has different version 3.8 than that in driver 3.10, PySpark cannot run with different minor versions"


jotamartos:
Hi @rich-java-dev,

Thank you for your contribution. I do not see any problem with this change, the Spark's documentation also mentions that it supports 3.7+ but I understand some plugins may need an updated version. 

https://spark.apache.org/docs/latest/

Let me include this change in our pipeline a make a quick test to ensure there is nothing broken. I'll merge this change and update this PR once I have more information. 

Thanks

rich-java-dev:
Thanks @jotamartos

jotamartos:
Hi again @rich-java-dev,

After evaluating this request, we decided not to upgrade Python in the solution due to multiple reasons: 

- The issue is with an external package and upgrading the solution to Python 3.10 will probably introduce more errors in other external packages that work nowadays.
- The functionalities of the application work with Python 3.8 and the official documentation mentions that it supports Python 3.7+

PR #8190, opened 2022-09-28
[bitnami/postgresql-repmgr] Adds a new environment variable to set failover in repmgr

### LINKED ISSUE #12263: [bitnami/postgresql-ha] Postgresql-HA ends up with multiple primaries on node changes
### Name and Version

bitnami/postgresql-ha

Extremely hard to replicate, but we've seen it often enough now.

Essentially, the root cause is we change our nodes weekly in our cluster. We add one new node, uncordon + drain an old node, then add the new one via Cluster API.

Occasionally, about once a month, one of our databases ends up in a state where 2 postgresql pods from the postgresql-ha chart believe they are primary. This is usually resolves by just picking a pod and restarting it, then the postgresql pods will then re-nominate the primary.

_No response_

Would expect postgresql-ha to not get in this state. Totally understand there is a customXProbe value exposed, but was hoping we could land on a better default.

### What do you see instead?

I believe altering the liveness probe or adding a startup to ensure that *-postgresql-0 is primary and *-postgresql-X is secondary would be best.

I believe this is related and could fix #8062 as well.

Could of ways to implement this:
1) psql -U postgres -c "SELECT pg_is_in_recovery()" is true for the primary and false for all others
2) psql -U postgres -c "SELECT pg_stat_receiver" returns contents if primary and there are secondaries. "SELECT pg_stat_wal_receiver" would return content on all secondaries.

For context, we always check via method 2, next time it occurs I'll be sure to check via 1 to ensure both are good ways to verify.

### Additional information

_No response_

ProNibs:
### Description of the change

This adds a new environment variable called REPMGR_FAILOVER to allow setting of the repmgr.conf file's failover to be automatic or manual.

### Benefits

When using repmgr in Kubernetes, failover should be set to manual as opposed to automatic and attempting to overwrite it results currently requires overwriting the entirety of the librepmgr.sh.

I ran a local docker build command under the postgresql-repmgr and setup the docker-compose to use that image, exec into the container and verify the conf file was setup correctly.

### Applicable issues

This is to setup the pre-requisites of using repmgr in the bitnami/postgresql-ha chart and fix [https://github.com/bitnami/charts/issues/12263](https://github.com/bitnami/charts/issues/12263).


rafariossaa (approved):
Thanks for adding this.
LGTM.

PR #8752, opened 2022-10-03
[bitnami/odoo] support custom addons directory

Yann-J:
### Description of the change

This makes the `ODOO_ADDONS_DIR` variable configurable.

### Benefits

This allows to better manage the custom addons directory, even if the main `/bitnami/odoo` is mapped to its own storage (like the [helm chart](https://artifacthub.io/packages/helm/bitnami/odoo) does). For instance:

- Using a separate persistence volume for these
- Build addons directly into a custom image

### Possible drawbacks

None foreseen since this is optional and backward-compatible.

### Applicable issues

N/A

### Additional information

FYI, this is the workaround we've been using with the [helm chart](https://artifacthub.io/packages/helm/bitnami/odoo) to make our custom addons available to our custom Odoo image:

```
# Create an emptyDir (temp) volume to hold the addons
extraVolumes:
  - name: addons
    emptyDir: {}

# copy the container's addons to the temp volume
initContainers:
  - name: copy-addons
    image: "{{ .Values.image.registry }}/{{ .Values.image.repository }}:{{ .Values.image.tag }}"
    command:
      - sh
      - -c
      - |
        # copy addons to a temporary volume so it can be mounted back later
        cp -r /bitnami/odoo/addons /addons-volume
    volumeMounts:
      - name: addons
        mountPath: /addons-volume

# mount the addons volume to the main container, so they won't be hidden by the /bitnami/odoo volume
extraVolumeMounts:
  - name: addons
    mountPath: /bitnami/odoo/addons
```

PR #8871, opened 2022-10-04
[bitnami/odoo] support custom addons directory as env var

Yann-J:
### Description of the change

This makes the `ODOO_ADDONS_DIR` variable configurable - replaces #8752 

### Benefits

This allows to better manage the custom addons directory, even if the main `/bitnami/odoo` is mapped to its own storage (like the [helm chart](https://artifacthub.io/packages/helm/bitnami/odoo) does). For instance:

- Using a separate persistence volume for these
- Build addons directly into a custom image

### Possible drawbacks

None foreseen since this is optional and backward-compatible.

### Applicable issues

N/A

### Additional information

FYI, this is the workaround we've been using with the [helm chart](https://artifacthub.io/packages/helm/bitnami/odoo) to make our custom addons available to our custom Odoo image:

```
# Create an emptyDir (temp) volume to hold the addons
extraVolumes:
  - name: addons
    emptyDir: {}

# copy the container's addons to the temp volume
initContainers:
  - name: copy-addons
    image: "{{ .Values.image.registry }}/{{ .Values.image.repository }}:{{ .Values.image.tag }}"
    command:
      - sh
      - -c
      - |
        # copy addons to a temporary volume so it can be mounted back later
        cp -r /bitnami/odoo/addons /addons-volume
    volumeMounts:
      - name: addons
        mountPath: /addons-volume

# mount the addons volume to the main container, so they won't be hidden by the /bitnami/odoo volume
extraVolumeMounts:
  - name: addons
    mountPath: /bitnami/odoo/addons
```


Yann-J:
FYI this kind of replicates the behaviour of the [official image](https://github.com/docker-library/docs/tree/master/odoo#mount-custom-addons) except that in that image the location is hardcoded to `/mnt/extra-addons`...

corico44 (approved):
LGTM!

hellozhangwei:
Can this be added to version 16?

hellozhangwei:
Can i have an example about using ODOO_ADDONS_DIR in docker compose file?

corico44:
Hello @hellozhangwei,

Looking closely at the logic of this new variable, it seems to have a persistence problem. I'm going to open a task internally to fix it properly. We will report any update in this PR!

PR #8906, opened 2022-10-04
[bitnami/containers] Fix result check in automatic PRs

fmulero:

### Description of the change

Use the right output to check if the PR can be merged or not. 

### Benefits

Currently if `VIB verify` fails the `auto-pr-review` job won't retrieve that failure because is using the exit code from `verification-summary`

### Possible drawbacks

None

### Additional information

The following log came from previous PR (#8772) which was reverted:

```
::set-output name=result::error
##[debug]steps.get-status.outputs.result='error'

::set-output name=result::
##[debug]steps.get-status.outputs.result=''
##[debug]Node Action run completed with exit code 0
##[debug]Finishing: Check Status
```

`github-script` uses the `result` output by default for any `return` from the script: https://github.com/actions/github-script#reading-step-results

Follow up: #8772

### DIFF
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -115,7 +115,7 @@ jobs:
       - get-containers
       - vib-verify
     outputs:
-      result: ${{ steps.get-status.outputs.state }}
+      result: ${{ steps.get-status.outputs.result }}
     if: ${{ always() }}
     steps:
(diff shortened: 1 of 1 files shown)


joancafom (dismissed):

FraPazGal (approved):
LGTM!

joancafom (approved):

PR #8953, opened 2022-10-05
[bitnami/*] Allow for publish pipeline's branch granularity

FraPazGal:
Signed-off-by: FraPazGal <fdepaz@vmware.com>

### Description of the change

This PR updates the CD logic so it checks for specific `vib-publish.json` pipeline definitions for the modified branch/es in that commit. For each asset's branch modified it will:

- Check if there is a branch's folder under `.vib/asset/branch`:
  - If the folder exists, use `.vib/asset/branch/vib-publish.json` as the DSL path.
  - If it doesn't exist, use the default `.vib/asset/vib-publish.json`.

Given that non-ARM branches are usually the oldest ones, we will create a non-ARM publish pipeline for each of those branches and leave a common publish pipeline supporting ARM for the rest of the asset branches. The assets with branches supporting different architectures are:

- Envoy --> branch 1.20 does not support ARM64
- MariaDB Galera --> branches 10.3 & 10.4 do not support ARM64
- MariaDB --> branches 10.3 & 10.4 do not support ARM64
- Odoo --> branches 13 & 14 do not support ARM64
- Parse --> branch 4 does not support ARM64

### Benefits

Allows for publish pipeline's branch granularity.

### Possible drawbacks

As it is, we are duplicating a lot of code, but this is unavoidable unless we do a major refactor to build our final vib pipelines using `jsonnet` files.

### Additional information

Tested in fork: https://github.com/FraPazGal/containers/actions/runs/3188969818/jobs/5202241895

### DIFF
--- a/.github/workflows/cd-pipeline-harbor.yml
+++ b/.github/workflows/cd-pipeline-harbor.yml
@@ -78,19 +78,29 @@ jobs:
               exit 1
             else
               name="$(echo "${{ matrix.container }}" | awk -F '/' '{print $2}')"
+              branch="$(echo "${{ matrix.container }}" | awk -F '/' '{print $3}')"
               echo "::set-output name=tag::${tag}"
               echo "::set-output name=name::${name}"
+              echo "::set-output name=branch::${branch}"
               echo "::set-output name=result::ok"
             fi
           else
             # Container folder doesn't exists we are assuming a deprecation
             echo "::set-output name=result::skip"
           fi
+      - id: get-dsl-filepath
+        name: Get pipeline DSL filepath
+        run: |
+          dsl_path="${{ steps.get-container-metadata.outputs.name }}"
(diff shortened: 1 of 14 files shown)


aruiz14 (approved):
Great!

joancafom (approved):

fmulero (commented):
Thanks a lot @FraPazGal. 

Only a quick question, is worth to apply same changes to CI pipelines? I mean the option of set a branch level dsl for CI pipeline

fmulero (approved):

FraPazGal:
@fmulero For now I decided not to add that logic to the CI as the logic involving architectures for `vib-verify` pipelines is not in place. 

Having said that, it is true that we will need to replicate these changes for the CI before we start creating test suites for containers. I recall there were a couple of concerns about this logic internally; I'll check with the team and create the task to update the CI/Verify logic if necessary.

PR #9096, opened 2022-10-06
[bitnami/postresql] Removed curl removal

dersvenhesse:
### Description of the change

The latest automated updates (e.g. https://github.com/bitnami/containers/pull/8834) removed curl. Since the removal might be breaking for some users (as noted here https://github.com/bitnami/containers/commit/6158f772fbdc67abc07a4c8a526dc27e182f6809#r85770695), it is hereby removed.

### Benefits

Projects using postresql and curl don't break.

### Possible drawbacks

I can't think of any.

### Applicable issues

  - relates to #8834


carrodher:
Hi, we have been working on removing non-needed dependencies to have more secure containers, reducing the attack surface and the container size. `curl` is not needed in those container images you're mentioning, that's why was removed from the container. You can use the `install_packages` script to install it if it is needed for your use case.

PR #10071, opened 2022-10-13
[bitnami/keycloak] Allow to prepend run flags 

### LINKED ISSUE #1047: [bitnami/keycloak] OpenID Connect Logout
### Name and Version

bitnami/keycloak:latest

1. run docker-compose keycloak with bitnami
2. my project use lib keycloak old and it support keycloak version 10 go back
3. when i logout domain return "keycloak/logout?redirect_uri=xxx"
4. I found solution enable **bin/kc.[sh|bat] --spi-login-protocol-openid-connect-legacy-logout-redirect-uri=true**
5. I read overview on docker hub bitnami but I dont know repair file yml 

image

enable  --spi-login-protocol-openid-connect-legacy-logout-redirect-uri=true in file yml
or something else fix the case

### What do you see instead?

image

### Additional information

solution: 
https://www.keycloak.org/2022/04/keycloak-1800-released#_openid_connect_logout

### LINKED ISSUE #9840: [bitnami/keycloak] Prepended extra startup args
### Name and Version

bitnami/keycloak:19.0.3-debian-11-r1

Keycloak has gone through multiple updates this year. They have changed some behavior that can be only enabled back using special flags (For example, redirect_uri param which was discussed in this [issue](https://github.com/bitnami/containers/issues/1047))

The command with such a flag should look like the following:
```
bin/kc.sh --spi-login-protocol-openid-connect-legacy-logout-redirect-uri=true start-dev
```

The issue is that bitnami keycloak docker image doesn't allow to prepend such flags _before_ the start command.

That's what is used as a start command in the image:
```
start_command=("${KEYCLOAK_BIN_DIR}/kc.sh" "-cf" "$conf_file" "$start_param")
```
We can only _append_ flags after the start command using `KEYCLOAK_EXTRA_ARGS` env variable.

Let's add KEYCLOAK_EXTRA_ARGS_PREPEND env variable or something similar which would allow adding some arbitrary flags before the start command. If it sounds like a sensible proposition, I can create a PR that adds such behaviour.

One could utilize the custom configuration env variable to enable such behavior. However, I think it's not convenient because you lose the benefits of autogenerated configs which you have to maintain now. The new env variable sounds like an elegant solution for flags that enable backward-compatible behavior that will be removed anyway after a few keycloak upgrades.

Tasyp:
### Description of the change

Adds KEYCLOAK_EXTRA_ARGS_PREPEND env variable

### Benefits

It allows using flags that need to be applied before the `start` command

### Possible drawbacks

This change shouldn't affect existing users in any way so I don't see real drawbacks.

### Applicable issues

  - fixes #9840
  - fixes #1047


Mauraza (approved):
Hi @Tasyp,

Thanks for the contribution!! 
LGTM. In the next revision of the container will be appearing this change!

PR #10098, opened 2022-10-13
[bitnami/schema-registry] enable HTTPS schema-registry usage by adding ssl.key.password

### LINKED ISSUE #9819: [bitnami/schema-registry] Keystore can`t be used for HTTPS Listener
### Name and Version

bitnami/schema-registry:7.2.2

Execute the docker-compose file with the related keystore.jks file. 

[schema-registry.zip](https://github.com/bitnami/containers/files/9758282/schema-registry.zip)

schema-registry starts up and can be reached via https://localhost:8443

### What do you see instead?

Keystore could not be used because of a missing key password.
The attached log prints the occurred exception.
[docker-compose.log](https://github.com/bitnami/containers/files/9758389/docker-compose.log)

### Additional information

The follogwing i tried:
- set the same key password than for keystore password
- set no key password
- set SCHEMA_REGISTRY_SSL_KEY_PASSWORD as env
None of them works.

It looks like [libschemaregistry.sh](https://github.com/bitnami/containers/blob/main/bitnami/schema-registry/7.2/debian-11/rootfs/opt/bitnami/scripts/libschemaregistry.sh) does not set the value for SCHEMA_REGISTRY_SSL_KEY_PASSWORD. By temporarily adding 
` [[ -n "$SCHEMA_REGISTRY_SSL_KEY_PASSWORD" ]] && schema_registry_conf_set "ssl.key.password" "$SCHEMA_REGISTRY_SSL_KEY_PASSWORD"` the keystore could be used with SCHEMA_REGISTRY_SSL_KEY_PASSWORD env and the HTTPS listener worked.

b3rnh8rd:
Signed-off-by: b3rnh8rd <bernhard.haring@gmx.at>
To use schema-registry via HTTPS the keystore key password is mapped in addition to keystore password.

Fixes https://github.com/bitnami/containers/issues/9819


marcosbc (approved):
LGTM! Thanks for the contribution!

PR #10736, opened 2022-10-19
[bitnami/jasperreports] fix: file corruption of attachments with special characters.

schakko:
### Description of the change

When UTF-8 special characters are included in output file name while scheduling, it might lead to file being corrupt in the attachment. See
https://community.jaspersoft.com/wiki/when-utf-8-special-characters-are-included-output-file-name-while-scheduling-it-might-lead-file.

This commit appends the `mail.mime.splitlongparameters=false` parameter to `JAVA_OPTS`.

### Benefits

- Allow sending of attachments like `EinnahmenÜberschussrechnung.pdf`

### Possible drawbacks

Not known.

### Applicable issues
  - fixes https://community.jaspersoft.com/wiki/when-utf-8-special-characters-are-included-output-file-name-while-scheduling-it-might-lead-file

### DIFF
--- a/bitnami/jasperreports/8/debian-11/rootfs/opt/bitnami/scripts/tomcat-env.sh
+++ b/bitnami/jasperreports/8/debian-11/rootfs/opt/bitnami/scripts/tomcat-env.sh
@@ -83,7 +83,9 @@ export TOMCAT_INSTALL_DEFAULT_WEBAPPS="${TOMCAT_INSTALL_DEFAULT_WEBAPPS:-yes}"
 
 # Default JVM configuration
 export JAVA_HOME="${BITNAMI_ROOT_DIR}/java"
-export JAVA_OPTS="${JAVA_OPTS:--Djava.awt.headless=true -XX:+UseG1GC -Dfile.encoding=UTF-8 -Duser.home=${TOMCAT_HOME}}"
+
+# @see https://community.jaspersoft.com/wiki/when-utf-8-special-characters-are-included-output-file-name-while-scheduling-it-might-lead-file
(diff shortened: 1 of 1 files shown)


Mauraza (commented):
Hi @schakko,

Thanks for the contribution! but this is a change in tomcat. 
Could you try to use [TOMCAT_EXTRA_JAVA_OPTS](ttps://github.com/bitnami/containers/tree/main/bitnami/tomcat#environment-variables)?
`TOMCAT_EXTRA_JAVA_OPTS=-Dmail.mime.splitlongparameters=false`

schakko:
@Mauraza I've just seen that a newer version of the official [Helm chart](https://github.com/bitnami/charts/tree/main/bitnami/jasperreports) supports `extraEnvVars`. That allows us to set `TOMCAT_EXTRA_JAVA_OPTS` from the outside.

Do we want to set the default argument inside `tomcat-env.sh` anyway? The problem with the special characters in the attachments occurs in all non-ASCII character sets. We could change

```bash
export TOMCAT_EXTRA_JAVA_OPTS="${TOMCAT_EXTRA_JAVA_OPTS:-}"
```

to

```bash
export TOMCAT_EXTRA_JAVA_OPTS="${TOMCAT_EXTRA_JAVA_OPTS:--Dmail.mime.splitlongparameters=false}"
```

Mauraza:
Hi @schakko,

we prefer no mix the code Tomcat with changes related to JasperReports so we can't accept the contribution, but you can use `TOMCAT_EXTRA_JAVA_OPTS`.

https://github.com/bitnami/containers/blob/ba420384ef07a6e93b692350a5e656bc4e664ffb/bitnami/jasperreports/8/debian-11/rootfs/opt/bitnami/scripts/libtomcat.sh#L148-L149
 

```bash
  extraEnvVars:
  - name: TOMCAT_EXTRA_JAVA_OPTS
    value: "-Dmail.mime.splitlongparameters=false"
```

PR #11389, opened 2022-10-25
[bitnami/containers] Add OSM reporting action (first batch)

fmulero:

### Description of the change

Include OSM reporting action in CD pipelines to report dependencies and OS licenses for following assets:
* envoy
* java
* kafka
* ksql
* kubectl
* mariadb
* mariadb-galera
* mysql
* nginx
* node
* postgresql
* postgresql-repmgr
* python
* ruby
* schema-registry
* symfony
* tomcat

### Benefits

Report dependencies and licenses for each asset in an automated way.

### Possible drawbacks

Reporting is following a different approach. With this change we will start reporting the packages and licenses for each release instead of for the whole catalogue.


rafariossaa (dismissed):
LGTM

gongomgra (approved):
LGTM

PR #11542, opened 2022-10-27
Fix unassigned SPARK_EXECUTOR_JAVA_OPTS

### LINKED ISSUE #11198: [bitnami/spark] SPARK_JAVA_OPTS variable is unassigned in entrypoint.sh script
### Name and Version

bitnami/spark:3.3.0

Run a spark job with --master k8s:// and Java 17 instead of Java8/11. We replace base java component in the image using the same approach as in Dockerfile.

This happens because in `entrypoint.sh` of `apache/spark` there is an additional line which is missing from bitnami script:

```
env | grep SPARK_JAVA_OPT_ | sort -t_ -k4 -n | sed 's/[^=]*=\(.*\)/\1/g' > /tmp/java_opts.txt
readarray -t SPARK_EXECUTOR_JAVA_OPTS < /tmp/java_opts.txt
```

while in bitnami script `SPARK_EXECUTOR_JAVA_OPTS` is used right away, while Spark driver creates no such variable.

You can also run the job with bitnami/spark:3.3.0 with Java 8 and notice that spark jvm doesn't have any custom options applied in k8s mode.

Spark job runs w/o issues with custom java options applied

### What do you see instead?
With Java 17:
```
Exception in thread "main" java.lang.IllegalAccessError: class org.apache.spark.storage.StorageUtils$ (in unnamed module @0x777c111) cannot access class sun.nio.ch.DirectBuffer (in module java.base) because module java.base does not export sun.nio.ch to unnamed module @0x777c111
	at org.apache.spark.storage.StorageUtils$.<init>(StorageUtils.scala:213)
	at org.apache.spark.storage.StorageUtils$.<clinit>(StorageUtils.scala)
	at org.apache.spark.storage.BlockManager.<init>(BlockManager.scala:222)
```
With Java11/8: no custom jvm options applied.
### Additional information

_No response_

george-zubrienko:
Signed-off-by: George Zubrienko <GZU@ecco.com>

### Description of the change

Fixes unassigned variable reference in spark 3.3 entrypoint.sh: `SPARK_EXECUTOR_JAVA_OPTS`

### Benefits

Spark with `k8s:/` master is able to use `spark.executor.extraJavaOptions`. This is also a mandatory change to make Java 17-based Spark work with k8s master.

### Possible drawbacks

N/A

### Applicable issues

  - fixes #11198

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,2 @@
+# IntelliJ configurations
+.idea/
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,2 +1,0 @@
-# IntelliJ configurations
-.idea/
--- a/bitnami/spark/3.3/debian-11/rootfs/opt/bitnami/scripts/spark/entrypoint.sh
+++ b/bitnami/spark/3.3/debian-11/rootfs/opt/bitnami/scripts/spark/entrypoint.sh
@@ -43,6 +43,14 @@ case "$1" in
     ;;
   executor)
     shift 1
+
+    set +o pipefail
+
+    env | grep SPARK_JAVA_OPT_ | sort -t_ -k4 -n | sed 's/[^=]*=\(.*\)/\1/g' > /tmp/java_opts.txt
+    readarray -t SPARK_EXECUTOR_JAVA_OPTS < /tmp/java_opts.txt
+
+    set -o pipefail
+
     CMD=(
       ${JAVA_HOME}/bin/java
       "${SPARK_EXECUTOR_JAVA_OPTS[@]}"


.idea/.gitignore
```diff
@@ -0,0 +1,2 @@
+# IntelliJ configurations
```
carrodher, line 1: It seems this file was added by mistake, could you please remove it? Thanks

george-zubrienko: Sure np. But it would be nice if the repo had global `.gitignore` because it is quite annoying to work w/o it when using an IDE.

rafariossaa (changes requested):
Could you add this env var to the README ?

george-zubrienko:
> Could you add this env var to the README ?

Why this should be part of README? This is not even an env var. `SPARK_JAVA_OPT_0`, `SPARK_JAVA_OPT_1` etc are env vars generated by Spark driver for its executors when `spark.executor.extraJavaOptions` is specified. This is not something a user can/should provide directly.

rafariossaa (commented):
What about using `cut` instead of `sed` ? I think it would be easier to read that the regex substitution used in `sed`.
```
env | grep SPARK_JAVA_OPT_ | sort -t_ -k4 -n | cut -d= -f2- > /tmp/java_opts.txt
```

george-zubrienko:
> What about using `cut` instead of `sed` ? I think it would be easier to read that the regex substitution used in `sed`.
> 
> ```
> env | grep SPARK_JAVA_OPT_ | sort -t_ -k4 -n | cut -d= -f2- > /tmp/java_opts.txt
> ```

I don't mind changing it, but in the upstream repo they use `sed` and I'm usually very hesitant to change anything like this unless there is an obvious issue/error happening:

https://github.com/apache/spark/blob/v3.3.1/resource-managers/kubernetes/docker/src/main/dockerfiles/spark/entrypoint.sh#L43-L45

Also java options provided might have `=` in their values:

```
 echo "SPARK_JAVA_OPT_0=--add-opens java.base/sun.ch=ALL_UNNAMED" | grep SPARK_JAVA_OPT_ | sort -t_ -k4 -n | cut -d= -f2
```
produces `--add-opens java.base/sun.ch` - incorrect.

While code from upstream works as intended:

```
echo "SPARK_JAVA_OPT_0=--add-opens java.base/sun.ch=ALL_UNNAMED" | sort -t_ -k4 -n | sed 's/[^=]*=\(.*\)/\1/g'
```

produces `--add-opens java.base/sun.ch=ALL_UNNAMED` as intended.

rafariossaa:
Hi,
  You missed the trailing `-` in the `-f2-` command in your test:
```bash
$ echo "SPARK_JAVA_OPT_0=--add-opens java.base/sun.ch=ALL_UNNAMED" | grep SPARK_JAVA_OPT_ | sort -t_ -k4 -n | cut -d= -f2
--add-opens java.base/sun.ch


$ echo "SPARK_JAVA_OPT_0=--add-opens java.base/sun.ch=ALL_UNNAMED" | grep SPARK_JAVA_OPT_ | sort -t_ -k4 -n | cut -d= -f2-
--add-opens java.base/sun.ch=ALL_UNNAMED
```
  but anyway, I agree we should keep what is upstream just in case.

rafariossaa (approved):
LGTM

PR #12535, opened 2022-11-04
[bitnami/dokuwiki] fix DOKU_INC path

kossmac:
### Description of the change

Sets auto_prepend_file in php.ini to let DokuWiki's templates and plugins know where DokuWiki really lives(default: "/opt/bitnami/dokuwiki") Since this container splits application from state files(/opt/bitnami/dokuwiki vs. /bitnami/dokuwiki) it is important that the DOKU_INC path is set accordingly. The PHP interpreter will prepend the file content specified in auto_prepend_file to every php file that gets executed.

### Benefits

Templates and plugins relying on the DOKU_INC path will work now(e.g. farmer plugin).

### Possible drawbacks

nothing known

### Additional information

As discussed with DokuWiki founder @splitbrain this is the best solution to get all templates and plugins to work.

### DIFF
--- a/bitnami/dokuwiki/20220731/debian-11/rootfs/opt/bitnami/scripts/dokuwiki/postunpack.sh
+++ b/bitnami/dokuwiki/20220731/debian-11/rootfs/opt/bitnami/scripts/dokuwiki/postunpack.sh
@@ -39,7 +39,7 @@ php_conf_set memory_limit "$PHP_DEFAULT_MEMORY_LIMIT"
 
 # Fix DOKU_INC, since we split application from state, DokuWiki's plugins and templates need to know where they live
 info "Fix DOKU_INC variable"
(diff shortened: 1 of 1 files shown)


dgomezleon:
Hi @kossmac and @splitbrain.

Thanks for contributing. I have a question about this approach:

Could it be possible to add this logic to `doku.php`? Note this can be problematic since `/tmp` is cleaned up after every restart.

dgomezleon (changes requested):
Great contribution.  Regarding [this](https://github.com/bitnami/containers/pull/12535#issuecomment-1309995942), I would use `dokuwiki/conf` instead, since it is persisted.

kossmac:
@dgomezleon you're right, i moved the file to dokuwiki/conf now.

dgomezleon (approved):
LGTM

PR #13146, opened 2022-11-09
[bitnami/clickhouse] listen on ipv4 and ipv6 address

LuckySB:
### Description of the change

change clickhouse command line options, hardcoded in Dockerfile CMD to
`listen_host=::`

Clickhouse listen on all ipv4 and ipv6 adresses

### Benefits

This image work on AWS ipv6-only EKS (amazon managment kubernetes)

### DIFF
--- a/bitnami/clickhouse/22.3/debian-11/Dockerfile
+++ b/bitnami/clickhouse/22.3/debian-11/Dockerfile
@@ -49,4 +49,4 @@ EXPOSE 8123 9000 9004 9005 9009
 
 USER 1001
 ENTRYPOINT [ "/opt/bitnami/scripts/clickhouse/entrypoint.sh" ]
-CMD [ "/opt/bitnami/scripts/clickhouse/run.sh", "--", "--listen_host=0.0.0.0" ]
+CMD [ "/opt/bitnami/scripts/clickhouse/run.sh", "--", "--listen_host=::" ]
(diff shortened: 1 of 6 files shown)


fmulero (changes requested):
Hi Sergei, thank you su much for your contribution.

Only one concern, now this image doesn't listen in IPv6 addresses, but with this change the image will fail if IPv6 is not enabled in the server, please see https://github.com/ClickHouse/ClickHouse/issues/20965. Could you make it configurable?

LuckySB:
> Hi Sergei, thank you su much for your contribution.
> 
> Only one concern, now this image doesn't listen in IPv6 addresses, but with this change the image will fail if IPv6 is not enabled in the server, please see [ClickHouse/ClickHouse#20965](https://github.com/ClickHouse/ClickHouse/issues/20965). Could you make it configurable?

https://github.com/ClickHouse/ClickHouse/issues/20965#issuecomment-782092930
It's not an error.

Address family for hostname not supported for IPv6 address means that IPv6 is disabled on your server.
The server will print warning and continue to work if it is able to listen at least one of the addresses.
No actions required.

fmulero:
Hi Sergei,

This is not an error in the product (ClickHouse) because you can configure this feature to fit with your requirements. 
These changes in the docker image is forcing to use IPv6 and breaking current behaviour for some deployments.

If you are using our chart to deploy in kubernetes, you can set the `args` the default CMD in the image:
```
helm install clickhouse bitnami/clickhouse --set args="{/opt/bitnami/scripts/clickhouse/run.sh,--,--listen_host=::}"
```

PR #14144, opened 2022-11-18
Remove SPARK_MOUNTED_CONFDIR and SPARK_VOLUME_DIR

### LINKED ISSUE #11199: Incorrect SPARK_VOLUMEDIR export
### Name and Version

bitnami/spark:3.3.0

Running spark you get this message as a warning:
```
15:18:13.65 
 15:18:13.65 Welcome to the Bitnami spark container
 15:18:13.65 Subscribe to project updates by watching https://github.com/bitnami/containers
 15:18:13.65 Submit issues and feature requests at https://github.com/bitnami/containers/issues
 15:18:13.65 
 15:18:13.65 INFO  ==> ** Starting Spark setup **
realpath: /bitnami/spark/conf: No such file or directory
```

This is due to incorrect path `/bitnami/spark` instead of `/opt/bitnami/spark` in SPARK_VOLUMEDIR

https://github.com/bitnami/containers/blob/main/bitnami/spark/3.3/debian-11/rootfs/opt/bitnami/scripts/libspark.sh#L28-L32

No errors are printed on startup.

### What do you see instead?

```
realpath: /bitnami/spark/conf: No such file or directory
```

### Additional information

_No response_

george-zubrienko:
Signed-off-by: George Zubrienko <GZU@ecco.com>

### Description of the change

Removes unused SPARK_VOLUME_DIR and associated code from a startup sequence.

### Benefits

Currently there are errors printed on Spark startup like `realpath: /bitnami/spark/conf: No such file or directory` which are misleading and might propagate to real errors in the future, while not being beneficial at all.

Spark configurations are mounted by replacing the default file as described in the readme:

```
$ docker run --name spark -v /path/to/spark-defaults.conf:/opt/bitnami/spark/conf/spark-defaults.conf bitnami/spark:latest
```

Same thing happens in Bitnami helm chart - spark-defaults.conf is mounted from configmap and there is no need to mount-then-copy it approach.

### Possible drawbacks

N/A

### Applicable issues

  - fixes #11199

### DIFF
--- a/bitnami/spark/3.3/debian-11/rootfs/opt/bitnami/scripts/libspark.sh
+++ b/bitnami/spark/3.3/debian-11/rootfs/opt/bitnami/scripts/libspark.sh
@@ -25,9 +25,7 @@ spark_env() {
     cat <<"EOF"
 # Spark directories
 export SPARK_BASEDIR="/opt/bitnami/spark"
-export SPARK_VOLUMEDIR="/bitnami/spark"
 export SPARK_CONFDIR="${SPARK_BASEDIR}/conf"
-export SPARK_MOUNTED_CONFDIR="${SPARK_VOLUMEDIR}/conf"
 export SPARK_WORKDIR="${SPARK_BASEDIR}/work"
 export SPARK_CONF_FILE="${SPARK_CONFDIR}/spark-defaults.conf"
 export SPARK_LOGDIR="${SPARK_BASEDIR}/logs"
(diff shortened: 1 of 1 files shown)


gongomgra:
Hi @george-zubrienko,

Thanks for using Bitnami and for your contribution. My main concern with your changes is that there must be other users using this functionality, which we will break. I think it is better to add an additional check to the `is_dir_empty` method in the `libfs.sh` library to avoid running the `realpath` command in case the directory doesn't exist. What do you think?

george-zubrienko:
> Hi @george-zubrienko,
> 
> Thanks for using Bitnami and for your contribution. My main concern with your changes is that there must be other users using this functionality, which we will break. I think it is better to add an additional check to the `is_dir_empty` method in the `libfs.sh` library to avoid running the `realpath` command in case the directory doesn't exist. What do you think?

I would 100% agree if I could find any trace where this functionality is actually used. I don't see this variable mentioned in any readmes and Bitnami's official Spark helm chart uses a completely difference approach for configuration loads. If you still think we should keep it, let me know, I'll update with the check. However in this case this dir simply doesn't exists, because there is no `/bitnami` path in this container.

gongomgra:
Hi @george-zubrienko,

Sorry for the delay. I checked your pull-request with other team members and they agree removing the method shouldn't have major drawbacks, so I'm accepting your changes and merging them. Thanks for your contribution!

gongomgra (approved):

PR #15986, opened 2022-12-05
[bitnami/redis] - Include overrides config once only

philippetrepanier:
Fixes issue where ACL in overrides file would result in an error:

`Duplicate user found. A user can only be defined once in config files`

This is cause by the behaviour of the `redis_append_include_conf` method in `libredis.sh` that only appends the overrides file to the end of the config. When restarting the container or the host the method would be ran again adding another include statement.

### Description of the change

Include overrides config only once in the base `redis.conf` configuration. Clean out stale includes that could be present also.

### Benefits

Able to use the overrides file to store ACL configuration.

### Possible drawbacks

None: if specifying a custom `redis.conf` the overrides logic is not executed. Will only be executed with the default configuration which doesn't have any include in it

### Applicable issues

N/A


mdhont (approved):
Lgtm. Thanks @philippetrepanier!

PR #16567, opened 2022-12-12
[bitnami/redmine] Feature: Add bundle post install call

c23omega:
### Description of the change

Some Redmine plugins (e.g. from RedmineUp) require some additional rubygems. They are right now not installed, which leads to a crash of the container. My change adds a function to post-install them during container startup.

### Benefits

Even with redmine plugins that require additional gems the container is able to operate.

### Possible drawbacks

I don't see any.

### Applicable issues

  - fixes chart issue https://github.com/bitnami/charts/issues/13863

### DIFF
--- a/bitnami/redmine/5/debian-11/rootfs/opt/bitnami/scripts/libredmine.sh
+++ b/bitnami/redmine/5/debian-11/rootfs/opt/bitnami/scripts/libredmine.sh
@@ -459,3 +459,20 @@ redmine_migrate_database() {
     redmine_rake_execute "db:migrate" >/dev/null 2>&1
     redmine_rake_execute "redmine:plugins:migrate" >/dev/null 2>&1
 }
+
+########################
+# Install additional Rubygems if needed
+# Plugins might require additional plugins, which are not shipped with the default
+# bundle. In order to keep Redmine operative, we need to install them additionally.
+# Globals:
+#   REDMINE_*
+# Arguments:
+#   None
+# Returns:
+#   None
+#########################
+redmine_ensure_gems() {
+    cd "$REDMINE_BASE_DIR" || false
+    bundle config unset deployment
+    bundle install
+}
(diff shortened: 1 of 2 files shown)


Mauraza:
Hi @c23omega,

I'm going to do tests. After that, I will update this thread.

Mauraza:
Hi @c23omega,

Did you try to create a new container from the bitnami/redmine with the gems you need?
I think that is a better solution that executing again the bundle.

c23omega:
Hello @Mauraza,

i've consiered that, but the downside of this approach is that would need to maintain a fork and upgrade it manually. Also it is fixed for specific plugins. Also others canot really benefit from it, except of switching to that very specific image.

Some Redmine plugins are rather complex and they might require additional gems more often. In the default case no gems are added, so this is fine. Only in the case where plugins require additional gems, these gems are added. Whenever Redmine is used productively, additional plugins might be added frequently, that might require then such gems. This is a design case of Redmine, since plugins can ship their own Gemfile, whose contents are then considered as relevant for the package. The original Gemfile doesn't need to be touched. All in all this is in my opinion a valuable community contribution. 

Another point to consider, debugging this why the container doesn't start up creates some headache to really figure out that the point are missing additional gems out of the plugins. This point then also doesn't matter anymore, since it makes container debugging easier.

Since plugins are essential for extension of Redmine in various ways, the container should provide a proper plugin support. This includes not only adding potentially additional gems, but also adding the plugin assets to the rails assets path. Similar to possible plugin database migrations. There are rake commands which do that quite reliably. In general, Redmine comes with a very good plugin support (see gems, assets path and db migration), and this is why this should also be reflected by the environments where Redmine works on. DB migrations might be fine to do manually, however since the regular db migrations are also ran there it would be consistent to also run the plugin db migrations automatically. DB migrations don't come frequently, this might be another acceptable step to do it manually for now. The assets however also prevent Redmine from starting up nicely. Once the bundle is fixed and the main parts work,  I'd like to provide another PR for adding the assets, and at some point maybe also db migrations.

Mauraza:
Hi @c23omega,

I was talking to create a Dockerfile using the image of bitnami, then you can add all the changes you want. 
For example, using a script.
```
FROM docker/redmine:5
COPY script.sh /scripts
...
```

c23omega:
Hi @Mauraza,

yes i got that idea, however still your container misses this part pf plugin support. Plugin support is something that affects a lot of users.

Mauraza:
Hi @c23omega,

I think you can create a script to install the plugins, or Am I wrong?

PR #17163, opened 2022-12-16
[bitnami/pgpool] Rename configuration parameter socket_dir to unix_socket_directories

iYUYUE:
### Description of the change

Rename configuration parameter socket_dir to [unix_socket_directories](https://www.pgpool.net/docs/latest/en/html/runtime-config-connection.html#GUC-UNIX-SOCKET-DIRECTORIES) as mentioned in the [4.4.0 release note](https://www.pgpool.net/docs/latest/en/html/release-4-4-0.html). Otherwise, the health check script fails with the following message.

```
psql: error: connection to server on socket "/opt/bitnami/pgpool/tmp/.s.PGSQL.5432" failed: No such file or directory
        Is the server running locally and accepting connections on that socket?
```

### Benefits

Set the right directory where the UNIX domain socket(s) accepting connections for Pgpool-II will be created.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A


iYUYUE:
Forgot to sign off the commit. Will open another one.

PR #17563, opened 2022-12-20
[bitnami/symfony] Deprecate branches reaching its EOL: 4.4, 6.0 and 6.1

carrodher:
### Description of the change
According to https://symfony.com/releases, we can deprecate the following versions: 4.4.x (Nov '22), 6.0.x (Jan '23) and 6.1.x (Jan '23).


migruiz4 (approved):
LGTM

PR #18507, opened 2022-12-29
Update keycloak/docker-compose.yml

fluidum:
I had issue running this keycloak/docker-compose.yml configuration. It is probably small mistake, because default KC configuration file doesn't have port 80 configured as http-port value, it is 8080. After fixing it, it started working. It is also mentioned in oneliner https://www.keycloak.org/getting-started/getting-started-docker

```
keycloak@7aced9861fde:/opt/bitnami/keycloak/conf$ grep port *
keycloak.conf:http-port = 8080
keycloak.conf:https-port = 8443
```

Signed-off-by: noraisk <noraisk@users.noreply.github.com>

### Description of the change

Specified resource service port and protocol.

### Benefits

Less time consuming for other engineers.

### Possible drawbacks

Unknown.

### Applicable issues

Didn't find from open issues.

### DIFF
--- a/bitnami/keycloak/docker-compose.yml
+++ b/bitnami/keycloak/docker-compose.yml
@@ -15,7 +15,7 @@ services:
     depends_on:
       - postgresql
     ports:
-      - "80:8080"
+      - "8080:8080/tcp"
 
 volumes:
   postgresql_data:


rafariossaa:
Hi,
  Which issues are you noticing ?
  I used the default `docker-compose.yaml` and I could log in without issues.
Screenshot_20230109_104646

PR #18547, opened 2022-12-30
[bitnami/mariadb] Change default charset to utf8mb4

Menelion:
### Description of the change

Changed the default character set for MariaDB from 3-byte `utf8` to 4-byte `utf8mb4` which is the recommended default for both MySQL and MariaDB nowadays. Also changed the default collation to `utf8mb4_unicode_ci` for better sorting.

### Benefits

With this change, any 4-byte Unicode characters like emojis and other recently introduced characters will appear correctly out of the box and not as question marks or squares. This will also benefit all MariaDB-powered installations like WordPress, Redmine etc.

### Possible drawbacks

I don't know about any limitations. The collation change, according to [this](https://stackoverflow.com/a/766996/1604592) very detailed StackOverflow answer, can lead to slight performance degradation. However, as CPU’s are fast enough in our times, I find sorting benefits far more important.

### Applicable issues

N/A

### Additional information

I applied the change to all MariaDB versions, from 10.3 to 10.10. If this PR is merged, I'll apply the same change to mariadb-galera.

### DIFF
--- a/bitnami/mariadb/10.10/debian-11/rootfs/opt/bitnami/scripts/mariadb-env.sh
+++ b/bitnami/mariadb/10.10/debian-11/rootfs/opt/bitnami/scripts/mariadb-env.sh
@@ -91,9 +91,9 @@ export DB_DAEMON_GROUP="mysql"
 # Default configuration (build-time)
 export MARIADB_DEFAULT_PORT_NUMBER="3306"
 export DB_DEFAULT_PORT_NUMBER="$MARIADB_DEFAULT_PORT_NUMBER" # only used at build time
-export MARIADB_DEFAULT_CHARACTER_SET="utf8"
+export MARIADB_DEFAULT_CHARACTER_SET="utf8mb4"
 export DB_DEFAULT_CHARACTER_SET="$MARIADB_DEFAULT_CHARACTER_SET" # only used at build time
-export MARIADB_DEFAULT_COLLATE="utf8_general_ci"
+export MARIADB_DEFAULT_COLLATE="utf8mb4_unicode_ci"
 export DB_DEFAULT_COLLATE="$MARIADB_DEFAULT_COLLATE" # only used at build time
 export MARIADB_DEFAULT_BIND_ADDRESS="0.0.0.0"
 export DB_DEFAULT_BIND_ADDRESS="$MARIADB_DEFAULT_BIND_ADDRESS" # only used at build time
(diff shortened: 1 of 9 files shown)


fmulero (commented):
Hi @Menelion Thank you so much for contributing.

I've been searching information about this topic and I have some concerns. For versions higher or equal to 10.6, `utf8` character set and the `utf8_general_ci` collation are alias for `utf8mb3` by default, but this behaviour can be changed to use `utf8mb4` instead. See [this KB](https://mariadb.com/kb/en/unicode/) for more information. I am not sure if changing that behaviour would be a better solution. Anyway I would check the impact of this solution and how it would behave in an upgrade, Have you done any test about that?

Menelion:
@fmulero Yes, I know that. But if you change `utf8` to be aliased from mb3 to mb4, and I specify mb4 directly, why should it impact anything? It would impact only those who had `utf8` (for mb3) before, then upgraded, then decided to go for new defaults (`utf8mb4`). But again, I don't see any "impact" per se, as I did convert several databases from mb3 to mb4 and saw only benefits and no degradation.

fmulero:
Hi Andre, we are testing your changes. About:

> utf8mb4 which is the recommended default for both MySQL and MariaDB nowadays.

Could you provide any link to oficial documentation with that information?

Menelion:
@fmulero Yes, sure (_italics_ are mine):

> The utf8mb3 character set is deprecated and you should expect it to be removed in a future MySQL release. Please use utf8mb4 instead. utf8 is currently an alias for utf8mb3, but it is now deprecated as such, and utf8 is expected subsequently to become a reference to utf8mb4. Beginning with MySQL 8.0.28, utf8mb3 is also displayed in place of utf8 in columns of Information Schema tables, and in the output of SQL SHOW statements.
>
> _To avoid ambiguity about the meaning of utf8, consider specifying utf8mb4 explicitly for character set references._

From [here](https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-sets.html).

Maria is more vague but even more confusing:

> MariaDB 10.6.1 changed the utf8  character set by default to be an alias for utf8mb3 rather than the other way around. It can be set to imply utf8mb4  by changing the value of the old_mode system variable. 

As far as I can see in the comments, the "old mode" will also eventually be removed for Maria.

fmulero (approved):
LGTM! @Menelion Don't worry about mariadb-galera or mysql, this change will be applied also in while by our automations.

Thanks for your contribution

PR #19124, opened 2023-01-04
[bitnami/pgpool] extend health check configuration

### LINKED ISSUE #12263: [bitnami/postgresql-ha] Postgresql-HA ends up with multiple primaries on node changes

pcerny:
### Description of the change

Introduce the following changes:
- Introduce PGPOOL_CONNECT_TIMEOUT variable and set default to 10000ms.
  User can setup pgpool to fail faster when trying to connect to unreachable
  PG node.
- Introduce PGPOOL_HEALTH_CHECK_PSQL_TIMEOUT variable with default to 15.
  User can control "show pool_nodes" command timeout in healthcheck
  function.
- Only nodes in pgpool state DOWN which are UP in reality are attached.

### Benefits

Makes pgpool pod configuration more versatile. It is used by PR https://github.com/bitnami/charts/pull/14175

### Possible drawbacks

Do not know.

### Applicable issues

It fixes https://github.com/bitnami/charts/issues/12263.

### Additional information

N/A

### DIFF
--- a/bitnami/pgpool/4/debian-11/rootfs/opt/bitnami/scripts/libpgpool.sh
+++ b/bitnami/pgpool/4/debian-11/rootfs/opt/bitnami/scripts/libpgpool.sh
@@ -75,6 +75,8 @@ export PGPOOL_HEALTH_CHECK_PERIOD="${PGPOOL_HEALTH_CHECK_PERIOD:-30}"
 export PGPOOL_HEALTH_CHECK_TIMEOUT="${PGPOOL_HEALTH_CHECK_TIMEOUT:-10}"
 export PGPOOL_HEALTH_CHECK_MAX_RETRIES="${PGPOOL_HEALTH_CHECK_MAX_RETRIES:-5}"
 export PGPOOL_HEALTH_CHECK_RETRY_DELAY="${PGPOOL_HEALTH_CHECK_RETRY_DELAY:-5}"
+export PGPOOL_CONNECT_TIMEOUT="${PGPOOL_CONNECT_TIMEOUT:-10000}"
+export PGPOOL_HEALTH_CHECK_PSQL_TIMEOUT="${PGPOOL_HEALTH_CHECK_PSQL_TIMEOUT:-15}"
 export PGPOOL_POSTGRES_CUSTOM_USERS="${PGPOOL_POSTGRES_CUSTOM_USERS:-}"
 export PGPOOL_POSTGRES_CUSTOM_PASSWORDS="${PGPOOL_POSTGRES_CUSTOM_PASSWORDS:-}"
 export PGPOOL_AUTO_FAILBACK="${PGPOOL_AUTO_FAILBACK:-no}"
@@ -209,7 +211,7 @@ pgpool_validate() {
             print_validation_error "The values allowed for $yn are: yes or no"
         fi
     done
-    local positive_values=("PGPOOL_NUM_INIT_CHILDREN" "PGPOOL_MAX_POOL" "PGPOOL_CHILD_MAX_CONNECTIONS" "PGPOOL_CHILD_LIFE_TIME" "PGPOOL_CONNECTION_LIFE_TIME" "PGPOOL_CLIENT_IDLE_LIMIT" "PGPOOL_HEALTH_CHECK_PERIOD" "PGPOOL_HEALTH_CHECK_TIMEOUT" "PGPOOL_HEALTH_CHECK_MAX_RETRIES" "PGPOOL_HEALTH_CHECK_RETRY_DELAY" "PGPOOL_RESERVED_CONNECTIONS")
+    local positive_values=("PGPOOL_NUM_INIT_CHILDREN" "PGPOOL_MAX_POOL" "PGPOOL_CHILD_MAX_CONNECTIONS" "PGPOOL_CHILD_LIFE_TIME" "PGPOOL_CONNECTION_LIFE_TIME" "PGPOOL_CLIENT_IDLE_LIMIT" "PGPOOL_HEALTH_CHECK_PERIOD" "PGPOOL_HEALTH_CHECK_TIMEOUT" "PGPOOL_HEALTH_CHECK_MAX_RETRIES" "PGPOOL_HEALTH_CHECK_RETRY_DELAY" "PGPOOL_RESERVED_CONNECTIONS" "PGPOOL_CONNECT_TIMEOUT" "PGPOOL_HEALTH_CHECK_PSQL_TIMEOUT")
     for p in "${positive_values[@]}"; do
         if [[ -n "${!p:-}" ]]; then
             if ! is_positive_int "${!p}"; then
@@ -292,11 +294,18 @@ pgpool_attach_node() {
 pgpool_healthcheck() {
     info "Checking pgpool health..."
     local backends
-    backends="$(PGCONNECT_TIMEOUT=15 PGPASSWORD="$PGPOOL_POSTGRES_PASSWORD" psql -U "$PGPOOL_POSTGRES_USERNAME" \
-        -d postgres -h "$PGPOOL_TMP_DIR" -p "$PGPOOL_PORT_NUMBER" -tA -c "SHOW pool_nodes;")"
-    if [[ "$backends" ]]; then
-        # look up backends that are marked offline
-        for node in $(echo "${backends}" | grep "down" | tr -d ' '); do
+    # Timeout should be in sync with liveness probe timeout and number of nodes which could be down together
+    # Only nodes marked UP in pgpool are tested. Each failed standby backend consumes up to PGPOOL_CONNECT_TIMEOUT
+    # to test. Network split is worst-case scenario.
+    # NOTE: command blocks indefinitely if primary node is marked UP in pgpool but is DOWN in reality and connection
+    # times-out. Example is again network-split.
+    backends="$(PGCONNECT_TIMEOUT=$PGPOOL_HEALTH_CHECK_PSQL_TIMEOUT PGPASSWORD="$PGPOOL_POSTGRES_PASSWORD" \
+        psql -U "$PGPOOL_POSTGRES_USERNAME" -d postgres -h "$PGPOOL_TMP_DIR" -p "$PGPOOL_PORT_NUMBER" \
+        -tA -c "SHOW pool_nodes;")" || backends="command failed"
+    if [[ "$backends" != "command failed" ]]; then
+        # look up backends that are marked offline and being up - attach only status=down and pg_status=up
+        # situation down|down means PG is not yet ready to be attached
+        for node in $(echo "${backends}" | grep "down|up" | tr -d ' '); do
             IFS="|" read -ra node_info <<< "$node"
             local node_id="${node_info[0]}"
             local node_host="${node_info[1]}"
@@ -307,6 +316,7 @@ pgpool_healthcheck() {
             fi
         done
     else
+        # backends command failed
         return 1
     fi
 }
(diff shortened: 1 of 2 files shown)


bitnami/pgpool/4/debian-11/rootfs/opt/bitnami/scripts/libpgpool.sh
```diff
@@ -71,10 +71,11 @@ export PGPOOL_ENABLE_STATEMENT_LOAD_BALANCING="${PGPOOL_ENABLE_STATEMENT_LOAD_BA
...
 export PGPOOL_HEALTH_CHECK_USER="${PGPOOL_HEALTH_CHECK_USER:-$PGPOOL_SR_CHECK_USER}"
-export PGPOOL_HEALTH_CHECK_PERIOD="${PGPOOL_HEALTH_CHECK_PERIOD:-30}"
-export PGPOOL_HEALTH_CHECK_TIMEOUT="${PGPOOL_HEALTH_CHECK_TIMEOUT:-10}"
-export PGPOOL_HEALTH_CHECK_MAX_RETRIES="${PGPOOL_HEALTH_CHECK_MAX_RETRIES:-5}"
-export PGPOOL_HEALTH_CHECK_RETRY_DELAY="${PGPOOL_HEALTH_CHECK_RETRY_DELAY:-5}"
+export PGPOOL_HEALTH_CHECK_PERIOD="${PGPOOL_HEALTH_CHECK_PERIOD:-13}"
+export PGPOOL_HEALTH_CHECK_TIMEOUT="${PGPOOL_HEALTH_CHECK_TIMEOUT:-2}"
+export PGPOOL_HEALTH_CHECK_MAX_RETRIES="${PGPOOL_HEALTH_CHECK_MAX_RETRIES:-3}"
+export PGPOOL_HEALTH_CHECK_RETRY_DELAY="${PGPOOL_HEALTH_CHECK_RETRY_DELAY:-2}"
+export PGPOOL_CONNECT_TIMEOUT="${PGPOOL_CONNECT_TIMEOUT:-2000}"
```
fmulero, line 78: I've been taking a look to the configuration and I think 2s is too agressive for the `PGPOOL_CONNECT_TIMEOUT` and the `PGPOOL_HEALTH_CHECK_RETRY_DELAY` should be greater than `PGPOOL_CONNECT_TIMEOUT` to handle properly errors triggered by that threshold.

What do you think about this possible default config?:

```suggestion
export PGPOOL_HEALTH_CHECK_PERIOD="${PGPOOL_HEALTH_CHECK_PERIOD:-13}"
export PGPOOL_HEALTH_CHECK_TIMEOUT="${PGPOOL_HEALTH_CHECK_TIMEOUT:-5}"
export PGPOOL_HEALTH_CHECK_MAX_RETRIES="${PGPOOL_HEALTH_CHECK_MAX_RETRIES:-2}"
export PGPOOL_HEALTH_CHECK_RETRY_DELAY="${PGPOOL_HEALTH_CHECK_RETRY_DELAY:-2}"
export PGPOOL_CONNECT_TIMEOUT="${PGPOOL_CONNECT_TIMEOUT:-3000}"
```

fmulero (changes requested):
Hi @pcerny Thanks a lot for the contribution. It looks a great improvement.

It LGTM but I've left a comment about the default config. Also could you update your branch with the latest changes from the `main` branch? The container is not spinning up correctly from your branch and I think that problem  is already fixed in the main branch.

bitnami/pgpool/4/debian-11/rootfs/opt/bitnami/scripts/libpgpool.sh
rafariossaa: Hi,
  I agree with @fmulero, I wouldn't go so aggressive by default and leave them as they are. As they are configurable by the user, they could be set easily to other values by the user if they need it.

pcerny:
TL;DR I would like to rework this PR because I have uncovered race condition in current solution. New solution will be based on marking PG headless service with flag `publishNotReadyAddresses: true`. If you have any thought or experience with that, please share them.

By yet another testing of this PR, I have uncovered that liveness check update contains race condition. It is working only under specific conditions. Most of my tests satisfied those condition so I did not uncover it earlier.
Below is simple rolling upgrade scenario for 3 PG backends which results in liveness check failure and by that to restart of pgpool:
- Starting rolling upgrade in state PG-0 UP (primary), PG-1 UP (standby), PG-2 UP (standby)
- Update PG-2, pgpoolu marks PG-2 DOWN
- As soon as new instance of PG-2 is ready, PG-1 is being terminated and pgpool can connect to PG-2. After some time, pgpool 's liveness probe attaches PG-2 and PG-2 is marked UP (standby)
- During PG-1 termination pgpool recognizes it as DOWN
- As soon as new instance of PG-1 is ready, PG-0 is terminated. PG-0 is primary. pgpool is in state PG-2 UP (standby), PG-1 DOWN (standby), PG-0 UP (primary)
- Repmgr detects PG-0 is unreachable and switches primary to PG-1. In-parallel, pgpool marks PG-0 as DOWN. pgpool is in state PG-2 UP (standby), PG-1 DOWN (standby), PG-0 DOWN (primary)
- pgpool refuses pcp interface actions attach_node or promote_node generally because it is in search of primary in UP state (DOWN state nodes are not included into search)
- pgpool's liveness probe is failing, pgpool is eventually restarted

Problem is that pgpool does not mark PG-1 as UP before PG-0 is marked as DOWN. This two operations are executed in parallel. In order to serialize them, one needs to delay termination of PG-0 which is always triggered by PG-1 achieving ready state. Similarly, pgpool can reach PG-1 once it gets ready because services showing only ready PODs (by default).

Luckily for me, our k8s guru advised me, there exists k8s service option `publishNotReadyAddresses: true`. I did various tests with this flag enabled. All of them successful. I can share more details from tests to answer questions like how repmgr works when PG headless service has this flag set, ideally in form of article. But I am fighting to find a free time slot for that.

 I will try to update both PRs in 2 days. If you have any comments sooner, please let me know.

fmulero:
Thanks a lot @pcerny! I am eager to see your next steps

`publishNotReadyAddresses: true`is very useful to discover new members in cluster like this case. You have to be careful if you are exposing that service to your clients because you could reach a non ready pod.

pcerny:
I decided to keep defaults to minimize impact on current PG-HA setups. So this PR adds just two new env variables as described in the first PR comment.

Related PR https://github.com/bitnami/charts/pull/14175 builds on it and introduce ability to setup PG headless service with `publishNotReadyAddresses: true`. I leave comment about how I test it in the related PR.

bitnami/pgpool/4/debian-11/rootfs/opt/bitnami/scripts/libpgpool.sh
```diff
@@ -292,21 +294,29 @@ pgpool_attach_node() {
...
+    if [[ "$backends" != "command failed" ]]; then
+        # look up backends that are marked offline and being up - attach only status=down and pg_status=up
+        # situation down|down means PG is not yet ready to be attached
+        for node in $(echo "${backends}" | grep "down|up" | tr -d ' '); do
             IFS="|" read -ra node_info <<< "$node"
             local node_id="${node_info[0]}"
             local node_host="${node_info[1]}"
             if [[ $(PGCONNECT_TIMEOUT=3 PGPASSWORD="${PGPOOL_POSTGRES_PASSWORD}" psql -U "${PGPOOL_POSTGRES_USERNAME}" \
-                -d postgres -h "${node_host}" -p "${PGPOOL_PORT_NUMBER}" -tA -c "SELECT 1" || true) == 1 ]]; then
+                 -d postgres -h "${node_host}" -p "${PGPOOL_PORT_NUMBER}" -tA -c "SELECT 1" || true) == 1 ]]; then
```
rafariossaa, line 313: Remove this unneeded white space.

bitnami/pgpool/README.md
```diff
@@ -267,10 +267,12 @@ Pgpool configuration:
...
-- `PGPOOL_HEALTH_CHECK_TIMEOUT`: Specifies the timeout in seconds to give up connecting to the backend PostgreSQL if the TCP connect does not succeed within this time. Defaults to `10`.
-- `PGPOOL_HEALTH_CHECK_MAX_RETRIES`: Specifies the maximum number of retries to do before giving up and initiating failover when health check fails. Defaults to `5`.
-- `PGPOOL_HEALTH_CHECK_RETRY_DELAY`: Specifies the amount of time in seconds to sleep between failed health check retries. Defaults to `5`.
+- `PGPOOL_HEALTH_CHECK_PERIOD`: Specifies the interval between the health checks in seconds. Defaults to `13`.
+- `PGPOOL_HEALTH_CHECK_TIMEOUT`: Specifies the timeout in seconds to give up connecting to the backend PostgreSQL if the TCP connect does not succeed within this time. Defaults to `2`.
+- `PGPOOL_HEALTH_CHECK_MAX_RETRIES`: Specifies the maximum number of retries to do before giving up and initiating failover when health check fails. Defaults to `3`.
+- `PGPOOL_HEALTH_CHECK_RETRY_DELAY`: Specifies the amount of time in seconds to sleep between failed health check retries. Defaults to `2`.
```
rafariossaa, line 273: I think for this we should keep the previous default values, and set then to other values, in the chart file.
This could break the current behavior for the current users of the image.

bitnami/pgpool/README.md
```diff
@@ -267,10 +267,12 @@ Pgpool configuration:
...
+- `PGPOOL_CONNECT_TIMEOUT`: Specifies the amount of time in milliseconds before giving up connecting to backend using `connect()` system call. Default is `10000`.
+- `PGPOOL_HEALTH_CHECK_PSQL_TIMEOUT`: Defines PGCONNECT_TIMEOUT value for "show pool_nodes" command in "pgpool_healthcheck" function. Default is `15`.
```
rafariossaa, line 275: Indicate the time unit (seconds).

rafariossaa (dismissed):

bitnami/pgpool/4/debian-11/rootfs/opt/bitnami/scripts/libpgpool.sh
pcerny: Fixed

bitnami/pgpool/README.md
pcerny: Fixed

bitnami/pgpool/README.md
pcerny: Fixed

fmulero (approved):
LGTM

fmulero (approved):

Issue #17802, opened 2022-12-22
MySQL check ignores SSL settings


duk3luk3:
### Name and Version

bitnami/wordpress:6.1.1

### What steps will reproduce the bug?

1. Set up a MySQL server with `require_secure_transport` configuration
2. Configure these wordpress options:
  * DB host etc.
  * `WORDPRESS_ENABLE_DATABASE_SSL=yes`
  * `WORDPRESS_VERIFY_DATABASE_SSL=no`
  * `BITNAMI_DEBUG=yes`
4. Start the container

### What is the expected behavior?

Wordpress is initialized and started, and all DB connections are secured

### What do you see instead?

Container fails with "Could not connect to the database" error

### Additional information

The root cause for this issue is that the `wordpress_wait_for_mysql_connection` never attempts a secure connection:

https://github.com/bitnami/containers/blob/208c7fdbc5330ae7122edb041fa693ccc0eee3a0/bitnami/wordpress/6/debian-11/rootfs/opt/bitnami/scripts/libwordpress.sh#L562-L575

This needs to be fixed.

carrodher:
Thanks for reporting this issue. Would you like to contribute by creating a PR to solve the issue? The Bitnami team will be happy to review it and provide feedback. [Here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md) you can find the contributing guidelines.

ferronsw:
Having the same issue!

PR #19204, opened 2023-01-05
[bitnami/wordpress] Added SSL support

FerronN:
Signed-off-by: Ferron Nijland <fernijland@hotmail.com>

Solves #17802


FerronN:
It looks like this works when not persisting the `wp-config.php`. 
I can't find why (at this time), but `db_flags="$(wordpress_conf_get "MYSQL_CLIENT_FLAGS")"` doesn't return the defined value in `wp-config.php` which is correct. It returns 2048...

bitnami/wordpress/6/debian-11/rootfs/opt/bitnami/scripts/libwordpress.sh
```diff
@@ -565,8 +587,11 @@ wordpress_wait_for_mysql_connection() {
     local -r db_name="${3:?missing database name}"
     local -r db_user="${4:?missing database user}"
     local -r db_pass="${5:-}"
+    local -r db_options="${6:-}"
     check_mysql_connection() {
-        echo "SELECT 1" | mysql_remote_execute "$db_host" "$db_port" "$db_name" "$db_user" "$db_pass"
+        #Remove test first line
+        echo "SELECT 1" "$db_host" "$db_port" "$db_name" "$db_user" "$db_pass" "$db_options"
```
Mauraza, line 593: I think this line isn't necessary, could you check it?

Mauraza:
Hi @FerronN,

I see that the PR is un Draft, could you ping me when it's ready?

FerronN:
@Mauraza I will. I'm currently in the process of solving the bug with the wp-cli team.

FerronN:
@Mauraza It seem like the WP-CLI has no clue for now regarding the issue mentioned earlier. 
So I've made a work around. For this to work the: `WORDPRESS_ENABLE_DATABASE_SSL` and `WORDPRESS_VERIFY_DATABASE_SSL` always need to be present. It will be saved into wp-config (for when the workaround can be fixed), but it won't get the value from the wp-config.

Mauraza:
Hi @FerronN,

I going to check all. Could you sing the commits?
You can follow this instruction https://github.com/bitnami/containers/pull/19204/checks?check_run_id=11219797719

PR #20537, opened 2023-01-17
[bitnami/discourse] Allow to parametrize Target App Version

### LINKED ISSUE #20536: [bitnami/discourse] Releases not updated
### Name and Version

bitnami/discourse

I would like to use recent versions of discourse but the bitnami-bot is not running so frequently as I would expect.
The last version released as a bitnami container was launched on November, and since then discourse has released at least 5 more [versions](https://github.com/discourse/discourse/tags).
I would like to used minimum version v3.0.0.
I can provide the PR, but I think that the github action should do it.

To keep versions of discourse updated from time to time

### What do you see instead?

Hardcoded version on Dockerfile, making imposible to use a parametrized Dockerfile ARG to build your own images.

### Additional information

_No response_

ausias-armesto:
### Benefits

Allows easily to change the APP_VERSION of the helm chart to be overriden by users.

### Possible drawbacks

It may have impact in any bot that performs this task or similar

### Applicable issues

  - fixes #20536

### DIFF
--- a/bitnami/discourse/2/debian-11/Dockerfile
+++ b/bitnami/discourse/2/debian-11/Dockerfile
@@ -40,10 +40,8 @@ RUN mkdir -p /tmp/bitnami/pkg/cache/ && cd /tmp/bitnami/pkg/cache/ && \
         curl -SsLf "https://downloads.bitnami.com/files/stacksmith/${COMPONENT}.tar.gz.sha256" -O ; \
       fi && \
       sha256sum -c "${COMPONENT}.tar.gz.sha256" && \
-      tar -zxf "${COMPONENT}.tar.gz" -C /opt/bitnami --strip-components=2 --no-same-owner --wildcards '*/files' && \
-      # rm -rf "${COMPONENT}".tar.gz{,.sha256} ; \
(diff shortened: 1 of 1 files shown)


javsalgar:
Hi,

Thank you so much for the PR! However, please note that these repos, and the Dockerfiles, are auto-generated by our internal pipeline, so the changes you are proposing in the Dockerfile will be automatically overidden in the next release. In order to allow the configuration you mention in the Dockerfile, we would need to perform non-trivial changes in our internal pipelines and generation tools that would require a substantial amount of work from the team. 

I will forward this to the team for getting feedback, but unfortunately we cannot accept this change at the moment.

ausias-armesto:
Hi @javsalgar ,

It's a pity, because this block users to be able to use only the specific versions that bitnami release, which derivatives from the ones that discourse releases in quite a lot of time.

I have no other option but to fork until you solve this issue.

ausias-armesto:
I've checked the PR, and is not going to work either as you are not publishing the new zip files to this url `https://downloads.bitnami.com/files/stacksmith/${COMPONENT}.tar.gz`

PR #21215, opened 2023-01-20
[bitnami/cassandra] cassandra.yaml changed from 4 Spaces to 2 Spaces in V4.1

jmederer:
Signed-off-by: Jan Mederer <jan@mederer.it>

### Description of the change
Client_Encryption_Options is disabled even TLS Option is set to true.

up to Version 4.0.7
```console
$ docker run --rm -ti --entrypoint /bin/bash docker.io/bitnami/cassandra:4.0.7-debian-11-r9  -c "cat /opt/bitnami/cassandra/conf.default/cassandra.yaml" | grep -i client_en -A5 | whitespace
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
#·Enabling·native·transport·encryption·in·client_encryption_options·allows·you·to·either·use§¶
#·encryption·for·the·standard·port·or·to·use·a·dedicated,·additional·port·along·with·the·unencrypted§¶
#·standard·native_transport_port.§¶
#·Enabling·client·encryption·and·keeping·native_transport_port_ssl·disabled·will·use·encryption§¶
#·for·native_transport_port.·Setting·native_transport_port_ssl·to·a·different·value§¶
#·from·native_transport_port·will·use·encryption·for·native_transport_port_ssl·while§¶
--¶
client_encryption_options:§¶
····#·Enable·client-to-server·encryption§¶
····enabled:·false§¶
····#·When·set·to·true,·encrypted·and·unencrypted·connections·are·allowed·on·the·native_transport_port§¶
····#·This·should·_only·be·true_·while·in·unencrypted·or·transitional·operation§¶
····#·optional·defaults·to·true·when·enabled·is·false,·and·false·when·enabled·is·true.§¶
```

since Version 4.1
```console
$ docker run --rm -ti --entrypoint /bin/bash docker.io/bitnami/cassandra:4.1.0-debian-11-r0  -c "cat /opt/bitnami/cassandra/conf.default/cassandra.yaml" | grep -i client_en -A5 | whitespace
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
#·Enabling·native·transport·encryption·in·client_encryption_options·allows·you·to·either·use§¶
#·encryption·for·the·standard·port·or·to·use·a·dedicated,·additional·port·along·with·the·unencrypted§¶
#·standard·native_transport_port.§¶
#·Enabling·client·encryption·and·keeping·native_transport_port_ssl·disabled·will·use·encryption§¶
#·for·native_transport_port.·Setting·native_transport_port_ssl·to·a·different·value§¶
#·from·native_transport_port·will·use·encryption·for·native_transport_port_ssl·while§¶
--¶
client_encryption_options:§¶
··#·Enable·client-to-server·encryption§¶
··enabled:·false§¶
··#·When·set·to·true,·encrypted·and·unencrypted·connections·are·allowed·on·the·native_transport_port§¶
··#·This·should·_only·be·true_·while·in·unencrypted·or·transitional·operation§¶
··#·optional·defaults·to·true·when·enabled·is·false,·and·false·when·enabled·is·true.§¶
```

### Benefits

TLS works again :-)
### Applicable issues

  - fixes #


marcosbc (changes requested):
Hi @jmederer, thanks for the contribution!

Unfortunately the `libcassandra.sh` file is shared across all branches (4.0 and 4.1) in our internal codebase, even though they appear in different paths in this repository.

Due to this, we will need to implement a workaround that works for both branches.

I propose changing `\s{4}` by `(\s+)` and then conserving the spaces by `\2`. That way, even if there are 4 spaces or 2, it should continue working fine, as long as there is at least 1 space.

jmederer:
@marcosbc thank you for the review.
I've updated the PR as you suggest.

jmederer:
@marcosbc sorry for the "wrong" commit 
I hope that its correct now

marcosbc (approved):
LGTM, thanks for contributing the fix!

PR #21550, opened 2023-01-24
Add $MATOMO_DATABASE_TABLE_PREFIX [default: "matomo_"] as wizard option

bebosudo:
### Description of the change

Add an option to set the prefix for database tables.

### Benefits

More options available to adapt this image to existing setups.

### Possible drawbacks

None

### Applicable issues

None

### Additional information

None

### DIFF
--- a/bitnami/matomo/4/debian-11/rootfs/opt/bitnami/scripts/matomo-env.sh
+++ b/bitnami/matomo/4/debian-11/rootfs/opt/bitnami/scripts/matomo-env.sh
@@ -41,6 +41,7 @@ matomo_env_vars=(
     MATOMO_DATABASE_SSL_CERT_FILE
     MATOMO_DATABASE_SSL_KEY_FILE
     MATOMO_VERIFY_DATABASE_SSL
+    MATOMO_DATABASE_TABLE_PREFIX
     MATOMO_SMTP_HOST
     MATOMO_SMTP_PORT_NUMBER
     MATOMO_SMTP_USER
@@ -128,6 +129,7 @@ export MATOMO_DATABASE_PORT_NUMBER="${MATOMO_DATABASE_PORT_NUMBER:-3306}" # only
 export MATOMO_DATABASE_NAME="${MATOMO_DATABASE_NAME:-bitnami_matomo}" # only used during the first initialization
 export MATOMO_DATABASE_USER="${MATOMO_DATABASE_USER:-bn_matomo}" # only used during the first initialization
 export MATOMO_DATABASE_PASSWORD="${MATOMO_DATABASE_PASSWORD:-}" # only used during the first initialization
+export MATOMO_DATABASE_TABLE_PREFIX="${MATOMO_DATABASE_TABLE_PREFIX:-matomo_}" # only used during the first initialization
 
 # PHP configuration
 export PHP_DEFAULT_MEMORY_LIMIT="256M" # only used at build time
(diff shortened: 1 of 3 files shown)


carrodher:
Thanks for creating this PR. Please, note it is needed to sign the commits. Your signature certifies that you wrote the patch or otherwise have the right to contribute the material. You can find more info about this topic, in [the “Sign your work” section from the contributing guidelines](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#sign-your-work).

bebosudo:
Done!
I didn't see tests in the repo, please let me know if I'm missing something.

bitnami/matomo/4/debian-11/rootfs/opt/bitnami/scripts/matomo-env.sh
```diff
@@ -128,6 +129,7 @@ export MATOMO_DATABASE_PORT_NUMBER="${MATOMO_DATABASE_PORT_NUMBER:-3306}" # only
 export MATOMO_DATABASE_NAME="${MATOMO_DATABASE_NAME:-bitnami_matomo}" # only used during the first initialization
 export MATOMO_DATABASE_USER="${MATOMO_DATABASE_USER:-bn_matomo}" # only used during the first initialization
 export MATOMO_DATABASE_PASSWORD="${MATOMO_DATABASE_PASSWORD:-}" # only used during the first initialization
+export MATOMO_DATABASE_TABLE_PREFIX="${MATOMO_DATABASE_TABLE_PREFIX:-}" # only used during the first initialization
```
marcosbc, line 132: For compatibility reasons, we should keep the default `matomo_` value. Could you change it?

```suggestion
export MATOMO_DATABASE_TABLE_PREFIX="${MATOMO_DATABASE_TABLE_PREFIX:-matomo_}" # only used during the first initialization
```

marcosbc (approved):
Thanks for the contribution! Please have a look at my comment whenever you have a chance.

marcosbc (changes requested):
Sorry, I chose the wrong option. Once it is fixed I will approve it.

bitnami/matomo/4/debian-11/rootfs/opt/bitnami/scripts/matomo-env.sh
bebosudo: Just pushed a fix with fde647c, apologies for the oversight.

marcosbc: It seems to be the same, could you re-check it?

marcosbc:
The default value for `MATOMO_DATABASE_TABLE_PREFIX` seems to be still empty in `matomo-env.sh`, could you re-check it?

bitnami/matomo/4/debian-11/rootfs/opt/bitnami/scripts/matomo-env.sh
bebosudo: Sorry about that, forgot to git add the edited file. Should be there now.

marcosbc (approved):
LGTM, thanks!

bebosudo:
Hi, any update on when would this be available in the [hub](https://hub.docker.com/r/bitnami/matomo/tags)? :-)

JoshuaSmeda:
There's a flaw with this because it doesn't allow for empty ("") values. So if a user doesn't have prefixed tables, they're out of luck because they can't override the behaviour - the ini-file will always default to `matomo_` in this case.

PR #22437, opened 2023-01-31
[bitnami/wordpress-nginx] Reverting change that broke behavior

bengaywins:
### Description of the change

This PR reverts behavior that broke the image for us and likely others.

### Benefits

The behavior is no longer broken.

### Possible drawbacks

No idea, but I would argue behavior changes should not be in minor versions.

### Applicable issues

  - fixes #12995

PR #22519, opened 2023-02-01
[bitnami/wordpress-nginx] Reverting change that broke behavior

### LINKED ISSUE #12995: [bitnami/wordpress-nginx] Wordpress startup is broken in latest wordpress-nginx
### Name and Version

bitnami/wordpress-nginx v6.0.3

The startup of 6.0.3 gives us an error with sed that does _not_ present with 6.0.2.

```
wordpress 18:51:29.39 Welcome to the Bitnami wordpress-nginx container
wordpress 18:51:29.39 Subscribe to project updates by watching https://github.com/bitnami/containers
wordpress 18:51:29.39 Submit issues and feature requests at https://github.com/bitnami/containers/issues
wordpress 18:51:29.40
wordpress 18:51:29.40 INFO  ==> ** Starting WordPress setup **
wordpress 18:51:29.42 INFO  ==> Validating settings in NGINX_* env vars
wordpress 18:51:29.45 INFO  ==> No custom scripts in /docker-entrypoint-initdb.d
wordpress 18:51:29.47 INFO  ==> Initializing NGINX
realpath: /bitnami/nginx/conf/vhosts: No such file or directory
sed: can't read /opt/bitnami/nginx/conf/server_blocks/default-https-server-block.conf: No such file or directory
```

I don't see exactly where this is happening, but it's surely not happening with our build process when we source your image. I can post a sanitized version of our `Dockerfile` if necessary.

The image has no failures on startup

### What do you see instead?

We get an error and it bails early

### Additional information

_No response_

bengaywins:
Signed-off-by: Ben Gray <gehzumteufel@gmail.com>

### Description of the change

This PR recerts bahvior that broke the image for us and likely others.

### Benefits

The behavior is no longer broken

### Possible drawbacks

No idea, but I would argue behavior that changes should not be in minor versions.

### Applicable issues

  - fixes #12995

### DIFF
--- a/bitnami/wordpress-nginx/6/debian-11/rootfs/opt/bitnami/scripts/nginx/setup.sh
+++ b/bitnami/wordpress-nginx/6/debian-11/rootfs/opt/bitnami/scripts/nginx/setup.sh
@@ -31,7 +31,7 @@ nginx_custom_init_scripts
 ! am_i_root || chmod o+w "$(readlink /dev/stdout)" "$(readlink /dev/stderr)"
 
 # Configure HTTPS port number
-if [[ -n "${NGINX_HTTPS_PORT_NUMBER:-}" ]] && [[ ! -f "${NGINX_SERVER_BLOCKS_DIR}/default-https-server-block.conf" ]] && is_dir_empty "${NGINX_SERVER_BLOCKS_DIR}"; then
+if [[ -n "${NGINX_HTTPS_PORT_NUMBER:-}" ]] && [[ ! -f "${NGINX_SERVER_BLOCKS_DIR}/default-https-server-block.conf" ]] || is_dir_empty "${NGINX_SERVER_BLOCKS_DIR}"; then
     cp "${BITNAMI_ROOT_DIR}/scripts/nginx/server_blocks/default-https-server-block.conf" "${NGINX_SERVER_BLOCKS_DIR}/default-https-server-block.conf"
 fi
 


bengaywins:
@carrodher Here's the new PR for the same change in #22437.

corico44:
Hello @bengaywins,

I don't quite understand the reason for the change. Some time ago the part of code that we want to remove was added in this PR to fix a problem: https://github.com/bitnami/charts/issues/13041. Why do you want to delete it again?

bengaywins:
Because nobody had any feedback on it breaking our usage. The link you mention I think gives me the help though to possibly ferry my needed changes. So I am going to try that.

corico44:
Should we close this PR then @bengaywins?

bengaywins:
I haven't gotten to test yet, but working on it.

bengaywins:
@corico44 I think this is the correct solution and will amend my PR. The reason I believe this, is that the logic seems to be intended to be:

If first two conditions aren't both true OR the dir is empty, then copy this file. Otherwise do nothing. 

There's no other real explanation I can gain from this.

`if [[ -n "${NGINX_HTTPS_PORT_NUMBER:-}" ]] && [[ ! -f "${NGINX_SERVER_BLOCKS_DIR}/default-https-server-block.conf" ]] || is_dir_empty "${NGINX_SERVER_BLOCKS_DIR}"; then`

If you believe otherwise, even in the context of the linked PR you gave, I'm all ears. Unfortunately, my original PR breaks the behavior that has been expected for a while (we've been using this image since 5.5).

corico44:
@bengaywins, thank you so much for the explanation. We are discussing it with the team. We will inform you of the decision made.

corico44 (approved):
LGTM

PR #22777, opened 2023-02-03
[bitnami/keycloak] Support for h2 database (#22776)

### LINKED ISSUE #22776: Keycloak: Hard coded postgresql database setup
### Name and Version

bitnami/keycloak:20.0.3

1. Start container without postgresql database available.
2. Watch container try to connect and fail during setup.

I would like to use the container with an embedded h2 database.

### What do you see instead?

Only postgresql supported.

### Additional information

_No response_

boehlke:
### Description of the change

The commit adds support for using the image with an embedded h2 database.

### Benefits

This comes in handy when using the image with a test setup.

### Possible drawbacks

The default behavior still is using a postgresql database. I don't expect, Jboss will drop support for h2, so the h2 setup should be stable. So I see no drawback.

### Applicable issues

  - fixes #22776


aoterolorenzo:
Hi @boehlke,

May you provide the steps to test the feature (maybe a docker-compose.yaml, or commands and config to try out from two containers)?

boehlke:
@aoterolorenzo Sorry for the late response. You can test the setup by setting the environment variable KEYCLOAK_DATABASE_VENDOR to dev-mem or dev-file. The result is that no postgresql database is required.

aoterolorenzo (approved):
LGTM

PR #23527, opened 2023-02-10
run dokuwiki_configure_DOKU_INC on every startup

### LINKED ISSUE #22466: DokuWiki: DOKU_INC not initialized on container upgrade
### Name and Version

bitnami/dokuwiki:20220731

When you upgrade to the newest image version, it includes the mechanism to autoprepend the setting of the DOKU_INC define as discussed in #12535, however since the initialization has already been done with a previous image version it will be skipped completely, making the fix unavailable to users of previous versions.

The existence of the auto prepend file should be checked on every container start and be initialized if necessary.

### What do you see instead?

No auto prepend is configured and DOKU_INC is pointing to the wrong location for non-main scripts.

### Additional information

_No response_

kossmac:
Signed-off-by: Karsten Kosmala <kosmala@cosmocode.de>

### Description of the change

We need to run dokuwiki_configure_DOKU_INC on **every** startup, not only when DW gets persisted the first time. So already persisted and new installations will benefit from this.

### Applicable issues

Fixes https://github.com/bitnami/containers/issues/22466

Related to https://github.com/bitnami/containers/pull/12535

### DIFF
--- a/bitnami/dokuwiki/20220731/debian-11/rootfs/opt/bitnami/scripts/libdokuwiki.sh
+++ b/bitnami/dokuwiki/20220731/debian-11/rootfs/opt/bitnami/scripts/libdokuwiki.sh
@@ -63,7 +63,6 @@ dokuwiki_initialize() {
             web_server_stop
             dokuwiki_enable_friendly_urls
         fi
-        dokuwiki_configure_DOKU_INC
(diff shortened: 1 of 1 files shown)


dgomezleon (approved):

PR #24120, opened 2023-02-15
[bitnami/moodle] Deprecate Moodle 3

corico44:
### Description of the change

Deprecate Moodle 3.x.x branch.


dgomezleon (approved):

carrodher (approved):

strickes:
Moodle 3.x.x is supported till end of the year, can you please continue building it until then?

carrodher:
We focus our support on those versions marked as Stable versions, in this case, 4.0 and 4.1. Because of that, there are no plans to add versions in a "security support" state

PR #24225, opened 2023-02-17
[bitnami/codeigniter] Add codeigniter tests

mdhont:
Including Goss VIB tests for the CodeIgniter container. The container build process does not involve any compilation or modification in the post-unpack phase. As a result, the scope for application-specific tests is limited.

Successful test: https://github.com/mdhont/containers/actions/runs/4203692633/jobs/7293411531

### DIFF
--- a/.vib/codeigniter/goss/goss.yaml
+++ b/.vib/codeigniter/goss/goss.yaml
@@ -1,5 +1,7 @@
 gossfile:
   # Goss tests exclusive to the current container
+  ../../codeigniter/goss/codeigniter.yaml: {}
+  # Goss tests exclusive to the current container
   # Load scripts from .vib/common/goss/templates
   ../../common/goss/templates/check-app-version.yaml: {}
   ../../common/goss/templates/check-broken-symlinks.yaml: {}
--- a/.vib/codeigniter/goss/goss.yaml
+++ b/.vib/codeigniter/goss/goss.yaml
@@ -1,7 +1,6 @@
 gossfile:
   # Goss tests exclusive to the current container
   ../../codeigniter/goss/codeigniter.yaml: {}
-  # Goss tests exclusive to the current container
   # Load scripts from .vib/common/goss/templates
   ../../common/goss/templates/check-app-version.yaml: {}
   ../../common/goss/templates/check-broken-symlinks.yaml: {}
--- a/.vib/codeigniter/goss/goss.yaml
+++ b/.vib/codeigniter/goss/goss.yaml
@@ -7,7 +7,6 @@ gossfile:
   ../../common/goss/templates/check-broken-symlinks.yaml: {}
   ../../common/goss/templates/check-ca-certs.yaml: {}
   ../../common/goss/templates/check-binaries.yaml: {}
-  ../../common/goss/templates/check-files.yaml: {}
   ../../common/goss/templates/check-directories.yaml: {}
   ../../common/goss/templates/check-linked-libraries.yaml: {}
   ../../common/goss/templates/check-sed-in-place.yaml: {}
--- a/.vib/codeigniter/goss/goss.yaml
+++ b/.vib/codeigniter/goss/goss.yaml
@@ -1,0 +1,11 @@
+gossfile:
+  # Goss tests exclusive to the current container
+  # Load scripts from .vib/common/goss/templates
+  ../../common/goss/templates/check-app-version.yaml: {}
+  ../../common/goss/templates/check-broken-symlinks.yaml: {}
+  ../../common/goss/templates/check-ca-certs.yaml: {}
+  ../../common/goss/templates/check-binaries.yaml: {}
+  ../../common/goss/templates/check-files.yaml: {}
+  ../../common/goss/templates/check-directories.yaml: {}
+  ../../common/goss/templates/check-linked-libraries.yaml: {}
+  ../../common/goss/templates/check-sed-in-place.yaml: {}
--- a/.vib/codeigniter/goss/vars.yaml
+++ b/.vib/codeigniter/goss/vars.yaml
@@ -7,11 +7,8 @@ binaries:
 files:
   - paths:
       - /opt/bitnami/codeigniter/app/Common.php
-      - /opt/bitnami/codeigniter/app/index.html
-      - /opt/bitnami/codeigniter/psalm-baseline.xml
       - /opt/bitnami/codeigniter/psalm.xml
       - /opt/bitnami/codeigniter/composer.json
-      - /opt/bitnami/codeigniter/spark
 directories:
   - mode: "0775"
     paths:
--- a/.vib/codeigniter/goss/vars.yaml
+++ b/.vib/codeigniter/goss/vars.yaml
@@ -2,7 +2,6 @@ binaries:
   - composer
   - php-fpm
   - gosu
-  - mmdblookup
   - mysql
 files:
   - paths:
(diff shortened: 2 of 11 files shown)


.vib/codeigniter/goss/vars.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - mmdblookup
+  - mysql
+files:
+  - paths:
+      - /opt/bitnami/codeigniter/app/Common.php
+      - /opt/bitnami/codeigniter/app/index.html
+      - /opt/bitnami/codeigniter/psalm-baseline.xml
+      - /opt/bitnami/codeigniter/psalm.xml
+      - /opt/bitnami/codeigniter/composer.json
+      - /opt/bitnami/codeigniter/spark
```
marcosbc, line 14: I think we are checking for too many paths here. A simple minor version could rename or move some of these files to different locations, causing it to fail. Maybe it's worth to be a bit more granular?

.vib/codeigniter/vib-verify.json
```diff
@@ -3,7 +3,8 @@
     "resources": {
       "url": "{SHA_ARCHIVE}",
       "path": "{VIB_ENV_PATH}"
-    }
+    },
+    "runtime_parameters": "Y29tbWFuZDogWyJ0YWlsIiwgIi1mIiwgIi9kZXYvbnVsbCJd"
```
marcosbc, line 7: We could take the opportunity to enable `runtime_parameters_file` here.

.vib/codeigniter/vib-publish.json
```diff
@@ -3,7 +3,8 @@
     "resources": {
       "url": "{VIB_ENV_CONTAINER_URL}",
       "path": "{VIB_ENV_PATH}"
-    }
+    },
+    "runtime_parameters": "Y29tbWFuZDogWyJ0YWlsIiwgIi1mIiwgIi9kZXYvbnVsbCJd"
```
marcosbc, line 7: We could take the opportunity to enable `runtime_parameters_file` here.

.vib/codeigniter/goss/vars.yaml
```diff
@@ -0,0 +1,22 @@
+binaries:
+  - composer
+  - php-fpm
+  - gosu
+  - mmdblookup
```
marcosbc, line 5: Why adding mmdblookup here? Is it required for CodeIgniter somehow?

marcosbc (dismissed):

.vib/codeigniter/goss/vars.yaml
mdhont: Ok I removed it

.vib/codeigniter/goss/vars.yaml
mdhont: I removed spark because it's used in another tests anywy, and psalm-baseline.

.vib/codeigniter/vib-publish.json
mdhont: Wdyt @FraPazGal?

FraPazGal: As of now, the runtime params for containers only accept a single parameter, which is the `command` one. This is used to override the container's default one. Unlike `bitnami/charts`, here we do not have custom parameters that actively affect the tests and are prone to changes when the chart is updated.

Maybe in the future, when in-depth test suites are created for some of these containers, we will use more complex runtime parameters, and having them in a `yaml` file adds some values. Until then, I think using the encoded string is enough.

FraPazGal (commented):
Hi @mdhont, thanks for the PR! Could you please check my comments?

.vib/codeigniter/goss/goss.yaml
```diff
@@ -0,0 +1,11 @@
+gossfile:
+  # Goss tests exclusive to the current container
```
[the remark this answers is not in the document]
mdhont: @FraPazGal, I've added a check for "spark serve" using "timeout 2s". I don't think it's possible to check if the development server is running afterwards.

.vib/codeigniter/goss/vars.yaml
```diff
@@ -0,0 +1,18 @@
+binaries:
+  - composer
+  - php-fpm
+  - gosu
+  - mysql
+files:
+  - paths:
+      - /opt/bitnami/codeigniter/app/Common.php
+      - /opt/bitnami/codeigniter/psalm.xml
```
joancafom, line 9: I don't think this path adds any value.

.vib/codeigniter/goss/vars.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - /opt/bitnami/codeigniter/composer.json
+directories:
+  - mode: "0775"
+    paths:
+      - /app
+root_dir: /opt/bitnami
+version:
+  bin_name: /opt/bitnami/codeigniter/spark
+  flag: --version
```
joancafom, line 18: Beware that the output this produces is:

```
...
CodeIgniter v4.3.1 Command Line Tool - Server Time: 2023-03-02 16:43:28 UTC+00:00
...
```

It might not be a problem, but it seems that version is related to the CLI itself, so there might be a mismatch between the product and the CLI.

mdhont: It is the only way to check for a version, I could remove the whole test if it's better.

FraPazGal: I would first investigate where are we getting both the CLI and product from. If they are both taken from the same source code, have different release cycles, etc. Unless absolutely necessary, we should maintain the test.

mdhont: @FraPazGal, the spark cli is included in the upstream tarball, it's a PHP script that runs codeigniter itself:

```
$app = Config\Services::codeigniter();
$app->initialize();
```
So I think we can be certain that the version will always coincide with the application's version.

.vib/codeigniter/goss/goss.yaml
```diff
@@ -0,0 +1,12 @@
...
+  ../../codeigniter/goss/codeigniter.yaml: {}
+  # Goss tests exclusive to the current container
```
FraPazGal, line 4: Duplicated
```suggestion
```

.vib/codeigniter/goss/goss.yaml
```diff
@@ -0,0 +1,12 @@
...
+  # Load scripts from .vib/common/goss/templates
+  ../../common/goss/templates/check-app-version.yaml: {}
+  ../../common/goss/templates/check-broken-symlinks.yaml: {}
+  ../../common/goss/templates/check-ca-certs.yaml: {}
+  ../../common/goss/templates/check-binaries.yaml: {}
+  ../../common/goss/templates/check-directories.yaml: {}
+  ../../common/goss/templates/check-linked-libraries.yaml: {}
+  ../../common/goss/templates/check-sed-in-place.yaml: {}
```
FraPazGal, line 12: Could you please update your from master and add here the new genera test `check-spdx` added at  https://github.com/bitnami/containers/pull/26723? I would be great if you could add it to each of your opened PRs, thanks.

FraPazGal (commented):
Commented a few minor things, overall the PR looks good!

.vib/codeigniter/goss/goss.yaml
mdhont: I will add the check-spdx for all PR's at once on a later moment

FraPazGal (commented):
LGTM but you need to remove the VIB trigger

FraPazGal (approved):
LGTM!

PR #24336, opened 2023-02-17
[bitnami/pgbouncer] Allow override of PGBOUNCER_LOG_FILE

linosgian:
### Description of the change

As per PgBouncer's [docs](https://www.pgbouncer.org/config.html#generic-settings) in order to disable logging to a file, one should leave `logfile` unset. That option is set to `$PGBOUNCER_LOG_DIR/pgbouncer.log` without any way to override it through env vars.

This PR falls back to the old default value, unless `PGBOUNCER_LOG_FILE` is explicitly set.

I've locally tested that this works by setting `PGBOUNCER_LOG_FILE=""` like so:
```diff
--- a/bitnami/pgbouncer/docker-compose.yml
+++ b/bitnami/pgbouncer/docker-compose.yml
@@ -14,6 +14,7 @@ services:
     environment:
       - POSTGRESQL_HOST=postgresql
       - PGBOUNCER_AUTH_TYPE=trust
+      - PGBOUNCER_LOG_FILE=""
 volumes:
   postgresql_data:
     driver: local
```
### Benefits

Downstream users of this image will be able to disable logging to file.
This is especially problematic for long-running pgbouncer containers. 
The ever-growing log file will eventually be a problem.


linosgian:
@fmulero bump, in case this was forgotten

jbaanus:
bump!

fmulero (approved):
Sorry for my very late response

Thank you very much @linosgian  for your contribution. LGTM

PR #24937, opened 2023-02-23
[bitnami/mastodon] Fix to allow connecting to Elasticsearch host over TLS

### LINKED ISSUE #15087: [bitnami/mastodon] TLS security for Elasticsearch is not functional via the Mastodon chart
### Name and Version

bitnami/mastodon 1.0.3

- Enable security settings for Elasticsearch either via the built-in deployment or an external Elasticsearch deployment.
- Deploy the Mastodon Helm chart. The init script will fail on the "Waiting for Elasticsearch to be ready" step.

I tried with both the built-in and an external Elasticsearch configuration (in separate deploys, of course).

External using the recommended approach of including `https://` in the hostname (see my additional info section below):
```
externalElasticsearch:
  host: https://elasticsearch
  port: 9200
  existingSecret: elasticsearch-secret
  existingSecretPasswordKey: elasticsearch-password
```

External with no protocol specified in the hostname:
```
externalElasticsearch:
  host: elasticsearch
  port: 9200
  existingSecret: elasticsearch-secret
  existingSecretPasswordKey: elasticsearch-password
```

Or built-in:
```
elasticsearch:
  enabled: true

  sysctlImage:
    enabled: true

  security:
    enabled: true
    tls:
      restEncryption: true
      autoGenerated: true

  master:
    replicaCount: 1

  coordinating:
    replicaCount: 1

  data:
    replicaCount: 1

  ingest:
    replicaCount: 1

  service:
    ports:
      restAPI: 9200
```

The init script should be able to connect to the Elasticsearch instance over TLS using password authentication.

### What do you see instead?

The init script is unable to connect to the Elasticsearch instance.

### Additional information

I believe the exact cause of this issue is that the Bitnami Mastodon init script is hard-coded to connect to Elasticsearch using `http://`. Instead, it should connect using `https://` whenever a password is utilized.

Elasticsearch cannot be deployed (anymore? it's possible that this was not always the case) with password authentication unless TLS is enabled, so it should be safe to assume that `https://` is being used. However, a better alternative would probably be to not specify a protocol at all and just let the user optionally include one in their Elasticsearch `host` config value. Using the `ES_HOST` environment variable is the current recommended way to enable TLS support for Mastodon instances connecting to Elasticsearch, as described in this issue: https://github.com/mastodon/mastodon/issues/19402

I have a PR out for this chart (#15076) which adds features and fixes a number of other minor issues, but I was not able to fix this Elasticsearch problem because I believe it will require changes to both the chart _and_ the actual Mastodon image to tweak the init script itself.

ajguyot:
### Description of the change

- Only prefix the Elasicsearch host with a protocol if it doesn't already contain one.

### Benefits

This change should allow Mastodon instances to communicate with Elasticsearch even when TLS is enabled. The current version hard-codes the `http://` protocol, so this change simply removes that protocol when an Elasticsearch Host env variable already includes it.

The recommended way to make TLS connections for Elasticsearch function in Mastodon is simply to prefix the ES Host with `https://` (as discussed here: https://github.com/mastodon/mastodon/issues/19402), so this change should make that option available.

### Possible drawbacks

None. If no protocol is included in the Host env variable then the existing behavior of automatically prefixing `http://` is maintained.

### Applicable issues

- fixes https://github.com/bitnami/charts/issues/15087

### DIFF
--- a/bitnami/mastodon/4/debian-11/rootfs/opt/bitnami/scripts/libmastodon.sh
+++ b/bitnami/mastodon/4/debian-11/rootfs/opt/bitnami/scripts/libmastodon.sh
@@ -431,7 +431,10 @@ mastodon_initialize() {
         # Elasticsearch is an optional component in Mastodon. It is necessary for enabling
         # text searches
         if is_boolean_yes "$MASTODON_ELASTICSEARCH_ENABLED"; then
-            local -r elasticsearch_connection_string="http://${MASTODON_ELASTICSEARCH_HOST}:${MASTODON_ELASTICSEARCH_PORT_NUMBER}"
+            # Only prefix the ES host with a protocol if it doesn't already contain one
+            [[ $MASTODON_ELASTICSEARCH_HOST != "http"* ]] && local -r protocol="http://"
+
+            local -r elasticsearch_connection_string="${protocol}${MASTODON_ELASTICSEARCH_HOST}:${MASTODON_ELASTICSEARCH_PORT_NUMBER}"
             mastodon_wait_for_elasticsearch_connection "$elasticsearch_connection_string"
             if is_boolean_yes "$MASTODON_MIGRATE_ELASTICSEARCH"; then
                 info "Migrating Elasticsearch"


rafariossaa (changes requested):
This environment variable is used in line 83 to check the hostname. If the protocol is added then the check is going to fail.
My suggestion is to add a new environment variable to set the protocol, and concatenate it when setting `elasticsearch_connection_string`

CeliaGMqrz:
Hi @ajguyot 

Thanks for your contribution! Please check [#1328074951](https://github.com/bitnami/containers/pull/24937#pullrequestreview-1328074951)

PR #24938, opened 2023-02-23
[bitnami/mongodb-sharded] Fix mongodb-sharded cannot create user root when initialize in slow machines

dung-tien-nguyen:
### Description of the change

When create a shard cluster using `bitnami/mongodb-sharded` in slow machines (do not meet the required amount of ram and cpu or running on vCenters with high overcommit rate or running with hard disk slow read and write), primary instances cannot create user root that leads to pod running but replicaset config server and shards cannot init.

Root cause: after function `mongodb_start_bg` in [libmongodb-sharded.sh](https://github.com/bitnami/containers/blob/main/bitnami/mongodb-sharded/7.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb-sharded.sh), while primary instances is in background, function `mongodb_create_users()`  throws an error saying `=> create user root fail`.

Change _grep_ command in `mongodb_is_primary_node_up` function to check output correctly.  Output from `mongodb_is_primary_node_up` function always have "true":

```
ReferenceError: admin is not defined
Current Mongosh Log ID: 63f588cb6992a8943cc6995e
Connecting to:          mongodb://127.0.0.1:27017/?directConnection=true&serverSelectionTimeoutMS=2000&appName=mongosh+1.7.1
Using MongoDB:          6.0.4
Using Mongosh:          1.7.1
```

### Benefits

It's avoid the error described above

### Possible drawbacks

N/A

### Applicable issues

Related to https://github.com/bitnami/charts/issues/13364

### Additional information

I do not use `mongodb_wait_for_primary_node` function because this function require username.

### DIFF
--- a/bitnami/mongodb-sharded/5.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb.sh
+++ b/bitnami/mongodb-sharded/5.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb.sh
@@ -682,6 +682,11 @@ mongodb_create_user() {
 mongodb_create_users() {
     info "Creating users..."
 
+    if ! retry_while "mongodb_is_primary_node_up 127.0.0.1 $MONGODB_PORT_NUMBER admin" "$MONGODB_MAX_TIMEOUT"; then
+        error "Unable to initialize primary config server: cannot become primary"
+        exit 1
+    fi
+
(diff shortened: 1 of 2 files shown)


Mauraza:
Hi @dung-tien-nguyen,

I will do some tests and after that, I will update this thread with the results.

doradlavamsi:
Observing the same issue on 5.0.14 as well, Can this be backported to that? @Mauraza @dung-tien-nguyen  ?

doradlavamsi (approved):

Mauraza:
Hi @dung-tien-nguyen,

I'm getting an error when I try MongoDB with this change the primary is not working as it used to... . Could you check it?

PR #25331, opened 2023-02-24
[bitnami/keycloak] Upgrade java version in keycloak image from 11 to 17

vruum:
### Description of the change

Install java version java-17.0.6-10-2 instead of the previous java-11.0.18-10-2 in the docker image

### Benefits
Keycloak extensions built with a newer version of java can be loaded by the container. 
Additionally, keycloak has deprecated support for java 11, and plans to no longer support it with version 22 https://www.keycloak.org/docs/latest/release_notes/index.html#java-11-support-for-keycloak-server-deprecated

### Possible drawbacks

none

### Applicable issues

- fixes #

### DIFF
--- a/bitnami/keycloak/20/debian-11/Dockerfile
+++ b/bitnami/keycloak/20/debian-11/Dockerfile
@@ -24,7 +24,7 @@ RUN install_packages ca-certificates curl krb5-user libaio1 procps zlib1g
 RUN mkdir -p /tmp/bitnami/pkg/cache/ && cd /tmp/bitnami/pkg/cache/ && \
     COMPONENTS=( \
       "wait-for-port-1.0.6-1-linux-${OS_ARCH}-debian-11" \
-      "java-11.0.18-10-2-linux-${OS_ARCH}-debian-11" \
+      "java-17.0.6-10-2-linux-${OS_ARCH}-debian-11" \
       "keycloak-20.0.5-1-linux-${OS_ARCH}-debian-11" \
       "gosu-1.16.0-2-linux-${OS_ARCH}-debian-11" \
     ) && \


jotamartos (approved):
Thanks for your contribution! Our pipeline will release a new version of the container with this update soon.

JoniJnm:
i'm using the keycloak bitnami helm chart. Is the problem there so?

vruum:
Sorry, I thought the SO question was by you also, so thought you were trying to run the same project mentioned there.
You'll need to provide the nashorn engine yourself, along with the scripts you're trying to run. How to do that depends on the scripts you're running and how you're providing them to keycloak.

JoniJnm:
I don't understand the problem, I'm not a Java developer. I just use keycloak with the chart. But ok, i will stay on the version 13.1.1. At least this change shoud be a mayor version

vruum:
@JoniJnm I'm sorry if this change broke anything for you. 
The issue is, in short, that java prior to java 15 included a javascript runtime, for later versions applications (or in this case keycloak extensions) will have to provide that runtime themselves. 
This should only affect you if you're using any javascript keycloak extensions. You don't really provide a lot of details about how you're running keycloak but if this affects the basic usage of the image or the bitnami helm chart, I think one of us should probably file an issue here.

JoniJnm:
It's not only for me. Keycloak creates a default JS policy. So nashorn must be incluted in the image or the bitnami chart needs changes.

I created a issue in bitnami repo with steps to reproduce the problem: https://github.com/bitnami/charts/issues/15659

JoniJnm:
@jotamartos @javsalgar mi inglés es muy malo y quizá no me entiende. Uso la versión normal del chart, no tengo nada personalizado. No soy experto en keycloak pero cuando se crea un nuevo cliente con autorización (al menos desde el chart de bitnami) se crea un policy que ya de por sí usa javascript. Así que creo que nashorn tiene que estar en la imagen si se quiere actualizar a java 17.

vruum:
@JoniJnm just fyi, I posted a potential fix for the image here https://github.com/bitnami/containers/issues/28129, I'm not going to make another pull request, unless I get some kind of feedback, just a bit nervous about breaking stuff.

javsalgar:
Thank you so much! The team will take a look

JoniJnm:
@vruum @javsalgar the change https://github.com/bitnami/containers/commit/50aa7cb50329b9631ffcacd3b42fa111eaf32ef0 works for me

dalbani:
This change broke (part of) my Keycloak deployment.
Like the session page in the admin UI in particular, which now throws:
```
ERROR [org.infinispan.CLUSTER] (jgroups-92,keycloak-0-9382) ISPN000474: Error processing request 1474@keycloak-2-44015: java.lang.NoClassDefFoundError: Could not initialize class org.jboss.marshalling.river.ClassDescriptors
```
I got the hint about a change in the JVM version in https://keycloak.discourse.group/t/not-exist-classdescriptors/19280.

I'm more than surprised that this change was pretty much hidden in a tiny version update of the image: `20.0.5-debian-11-r3` -> `20.0.5-debian-11-r4`.
And now that Keycloak 20 is considered deprecated by Bitnami, it's not even possible to suggest a rollback of this change.
I just don't understand why this change didn't take place in the image for Keycloak 21 only??

carrodher:
Keycloak 20 was deprecated a couple of months ago, so there is not much difference between using `20.0.5-debian-11-r3` or `20.0.5-debian-11-r4` or any other pinned tag because they won't receive updates anymore

Issue #24146, opened 2023-02-16
[bitnami/solr] Multiple SOLR_CORES with different SOLR_CORE_CONF_DIR is impossible


sylvanr:
### Name and Version

bitnami/solr:latest

### What steps will reproduce the bug?

I've got a docker compose setup for the latest bitnami/solr image, and I'm trying to start up two cores, with two different configurations as they have different content in their managed-schema specifically. 
```
solr:
    container_name: solr
    restart: always
    image: bitnami/solr:latest
    ports:
      - 8983:8983
    environment:
      # core setup
      - SOLR_CORES=activity,dataset
      - SOLR_CORE_CONF_DIR=/core_conf/activity,/core_conf/dataset
    volumes:
      - './activity/conf:/core_conf/activity'
      - './dataset/conf:/core_conf/budget'
```
where the core_conf is mounted with those different configurations.

Of course if I remove one of the list of conf dir's, this works, but both dataset and activity cores have the same config, which makes this unusable for now

### What is the expected behavior?

Two cores to be created with different configs as provided by the user.

### What do you see instead?

This is not possible and the process crashes

### Additional information

I'm not entirely sure this is a bug, it might be intended behaviour, so this might be a feature request? But for me logically if you are able to add different cores through a list, you should be able to pass a list of configurations as well, matching those cores.

If anyone knows whether or not this is intended, and if there is anything i can do to bypass this other than completely building the image from scratch, let me know!

javsalgar:
Hi,

Thank you so much for opening the ticket! It seems to me that this is a feature request in this case.

Looking at the code

```
    info "Creating cores..."
    for core in "${cores[@]}"; do
        mkdir -p "${SOLR_SERVER_DIR}/solr/${core}/data"
        mkdir -p "${SOLR_SERVER_DIR}/solr/${core}/conf"
        cp -Lr "${SOLR_CORE_CONF_DIR}"/* "${SOLR_SERVER_DIR}/solr/${core}/conf/"
```

It is assuming all the cores have the same configuration. Maybe it would make sense to allow the list of folders just as you mentioned. If you want to speed up the process, would you like to submit a PR with the functionality?

PR #25657, opened 2023-02-27
[bitnami/solr] Multiple SOLR_CORES with different SOLR_CORE_CONF_DIR

sylvanr:
implements #24146, this allows the user to provide a SOLR_CORE_CONF_DIR for each SOLR_CORE.

This change still allows the user to provide a single directory for each core, and the default will still be used when the conf dir is not provided.


CeliaGMqrz:
Hi @sylvanr, 

Thanks for your contribution! Since `SOLR_CORE_CONF_DIR` and `SOLR_CORES` are input elements provided by the user, could you add them to the [`solr_validate()`](https://github.com/bitnami/containers/blob/main/bitnami/solr/8/debian-11/rootfs/opt/bitnami/scripts/libsolr.sh#L88) function?

We welcome your feedback.

PR #26122, opened 2023-03-03
[bitnami/rails] Add Rails container VIB tests

mdhont:
### Description of the change

Adding the VIB tests for Rails container


.vib/rails/goss/rails.yaml
```diff
@@ -0,0 +1,12 @@
+command:
```
FraPazGal, line 1: I would add a check for the database-related gems added by us at compilation time.

.vib/rails/goss/vars.yaml
```diff
@@ -0,0 +1,12 @@
+binaries:
```
FraPazGal, line 1: You are missing a lot of subcomponents added in the Dockerfile. Please check the [subcomponent list](https://github.com/bitnami/containers/blob/main/bitnami/rails/7/debian-11/prebuildfs/opt/bitnami/.bitnami_components.json).

.vib/rails/goss/rails.yaml
```diff
@@ -0,0 +1,12 @@
...
+  # A new Rails project can be created
+  check-project:
```
FraPazGal, line 3: Nice test, though I would include a server initialization here.

.vib/rails/goss/vars.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - rails
+  - ruby
+directories:
+  - mode: "0755"
+    paths:
+      - /opt/bitnami/ruby/lib
+      - /opt/bitnami/ruby/bin
```
FraPazGal, line 8: I don't think we are modifying either of these folders' permissions. We are only doing it so for [/app](https://github.com/bitnami/containers/blob/main/bitnami/rails/7/debian-11/rootfs/opt/bitnami/scripts/rails/postunpack.sh), which should be included here.

FraPazGal (changes requested):
I added some comments here @mdhont, please check them when you can.

.vib/rails/goss/rails.yaml
mdhont: I've added server initialization, it needs to be in the same test, as Goss tests aren't executed in the listed order.

.vib/rails/goss/vars.yaml
```diff
@@ -0,0 +1,20 @@
+binaries:
+  - gosu
+  - mysql
+  - node
+  - python
+  - rails
+  - ruby
+  - wait-for-port
```
FraPazGal, line 8: We usually do not add `wait-for-port` here, as we understand that it is only a minor tool used at initialization time. As such, the existence of this binary is implicitly tested when the container is fully deployed. No strong thought on this though, wdyt @joancafom?

.vib/rails/goss/rails.yaml
```diff
@@ -0,0 +1,21 @@
...
+    timeout: 120000
+    exec: "rails new /app && cd /app && timeout 10s rails s -p 4000 || true"
+    exit-status: 0
+    stdout:
+      - Listening
+  check-app-version:
+  # rails --version returns other version format than APP_VERSION
```
FraPazGal, line 10: It'll be great if you could expand this comment a bit and give an overview of the syntax used by the asset here.

FraPazGal (commented):
Added a couple more comments!

FraPazGal (approved):
LGTM!

PR #26436, opened 2023-03-06
[bitnami/airflow-scheduler] Add Kerberos package libkrb5-dev to airflow-scheduler container Dockerfile

interniq:
Description of the change
Change adds the Kerberos development package libkrb5-dev to image.

Benefits
Adding the libkrb5-dev package is required for requests/HTTPKerberosAuth Kerberos support/installation.

Possible drawbacks
Minimal increase in the size of the resulting image.

Applicable issues
Airflow has support for Kerberos. This means that Airflow can renew Kerberos tickets for itself and store them in its ticket cache. The hooks and dags can make use of tickets to authenticate against Kerberized services. The libkrb5-dev supplied utilities (krb5-config) and libraries are required for this.

Additional information
https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/security/kerberos.html

### DIFF
--- a/bitnami/airflow-scheduler/2/debian-11/Dockerfile
+++ b/bitnami/airflow-scheduler/2/debian-11/Dockerfile
@@ -19,7 +19,7 @@ ENV HOME="/" \
 COPY prebuildfs /
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 # Install required system packages and dependencies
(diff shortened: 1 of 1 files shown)


aoterolorenzo:
Hi @interniq,

Is `libgss-dev` also neccessary to use the HTTPKerberosAuth? Whats the motivation behind that second dependency? Also, would be very appreciated a composer or other environment setup to properly reproduce the issue.

interniq:
> 



> Hi @interniq,
> 
> Is `libgss-dev` also neccessary to use the HTTPKerberosAuth? Whats the motivation behind that second dependency? Also, would be very appreciated a composer or other environment setup to properly reproduce the issue.

> a composer or other environment setup
No problem, could you advise on what's required here, I'm not familiar with composer?

regarding libgss-dev, I'll amend the commit -- you're correct,  that's not required for HTTPKerberosAuth

PR #26728, opened 2023-03-08
[bitnami/containers] Add markdown linter

fmulero:
### Description of the change

Add new workflow to run markdown linter on PRs with modifications in markdown files. 

### Benefits

Prevent new contributions to modify the markdown files in the wrong way.

### Possible drawbacks

None identified.

### Additional information

Similar PRs:
* https://github.com/bitnami/vms/pull/747
* https://github.com/bitnami/charts/pull/15338

Execution examples:
* https://github.com/bitnami/charts/pull/15379/commits/c826c77465d36be3e8da05ef807aaaaf6dfb68b5

### DIFF
--- a/.github/workflows/markdown-linter.yml
+++ b/.github/workflows/markdown-linter.yml
@@ -1,0 +1,49 @@
+name: '[CI/CD] Markdown linter'
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - '**.md'
+permissions:
+  contents: read
+jobs:
+  markdown-linter:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install mardownlint
+        run: npm install -g markdownlint-cli@0.33.0
+      - name: Checkout project
+        uses: actions/checkout@v3
+      - name: Execute markdownlint
+        run: |
+          # Using the Github API to detect the files changed as git merge-base stops working when the branch is behind
+          # and jitterbit/get-changed-files does not support pull_request_target
+          URL="https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files"
+          files_changed_data=$(curl -s --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' -X GET -G "$URL")
+          files_changed="$(echo "$files_changed_data" | jq -r '.[] | .filename')"
+          md_files="$(echo "$files_changed" | grep -o ".*\.md$" | sort | uniq || true)"
+          # Create an empty file, useful when the PR changes ignored files
+          touch ${{runner.temp}}/output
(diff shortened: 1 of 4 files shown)


.github/workflows/markdown-linter.yml
```diff
@@ -0,0 +1,49 @@
...
+          # Using the Github API to detect the files changed as git merge-base stops working when the branch is behind
+          # and jitterbit/get-changed-files does not support pull_request_target
+          URL="https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files"
+          files_changed_data=$(curl -s --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' -X GET -G "$URL")
+          files_changed="$(echo "$files_changed_data" | jq -r '.[] | .filename')"
+          md_files="$(echo "$files_changed" | grep -o ".*\.md$" | sort | uniq || true)"
+          # Create an empty file, useful when the PR changes ignored files
+          touch ${{runner.temp}}/output
+          exit_code=0
+          markdownlint -o ${{runner.temp}}/output ${md_files[@]} || exit_code=$?
```
andresbono, line 29: Why is there an array expansion in `${md_files[@]}`?

Would it be a problem if we have any path with whitespaces? I guess we don't have to care about that case anyways.

.github/workflows/markdown-linter.yml
```diff
@@ -0,0 +1,49 @@
...
+            elif [[ "${#strarr[@]}" -eq 2 ]]; then
+              echo "::warning file=${strarr[0]},line=${strarr[1]}::${message}"
+            else
+              echo "::warning:: Error processing: ${line}"
+            fi
+          done < ${{runner.temp}}/output
+          if [[ $exit_code -ne 0 ]]; then
+            echo "::error:: Please review linter messages"
+            exit "$exit_code"
+          fi
```
andresbono, line 49: Should we add the new line at the end of file?

andresbono (approved):
Very nice, LGTM! 👏 

I just left a couple of questions, nothing important as we already know this works as expected in bitnami/charts, right?

Also, if we now run this linter on all the existing markdown files, does it return any warning/error? To avoid users having to fix old issues not related to their changes.

.github/workflows/markdown-linter.yml
fmulero: Not really needed

.github/workflows/markdown-linter.yml
fmulero: The first try was using `'**/*.md'` but it took so long. Users mainly touch one or two files so I decided to calculate and pass them to the linter.

About the whitespaces, we might run into troubles if we had them

PR #26791, opened 2023-03-09
[bitnami/git] Add VIB tests to Git container

mdhont:
### Description of the change

Adding the VIB tests to the Git container

### DIFF
--- a/.vib/git/goss/git.yaml
+++ b/.vib/git/goss/git.yaml
@@ -1,0 +1,7 @@
+command:
+  check-git-init:
+    exec: git init /tmp/test
+    exit-status: 0
+  check-git-clone:
+    exec: git clone https://github.com/bitnami/bncert.git
+    exit-status: 0
--- a/.vib/git/goss/git.yaml
+++ b/.vib/git/goss/git.yaml
@@ -1,7 +1,7 @@
 command:
   check-git-init:
-    exec: git init /tmp/test
+    exec: git init $RANDOM
     exit-status: 0
   check-git-clone:
-    exec: git clone https://github.com/bitnami/bncert.git
+    exec: git clone https://github.com/bitnami/bncert.git $RANDOM
     exit-status: 0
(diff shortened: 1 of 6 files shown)


.vib/git/goss/goss.yaml
```diff
@@ -0,0 +1,10 @@
+gossfile:
+  # Goss tests exclusive to the current container
+  ../../git/goss/git.yaml: {}
+  # Load scripts from .vib/common/goss/templates
+  ../../common/goss/templates/check-app-version.yaml: {}
+  ../../common/goss/templates/check-binaries.yaml: {}
+  ../../common/goss/templates/check-broken-symlinks.yaml: {}
+  ../../common/goss/templates/check-ca-certs.yaml: {}
+  ../../common/goss/templates/check-linked-libraries.yaml: {}
+  ../../common/goss/templates/check-sed-in-place.yaml: {}
```
FraPazGal, line 10: Please add `check-spdx.yaml`

.vib/git/goss/git.yaml
```diff
@@ -0,0 +1,7 @@
+command:
```
FraPazGal, line 1: These tests don't seem to be retriable.

mdhont: I've added a random folder

FraPazGal: Have you tested this? IIRC, using `$RANDOM` was not possible in Goss (note it is a bash function, not and env). Please review this; the alternative we have always used here is the go template function `randAlpha`.

bitnami/git/2/debian-11/Dockerfile
```diff
@@ -17,7 +17,7 @@ ENV OS_ARCH="${TARGETARCH:-amd64}" \
 
 COPY prebuildfs /
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
-# Install required system packages and dependencies
+## Install required system packages and dependencies
```
joancafom, line 20: Please remove this change

joancafom (approved):

PR #27115, opened 2023-03-12
Document `POSTGRESQL_PORT_NUMBER`

### LINKED ISSUE #24930: [bitnami/postgresql] Unable to work around the `EXPOSE` keyword in the postgresql `Dockerfile` in GitLab services
### Name and Version

bitnami/postgresql:latest

1. Spin up a [GitLab Service](https://docs.gitlab.com/ee/ci/services/) using the `bitnami/postgresql` image
2. Spin up a second GitLab Service using the `bitnami/postgresql` image
  * Attempt to set either `POSTGRESQL_PORT_NUMBER=5433` (undocumented) or `POSTGRESQL_EXTRA_FLAGS='--port=5433'`
3. Attempt to connect to both services

I will be able to successfully use both postgresql services on different ports

### What do you see instead?

The container that I attempt to move to `5433` cannot be connected to.

### Additional information

I believe that this is because the GitLab Services are using the `EXPOSE` hint in the `Dockerfile` to limit the exposed ports.

I have not seen the same behavior with the `bitnami/keycloak` image which does not appear to have the `EXPOSE` keyword set.

trevor-vaughan:
### Description of the change

Documents the `POSTGRESQL_PORT_NUMBER` option.

### Benefits

Users know how to change the port number in a supported manner.

### Possible drawbacks

None

### Applicable issues

- fixes #24930

### Additional information

N/A

### DIFF
--- a/bitnami/postgresql/README.md
+++ b/bitnami/postgresql/README.md
@@ -269,6 +269,7 @@ The Bitnami PostgreSQL Image comes with the pgAudit module enabled by default. T
 
 The Bitnami PostgreSQL Image allows configuring several parameters for the connection and session management:
 
+* `POSTGRESQL_PORT_NUMBER`: Set the listening port number
 * `POSTGRESQL_USERNAME_CONNECTION_LIMIT`: If a user different from `postgres` is created, set the connection limit. No defaults.
 * `POSTGRESQL_POSTGRES_CONNECTION_LIMIT`: Set the connection limit for the `postgres` user. No defaults.
 * `POSTGRESQL_STATEMENT_TIMEOUT`: Set the statement timeout. No defaults.


bitnami/postgresql/README.md
```diff
@@ -269,6 +269,7 @@ The Bitnami PostgreSQL Image comes with the pgAudit module enabled by default. T
 
 The Bitnami PostgreSQL Image allows configuring several parameters for the connection and session management:
 
+* `POSTGRESQL_PORT_NUMBER`: Set the listening port number
```
mdhont, line 272: Could you change the description? 

```suggestion
* `POSTGRESQL_PORT_NUMBER`: PostgreSQL port number. Defaults to 5432.
```

mdhont (changes requested):

PR #27474, opened 2023-03-15
[bitnami/ksql] Add ksql container tests

mdhont:
Adding VIB test for Ksql container

### DIFF
--- a/.vib/ksql/goss/goss.yaml
+++ b/.vib/ksql/goss/goss.yaml
@@ -1,0 +1,9 @@
+gossfile:
+  # Load scripts from .vib/common/goss/templates
+  ../../common/goss/templates/check-app-version.yaml: {}
+  ../../common/goss/templates/check-binaries.yaml: {}
+  ../../common/goss/templates/check-broken-symlinks.yaml: {}
+  ../../common/goss/templates/check-ca-certs.yaml: {}
+  ../../common/goss/templates/check-directories.yaml: {}
+  ../../common/goss/templates/check-linked-libraries.yaml: {}
+  ../../common/goss/templates/check-sed-in-place.yaml: {}
--- a/.vib/ksql/goss/goss.yaml
+++ b/.vib/ksql/goss/goss.yaml
@@ -2,7 +2,6 @@ gossfile:
   # Goss tests exclusive to the current container
   ../../ksql/goss/ksql.yaml: {}
   # Load scripts from .vib/common/goss/templates
-  ../../common/goss/templates/check-app-version.yaml: {}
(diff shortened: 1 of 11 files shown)


.vib/ksql/goss/vars.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - /opt/bitnami/ksql/etc/ksqldb
+      - /bitnami/ksql/data
+      - /opt/bitnami/ksql/logs
+      - /opt/bitnami/ksql/certs
+root_dir: /opt/bitnami
+# ksql does not have a version flag so we initiate the shell
+# which does show the version.
+version:
+  bin_name: timeout
+  flag: 5s ksql
```
FraPazGal, line 25: Please use a custom `command` test in `ksql.yaml` instead of using the `check-app-version.yaml` template. Even though we are adding a comment, this use of the template is confusing.

.vib/ksql/goss/vars.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - java
+  - ksql
+  - ksql-datagen
+  - ksql-print-metrics
+  - ksql-run-class
+  - ksql-server-start
+  - ksql-server-stop
+  - ksql-stop
+  - ksql-test-runner
+  - ksql-restore-command-topic
```
FraPazGal, line 12: I'm not a fan of checking every single binary here, as we are introducing flakiness future-wise.

.vib/ksql/goss/goss.yaml
FraPazGal, line 1: We should create a custom test at `ksql.yaml` to check the default config file created at the postunpack stage.

FraPazGal (changes requested):
I have some concerns with the current tests, could you please check my comments @mdhont ?

.vib/ksql/goss/ksql.yaml
```diff
@@ -0,0 +1,18 @@
+file:
+  # Checks default config for erlang is set
+  {{ .Vars.root_dir }}/ksql/etc/ksqldb/ksql-server.properties.default:
+    exists: true
+    filetype: file
+    contains:
+      - /http://0.0.0.0:8088/
```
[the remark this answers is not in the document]
joancafom: You don't need a regex for this check, you can use a string. Only use regex when you plan to use their capabilities.

joancafom (changes requested):
Thanks for your PR! I left some comments with my thougths

FraPazGal (approved):
LGTM!

joancafom (approved):

Issue #26480, opened 2023-03-06
SonarQube 8.9.10 (ARM64) does not start up on M1 Mac


ajeetdsouza:
### Name and Version

bitnami/sonarqube:8.9.10

### What architecture are you using?

arm64

### What steps will reproduce the bug?

`docker-compose up` with the following script:

```yaml
version: '3'
services:
  postgresql:
    image: docker.io/bitnami/postgresql:13
    environment:
      - ALLOW_EMPTY_PASSWORD=yes
    volumes:
      - 'postgresql_data:/bitnami/postgresql'
  sonarqube:
    image: docker.io/bitnami/sonarqube:8
    ports:
      - '9000:9000'
    environment:
      - ALLOW_EMPTY_PASSWORD=yes
      - POSTGRESQL_HOST=postgresql
      - POSTGRESQL_ROOT_USER=postgres
      - POSTGRESQL_CLIENT_CREATE_DATABASE_NAME=bitnami_sonarqube
      - POSTGRESQL_CLIENT_CREATE_DATABASE_USERNAME=bn_sonarqube
      - POSTGRESQL_CLIENT_CREATE_DATABASE_PASSWORD=bitnami1234
      - SONARQUBE_DATABASE_NAME=bitnami_sonarqube
      - SONARQUBE_DATABASE_USER=bn_sonarqube
      - SONARQUBE_DATABASE_PASSWORD=bitnami1234
    volumes:
      - sonarqube_data:/bitnami
volumes:
  sonarqube_data:
    driver: local
  postgresql_data:
    driver: local
```

### What is the expected behavior?

We don't seem to have a script to support ARM64. If we're not supporting ARM64, we should not publish broken ARM64 images. If we do support ARM64, we should add a script that works.

### What do you see instead?

```
sonarqube-sonarqube-1   | grep: /opt/bitnami/sonarqube/logs/sonar.log: No such file or directory
sonarqube-sonarqube-1   | grep: /opt/bitnami/sonarqube/logs/sonar.log: No such file or directory
sonarqube-sonarqube-1   | sonarqube 19:15:10.29 ERROR ==> Could not find entry "SonarQube is (up|operational)" in 
...
/opt/bitnami/sonarqube/logs/sonar.log after 300 retries
```

Upon investigation, I found the actual source of the error. We have a directory called `/opt/bitnami/sonarqube/bin/` where we store startup scripts for different systems. I'm seeing only the following 3 - ARM64 is not present here!

- `linux-x86-64`
[...]

PR #27514, opened 2023-03-16
[bitnami/sonarqube] Remove ARM achitecture

fmulero:
### Description of the change

Remove support for ARM64.

### Benefits

SonarQube binary doesn't include ARM launchers/wrappers, so we can not support ARM64 at the moment.

### Applicable issues

- Related to #26480


marcosbc (approved):
LGTM!

PR #27522, opened 2023-03-17
[bitnami/*] Pass VIB pipelines through linter and formatter

FraPazGal:
### Description of the change

This PR gets our VIB pipelines `.json` files linter-compliant and properly formatted.

Using `jq` I found the unquoted value `{VIB_ENV_ROLLING_TAGS}` returned a parsing error. After fixing that, `jq` formatted the pipelines properly, fixing minor issues introduced recently.

### Benefits

Removes parsing and format issues.

### Possible drawbacks

None


.vib/acmesolver/vib-publish.json
```diff
@@ -78,14 +78,14 @@
           "action_id": "container-image-publish",
           "params": {
             "metadata": {
-              "tags": {VIB_ENV_ROLLING_TAGS}
+              "tags": "{VIB_ENV_ROLLING_TAGS}"
```
fmulero, line 81: Have you tested it? `tags` requires an array and you are passing a string

fmulero (commented):
Thanks a lot @FraPazGal!

I have one doubt, Could you give it a glance?

.vib/acmesolver/vib-publish.json
FraPazGal: You are right, this is failing the DSL schema validation, as it expects an array. I don't see an easy solution, we would need to add or parse the `tags` field directly in the CD pipeline to avoid this issue.

I will revert this change back and make do with the parsing error, IMO it is not critical enough to obscure part of the DSL by adding logic elsewhere. wdyt @fmulero, @joancafom?

fmulero: LGTM. We can survive at the moment. Also we have plans to change these files in order to reuse as much as possible.

fmulero (dismissed):

joancafom (dismissed):

joancafom (approved):

PR #27523, opened 2023-03-17
[bitnami/mastodon] Fix Import functionality for user data

joelparker:
The official mastodon image includes file and bitnami/mastodon does not. This leads to import failing for users because the file command is not found:

Command :: file -b --mime '/tmp/86c8109f207a6432520364460861301b20230316-11-iru96r.csv'

### Description of the change

Add the file binary to the docker image for bitnami/mastodon.  This binary is required for user data import to work.

### Benefits

This allows Import of 

- Followers list
- Block list
- Mute list
- Domain Blocking lists
- Bookmarks

### Possible drawbacks

None

### Applicable issues

- fixes #

### Additional information

To reproduce the problem:

- Create a functioning mastodon instance
- From a logged in account go to Preferences->Import and Export->Export
- Export one or more CSV files (Add a follower and export the Follows CSV)
- Import that same CSV file.  It will fail with a message in the UI "has contents that are not what they are reported to be"

The fix is to provide the file binary which is used to verify that the uploaded CSV file is actually a CSV file.  If you exec into the container file is not part of the container image so paperclip fails to execute the binary and the CSV file is rejected.

This pull request fixes this by adding file to the list of installed apt package in the bitnami/mastodon Dockerfile.

### DIFF
--- a/bitnami/mastodon/4/debian-11/Dockerfile
+++ b/bitnami/mastodon/4/debian-11/Dockerfile
@@ -19,7 +19,7 @@ ENV HOME="/" \
 COPY prebuildfs /
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 # Install required system packages and dependencies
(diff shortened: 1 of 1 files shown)


CeliaGMqrz (approved):
Thanks for your contribution! LGTM

PR #27831, opened 2023-03-21
[bitnami/redis] Added dependency for running official Redis module

haysch:
RedisGraph requires OpenMP for C in order to run, which is not part of the current `bitnami/redis` image.

### Description of the change

Include the `libgomp1` package (Debian) in the base `bitnami/redis` image to enable the use of RedisGraph without requiring the need to maintain a separate image.

### Benefits

The RedisGraph module is supported out of the box when using `bitnami/redis` image.
Extends the number of Redis modules that can be run as part of the `bitnami/redis` image to match the modules included in `redis/redis-stack-server`.

### Possible drawbacks

Additional apt package, however, its dependency tree is relatively small and does not depend on any exotic packages, only `gcc-10-base` and `libc6` ([Debian Bullseye libgomp1](https://packages.debian.org/bullseye/libgomp1)).

### Applicable issues

N/A

### Additional information

The change can be tested by running the following docker command assuming that the `redisgraph.so` binary exists:

```bash
cd [6.2|7.0]/debian-11
docker build -t bitnami/redis:redisgraph -f Dockerfile .
docker run -it --rm --name bitnami-redis-modules -v /path/to/redisgraph.so:/opt/bitnami/redis/modules/redisgraph.so bitnami/redis:redisgraph /bin/bash
# inside the bitnami/redis container run the following command
bash run.sh --loadmodule /opt/bitnami/redis/modules/redisgraph.so

# should produce line
Module 'graph' loaded from /opt/bitnami/redis/modules/redisgraph.so
```

### DIFF
--- a/bitnami/redis/6.2/debian-11/Dockerfile
+++ b/bitnami/redis/6.2/debian-11/Dockerfile
@@ -19,7 +19,7 @@ ENV HOME="/" \
 COPY prebuildfs /
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 # Install required system packages and dependencies
-RUN install_packages ca-certificates curl libssl1.1 procps
+RUN install_packages ca-certificates curl libssl1.1 procps libgomp1
 RUN mkdir -p /tmp/bitnami/pkg/cache/ && cd /tmp/bitnami/pkg/cache/ && \
     COMPONENTS=( \
       "wait-for-port-1.0.6-2-linux-${OS_ARCH}-debian-11" \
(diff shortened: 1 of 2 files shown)


jotamartos:
Thanks for the PR. Let us know some time to review it and apply the changes in our system to build the containers with this update.

jotamartos (approved):
Thanks for the PR. We've merged the information in our system and a new version of the containers should be released soon.

jotamartos:
Just a quick note to let you know that version 7.0.10-debian-11-r2 contains the change 

https://github.com/bitnami/containers/pull/28192

PR #27835, opened 2023-03-21
[bitnami/containers] Reduce GitHub API requests

fmulero:
### Description of the change

Modify workflows involved in automated PRs to reduce the number of requests to the GitHub API. With these changes we remove 10 API requests for each automated PR.

### Benefits

Reduce errors due to the API rate limit.

### Possible drawbacks

None identified

### DIFF
--- a/.github/workflows/assign-asset-label.yml
+++ b/.github/workflows/assign-asset-label.yml
@@ -15,11 +15,11 @@ jobs:
         name: Get modified assets
         env:
           DIFF_URL: "${{github.event.pull_request.diff_url}}"
-          TEMP_DIR: "${{runner.temp}}"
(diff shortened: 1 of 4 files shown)


FraPazGal (commented):
The PR LGTM, but I do have a question regarding the card creation for automated PRs. Won't the `move-closed-issues` workflow create a card for the automated PR when it is closed just before removing it after?

fmulero:
> The PR LGTM, but I do have a question regarding the card creation for automated PRs. Won't the `move-closed-issues` workflow create a card for the automated PR when it is closed just before removing it after?

You are the best!!! You are completely right. I've just removed refactor that piece of code in my last commit. With that change we can save maybe 7 or 8 API request per each automated PR. Thanks a lot

joancafom (approved):
LGTM! Thanks for the PR

PR #28158, opened 2023-03-23
[bitnami/redis] Add VIB tests

mdhont:
Signed-off-by: mdhont [michield@vmware.com](mailto:michield@vmware.com)

### Description of the change

The main objective of this PR is to publish our Bitnami MySQL container using VMware Image Builder. In order to do that, several changes are included:

Increasing the existing test coverage of the asset by adding Goss tests.
Update verify and publish VIB pipeline's definitions.

### Benefits
Ensuring higher quality of the container catalog.
Increased pool of assets completely handled by VMware Image Builder.

### Possible drawbacks


.vib/redis/goss/vars.yaml
```diff
@@ -0,0 +1,16 @@
...
+  - gosu
+  - redis-benchmark
+  - redis-cli
+  - redis-server
+  - wait-for-port
+directories:
+  - mode: "0775"
+    paths:
+      - /bitnami/redis/data
+      - /opt/bitnami/redis/etc
```
FraPazGal, line 11: I think there are some inconsistencies here compared to the permissions modifications and folder creations done at the postunpack stage. Could you please review this?

mdhont: I've added the paths of the folders that are created in the postunpack stage.  Although I'm not sure what you mean with inconsistencies in permission modifications.

.vib/redis/goss/redis.yaml
```diff
@@ -0,0 +1,15 @@
+file:
+  /opt/bitnami/redis/etc/redis.conf:
+    exists: true
+    mode: "0664"
+    contains:
+      - port {{ .Vars.port }}
```
joancafom, line 6: Do not use a variable here. You can replace it by the port instead, although I would personally drop this line as there is no real need of checking every modification in the file

.vib/redis/goss/redis.yaml
```diff
@@ -0,0 +1,15 @@
...
+      - dir /bitnami/redis/data
+      - pidfile /opt/bitnami/redis/tmp/redis.pid
+      - daemonize yes
+command:
+  check-redis-server:
+    exec: timeout 5s redis-server || true
```
joancafom, line 12: Why are you using `|| true` here?

.vib/redis/goss/vars.yaml
```diff
@@ -0,0 +1,19 @@
...
+      - /bitnami
+      - /bitnami/redis
+      - /bitnami/redis/data
+      - /opt/bitnami/redis
+      - /opt/bitnami/redis/etc
+root_dir: /opt/bitnami
+version:
+  bin_name: redis-server
+  flag: --version
+port: 6379
```
joancafom, line 19: Remove this line

joancafom (changes requested):
Thanks for your changes! I left some comments as a review

joancafom (changes requested):
Thanks for your changes! I left some comments as a review

.vib/redis/goss/redis.yaml
mdhont: Timeout does not exit with status 0. Using || true, seems cleaner.

.vib/redis/goss/redis.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - /dir.*/bitnami/redis/data/
+      - /pidfile.*/opt/bitnami/redis/tmp/redis.pid/
+      - daemonize yes
+command:
+  check-redis-server:
+    exec: timeout 5s redis-server || true
+    exit-status: 0
+    stdout:
+      - Ready to accept connections
+  check-redis-server-ssl:
```
FraPazGal, line 16: I'm not sure about these 2 ssl checks, there is no related compilation or postunpack logic that requires it right?

.vib/redis/goss/redis.yaml
FraPazGal: If we are configuring Redis to be used as a daemon, couldn't we initialize it so using the `redis.conf` file?

.vib/redis/goss/redis.yaml
mdhont: I've added those based on review comments:

https://github.com/bitnami/containers/pull/28158#discussion_r1149286198

.vib/redis/goss/redis.yaml
mdhont: I've changed it and it's better, as it exits properly, and then we can check if the process is running.

joancafom (approved):
LGTM!

FraPazGal (approved):

PR #28187, opened 2023-03-24
[bitnami/containers] Use latest changes from labeler action

fmulero:
### Description of the change

Take advantage of the latest changes in the labeler action. We are mainly interested on [this change](https://github.com/fmulero/labeler/pull/3).
Remove an unnecessary `remove-labels` parameter in labeler action during triage.

### Benefits

With those changes, the action will fail clearly if API rate limit is reached and the user can retry the actions failed.
Removing the `remove-labels` parameter from the labeler action in the triage workflow, we remove one unnecessary GitHub API request.

### Possible drawbacks

None identified.

### DIFF
--- a/.github/workflows/assign-asset-label.yml
+++ b/.github/workflows/assign-asset-label.yml
@@ -43,6 +43,6 @@ jobs:
             }
       - name: Labeling
         if: ${{ steps.get-asset.outputs.result == 'ok' }}
-        uses: fmulero/labeler@main
+        uses: fmulero/labeler@1.1.0
         with:
           add-labels: "${{ steps.get-asset.outputs.name }}"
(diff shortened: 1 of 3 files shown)


jotamartos:
Isn't better to make a new release of labeler (1.1.0 or 2.0.0) and set that version. That way we will avoid issues if something wrong is commited in the repo.

fmulero:
> Isn't better to make a new release of labeler (1.1.0 or 2.0.0) and set that version. That way we will avoid issues if something wrong is commited in the repo.

Ok, I've just changed it. Iniatially I followed the same approach we are already using with `vmware-image-builder-action`, where we are the main users and maintainers.

jotamartos (approved):

PR #28667, opened 2023-03-30
[bitnami/java] postunpack.sh should be copying the contents of JAVA_EXTRA_SECURITY_DIR, not the directory itself

### LINKED ISSUE #28666: java/postunpack.sh copies JAVA_EXTRA_SECURITY_DIR directory rather than just the files under it
### Name and Version

bitnami/elasticsearch:8.6.2-debian-11-r13 and more

amd64

1.  In bitnami/elasticsearch/8/debian-11/rootfs, create a directory `bitnami/java/extra-security`
2. Put a file in there
3. Build image
4. Start image, running a shell rather than the default CMD, and look in the `/opt/bitnami/java/lib/security` directory

From the `bitnami/java/README.md` file, it says:

> ...it is possible to replace the default truststore located in the `/opt/bitnami/java/lib/security` folder. This is done by setting the `JAVA_EXTRA_SECURITY_DIR` docker build ARG variable, which needs to point to a location that contains a *cacerts* file that would substitute the originally bundled truststore. 

From that, I expected files in `bitnami/java/extra-security` to overwrite, or are added to, the files under `/opt/bitnami/java/lib/security`

For instance, if I create a file:
`rootfs/bitnami/java/extra-security/cacerts`
and build the image, I expect this file to replace the `/opt/bitnami/java/lib/security/cacerts` originally in the image.

### What do you see instead?

The `extra-security` directory, with the files in it, is created as a subdir in `/opt/bitnami/java/lib/security`

In my above example, rather than my `cacerts` file replacing the `/opt/bitnami/java/lib/security/cacerts`, it creates a new `/opt/bitnami/java/lib/security/extra-security/cacerts`.  This seems wrong.

### Additional information

As far I can tell, java is not looking for a lib/security/extra-security directory, and my new CA file is not being used.

I think the solution is, in `rootfs/opt/bitnami/scripts/java/postunpack.sh`, where it reads:
```
if [[ -n "${JAVA_EXTRA_SECURITY_DIR:-}" ]] && ! is_dir_empty "$JAVA_EXTRA_SECURITY_DIR"; then
    info "Adding custom CAs to the Java security folder"
    cp -Lr "$JAVA_EXTRA_SECURITY_DIR" /opt/bitnami/java/lib/security
fi
```
change that  `cp` command to add '/.' to the source, making it read:
```
    cp -Lr "$JAVA_EXTRA_SECURITY_DIR/." /opt/bitnami/java/lib/security
```
thus copying all the files under JAVA_EXTRA_SECURITY_DIR, rather than the directory itself.

This is not just a problem for this particular elasticsearch image builder, but rather all the image builders that utilize this JAVA_EXTRA_SECURITY_DIR mechanism in a `java/postinstall.sh` script. 

I've submitted a PR with a fix for all of them.

QuirkyMarc:
`java/postunpack.sh` should be coping just the contents of JAVA_EXTRA_SECURITY_DIR to `/opt/bitnami/java/lib/security` directory.

Instead, `java/postunpack.sh` copies the directory pointed to by JAVA_EXTRA_SECURITY_DIR as a subdir of `/opt/bitnami/java/lib/security` directory.

For example, using the default `JAVA_EXTRA_SECURITY_DIR` value "/bitnami/java/extra-security", create the file `rootfs/bitnami/java/extra-security/cacerts`, and build the image.  Rather than `/opt/bitnami/java/lib/security/cacerts` being replaced with the new file, a new subdir `/opt/bitnami/java/lib/security/extra-security` directory is created, with the new `cacerts` file under there.

### Description of the change

Change the relevant `cp` commands in all `java/postunpack.sh` in all image builders, from
```
   cp -Lr "$JAVA_EXTRA_SECURITY_DIR" /opt/bitnami/java/lib/security
```
to this, by added a `/.` to the end of the source argument:
```
   cp -Lr "$JAVA_EXTRA_SECURITY_DIR/." /opt/bitnami/java/lib/security
```

### Benefits

Functionality of this mechanism, as described in `bitname/java/README.md`, will work.

### Possible drawbacks

If somebody, somehow is successfully using this mechanism as it exists now, then this change will probably break them.

Like instead of pointing `JAVA_EXTRA_SECURITY_DIR` to a directory, they point it directly to a file, like say:
```
JAVA_EXTRA_SECURITY_DIR=bitnami/java/cacerts
```
This would currently work, primarily because there is no test if JAVA_EXTRA_SECURITY_DIR is actually a dir, just that it exists and is non-empty.  If the file exists, other test is that it's not empty, and `is_dir_empty()` from `scripts/libfs.sh`, returns false if passed a file.  From where this is called in postunpack.sh, a false response causes the cp command to be run.

If this PR is merged, then this change will break that, as "file/." still exists, but `is_dir_empty()` passed "somefile/."` returns true, and thus the file will no longer be copied.

With this PR though, this mechanism will match the documentation and the intended use case it describes.

### Applicable issues

- fixes #28666

### Additional information

🤔

### DIFF
--- a/bitnami/cassandra/4.0/debian-11/rootfs/opt/bitnami/scripts/java/postunpack.sh
+++ b/bitnami/cassandra/4.0/debian-11/rootfs/opt/bitnami/scripts/java/postunpack.sh
@@ -20,5 +20,5 @@ set -o pipefail
 
 if [[ -n "${JAVA_EXTRA_SECURITY_DIR:-}" ]] && ! is_dir_empty "$JAVA_EXTRA_SECURITY_DIR"; then
     info "Adding custom CAs to the Java security folder"
-    cp -Lr "$JAVA_EXTRA_SECURITY_DIR" /opt/bitnami/java/lib/security
+    cp -Lr "$JAVA_EXTRA_SECURITY_DIR/." /opt/bitnami/java/lib/security
 fi
--- a/bitnami/cassandra/4.0/debian-11/rootfs/opt/bitnami/scripts/java/postunpack.sh
+++ b/bitnami/cassandra/4.0/debian-11/rootfs/opt/bitnami/scripts/java/postunpack.sh
@@ -20,5 +20,5 @@ set -o pipefail
 
 if [[ -n "${JAVA_EXTRA_SECURITY_DIR:-}" ]] && ! is_dir_empty "$JAVA_EXTRA_SECURITY_DIR"; then
     info "Adding custom CAs to the Java security folder"
-    cp -Lr "$JAVA_EXTRA_SECURITY_DIR/." /opt/bitnami/java/lib/security
+    cp -Lr "${JAVA_EXTRA_SECURITY_DIR}/." /opt/bitnami/java/lib/security
 fi
--- a/bitnami/cassandra/4.1/debian-11/rootfs/opt/bitnami/scripts/java/postunpack.sh
+++ b/bitnami/cassandra/4.1/debian-11/rootfs/opt/bitnami/scripts/java/postunpack.sh
@@ -20,5 +20,5 @@ set -o pipefail
 
 if [[ -n "${JAVA_EXTRA_SECURITY_DIR:-}" ]] && ! is_dir_empty "$JAVA_EXTRA_SECURITY_DIR"; then
     info "Adding custom CAs to the Java security folder"
-    cp -Lr "$JAVA_EXTRA_SECURITY_DIR" /opt/bitnami/java/lib/security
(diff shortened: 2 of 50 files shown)


QuirkyMarc:
extra commits to fixup that DCO / signoff check, to which, I just followed the instruction given.  Hope that was right.  🤔

andresbono:
Hi @QuirkyMarc, thank you so much for your contribution! We are going to do some additional testing on our side and this PR should be good to go.

I have just pushed a commit on top of yours because `${}` format for env-vars expansion is preferred in string interpolations.

Thanks again for resolving this bug! :)

andresbono (approved):

andresbono:
This is now merged. The fix will be available within the next release of the containers. Thanks again!

PR #28670, opened 2023-03-30
[bitnami/redis-cluster] fix nodes.conf not found

tomari:
### Description of the change

There is an issue in the redis-cluster docker image where the entrypoint script does not succeed when the container is restarted in a specific timing.

When the container is stopped after the entrypoint scripts are completed and before the redis server starts, nodes.sh is created but nodes.conf is missing. This prevents the entrypoint to successfully complete on next restart. The redis-cluster will fail to start forever when this happens. This change fixes this behavior by checking the existence of nodes.conf in the entrypoint scripts.

### Benefits

This change makes it more likely that the redis-cluster is successfully assembled when it is installed.

### Possible drawbacks

N/A


fmulero (approved):
Brilliant!! Thanks a lot @tomari. Greate contribution.

LGreatTM

PR #29016, opened 2023-04-03
[bitnami/rails] ARM not supported

fmulero:
### Description of the change

Remove ARM support.

### Benefits

Prevent pipeline failures. ARM is not supported in our rails container at the moment.

### Possible drawbacks

None.

### Additional information

There is an internal task to include ARM support.


andresbono (approved):
LGTM

PR #29326, opened 2023-04-05
[bitnami/gitlab-runner] Add VIB tests

corico44:
### Description of the change

The main objective of this PR is to publish our Bitnami Gitlab-Runner container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### DIFF
--- a/.vib/gitlab-runner/goss/gitlab-runner.yaml
+++ b/.vib/gitlab-runner/goss/gitlab-runner.yaml
@@ -1,6 +1,0 @@
-file:
-  /opt/bitnami/gitlab-runner/bin/gitlab-runner:
-    exists: true
-    filetype: file
-    contains:
-      - "arm64"
--- a/.vib/gitlab-runner/goss/gitlab-runner.yaml
+++ b/.vib/gitlab-runner/goss/gitlab-runner.yaml
@@ -1,0 +1,6 @@
+file:
+  /opt/bitnami/gitlab-runner/bin/gitlab-runner:
+    exists: true
+    filetype: file
+    contains:
+      - "arm64"
--- a/.vib/gitlab-runner/goss/goss.yaml
+++ b/.vib/gitlab-runner/goss/goss.yaml
@@ -1,0 +1,9 @@
+gossfile:
+  # Load scripts from .vib/common/goss/templates
+  ../../common/goss/templates/check-binaries.yaml: {}
+  ../../common/goss/templates/check-broken-symlinks.yaml: {}
+  ../../common/goss/templates/check-ca-certs.yaml: {}
+  ../../common/goss/templates/check-directories.yaml: {}
+  ../../common/goss/templates/check-files.yaml: {}
+  ../../common/goss/templates/check-linked-libraries.yaml: {}
+  ../../common/goss/templates/check-spdx.yaml: {}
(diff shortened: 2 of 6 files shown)


.vib/gitlab-runner/goss/vars.yaml
```diff
@@ -0,0 +1,11 @@
+binaries:
+  - gitlab-runner
```
joancafom, line 2: You should also check for `gosu`, as the Dockerfile includes it:

https://github.com/bitnami/containers/blob/d31477da6b9e83c8b255284531ca9db2596a1562/bitnami/gitlab-runner/15/debian-11/Dockerfile#L25

.vib/gitlab-runner/goss/vars.yaml
```diff
@@ -0,0 +1,11 @@
...
+files:
+  - paths:
+      - /opt/bitnami/gitlab-runner/bin/gitlab-runner
```
joancafom, line 5: You are implicitly checking this binary file existence in the above `binaries` section, so I don't think this adds much value.

.vib/gitlab-runner/goss/gitlab-runner.yaml
```diff
@@ -0,0 +1,6 @@
+file:
+  /opt/bitnami/gitlab-runner/bin/gitlab-runner:
+    exists: true
+    filetype: file
+    contains:
+      - "arm64"
```
joancafom, line 6: This file is a **binary**. Checking for the presence of a string that comes from the compilation stage in files of this type is not a good practice (or simply, checking strings in binary files in most cases). I'd remove this.

joancafom: Besides, the binary does not seem to include the string. The tests do pass because you forgot to include this file in the `goss.yaml` general file.

.vib/gitlab-runner/goss/vars.yaml
```diff
@@ -0,0 +1,11 @@
...
+directories:
+  - mode: "0755"
+    paths:
+      - /opt/bitnami/gitlab-runner
+      - /opt/bitnami/gitlab-runner/bin
```
joancafom, line 10: I think there is no explicit logic setting the mode for these paths, am I right?

joancafom (commented):

.vib/gitlab-runner/goss/vars.yaml
```diff
@@ -0,0 +1,11 @@
+binaries:
+  - gitlab-runner
+files:
+  - paths:
+      - /opt/bitnami/gitlab-runner/bin/gitlab-runner
+directories:
+  - mode: "0755"
+    paths:
+      - /opt/bitnami/gitlab-runner
```
FraPazGal, line 9: I think we don't really have specific logic regarding this path right? We do create the `...gitlab-runner/bin` one, but this one is not modified in any form. Even if it is the base path I don't think there is any need to check it here.

.vib/gitlab-runner/goss/goss.yaml
FraPazGal, line 1: I think you missed adding both `check-app-version` and `check-sed-in-place` goss templates, is there a reason we can't run those tests?

corico44: Added! Thank you Fran.

FraPazGal (approved):
LGTM!

joancafom (approved):

PR #29328, opened 2023-04-05
[bitnami/gitlab-runner-helper] Add VIB tests

corico44:
### Description of the change

The main objective of this PR is to publish our Bitnami Gitlab-Runner-Helper container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### DIFF
--- a/.vib/gitlab-runner-helper/goss/vars.yaml
+++ b/.vib/gitlab-runner-helper/goss/vars.yaml
@@ -5,3 +5,6 @@ files:
   - paths:
       - /opt/bitnami/gitlab-runner-helper/bin/gitlab-runner-build
 root_dir: /opt/bitnami
+version:
+  bin_name: gitlab-runner-helper
+  flag: -v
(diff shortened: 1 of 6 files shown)


.vib/gitlab-runner-helper/goss/vars.yaml
```diff
@@ -0,0 +1,10 @@
+binaries:
+  - gitlab-runner-helper
+  - gosu
+files:
+  - paths:
+      - /opt/bitnami/gitlab-runner-helper/bin/gitlab-runner-build
```
FraPazGal, line 6: I'm not sure where are we obtaining this path from; I do not see any mention of it on the asset's compilation or setup logic. Even so, given that it is a binary I think it would be better to move this check to the `binaries` section (if necessary).

FraPazGal (approved):
LGTM, thanks for the PR!

joancafom (approved):

PR #29984, opened 2023-04-11
[bitnami/rclone] Add VIB tests

migruiz4:
### Description of the change

The main objective of this PR is to publish our Bitnami Rclone container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### DIFF
--- a/.vib/rclone/goss/goss.yaml
+++ b/.vib/rclone/goss/goss.yaml
@@ -1,0 +1,9 @@
+gossfile:
+  # Load scripts from .vib/common/goss/templates
+  ../../common/goss/templates/check-app-version.yaml: {}
+  ../../common/goss/templates/check-binaries.yaml: {}
+  ../../common/goss/templates/check-broken-symlinks.yaml: {}
+  ../../common/goss/templates/check-ca-certs.yaml: {}
+  ../../common/goss/templates/check-linked-libraries.yaml: {}
+  ../../common/goss/templates/check-sed-in-place.yaml: {}
+  ../../common/goss/templates/check-spdx.yaml: {}
--- a/.vib/rclone/goss/goss.yaml
+++ b/.vib/rclone/goss/goss.yaml
@@ -1,4 +1,6 @@
 gossfile:
+  # Goss tests exclusive to the current container
+  ../../rclone/goss/rclone.yaml: {}
   # Load scripts from .vib/common/goss/templates
   ../../common/goss/templates/check-app-version.yaml: {}
   ../../common/goss/templates/check-binaries.yaml: {}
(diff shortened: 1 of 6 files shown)


.vib/rclone/goss/goss.yaml
FraPazGal, line 1: Given that rclone defines itself as "a command-line program to manage files on cloud storage", I think we could test some of its capabilities with a couple of tests, even if we are working only locally and do not have access to a cloud vendor's filesystem. To do so, we could test some basic functionalities like the following ones:

```
$ rclone sync /opt/bitnami/rclone/bin/ /tmp/RANDOM_FOLDER && ls /tmp/RANDOM_FOLDER/rclone
$ rclone ls /opt/bitnami/rclone/licenses
```

migruiz4: I added an `rclone sync` test to assert the application functionality:
```
  check-rclone-sync:
    exec: rclone sync /opt/bitnami/rclone/bin /tmp/RANDOM_FOLDER && [ -f /tmp/RANDOM_FOLDER/rclone ]
    exit-status: 0
```
Although, I didn't add the `rclone ls` test because I don't think it provides additional value in addition to the sync test.

bitnami/rclone/1/debian-11/docker-compose.yml
```diff
@@ -1,4 +1,5 @@
 version: '2'
 services:
+  # [TEST]
```
jotamartos, line 3: Is this supposed to be here?

migruiz4: It is a trigger to run the VIB pipeline. It will be removed before merging the PR

.vib/rclone/goss/rclone.yaml
```diff
@@ -0,0 +1,4 @@
+command:
+  check-rclone-sync:
+    exec: rclone sync /opt/bitnami/rclone/bin /tmp/RANDOM_FOLDER && [ -f /tmp/RANDOM_FOLDER/rclone ]
```
FraPazGal, line 3: We always try to create our tests so they are retry-able and independent from each other. Here the test would run without issues even if we repeat the same folder, but it is always good to follow the best practices. In this type of situation, we normally randomise the folder name like this:

```suggestion
    {{- $rnd_folder := printf "folder_%s" (randAlpha 5) }}
    exec: rclone sync /opt/bitnami/rclone/bin /tmp/{{ $rnd_folder }} && [ -f /tmp/{{ $rnd_folder }}/rclone ]
```

You are going to kill me, but my suggestion was more of a sketch than final code. Sorry about that 😓

migruiz4: Ah! My bad!
Let me fix it real quick

FraPazGal (approved):
LGTM!

PR #30023, opened 2023-04-12
[bitnami/trivy] Add VIB tests

migruiz4:
### Description of the change

The main objective of this PR is to publish our Bitnami Trivy container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### DIFF
--- a/.vib/trivy/goss/trivy.yaml
+++ b/.vib/trivy/goss/trivy.yaml
@@ -1,9 +1,4 @@
 command:
-  check-wire-help:
-    exec: wire --help
-    exit-status: 0
-    stderr:
-      - Subcommands
   trivy-scanner:
     exec: trivy image docker.io/bitnami/minideb:buster-snapshot-20201101T210343Z
     timeout: 60000
--- a/.vib/trivy/goss/trivy.yaml
+++ b/.vib/trivy/goss/trivy.yaml
@@ -1,9 +1,4 @@
 command:
-  check-trivy-help:
-    exec: trivy --help
-    exit-status: 0
-    stdout:
-      - Commands
   check-wire-help:
     exec: wire --help
     exit-status: 0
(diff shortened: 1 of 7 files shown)


.vib/trivy/goss/trivy.yaml
```diff
@@ -0,0 +1,17 @@
...
+  check-trivy-help:
+    exec: trivy --help
+    exit-status: 0
+    stdout:
+      - Commands
+  check-wire-help:
+    exec: wire --help
+    exit-status: 0
+    stderr:
+      - Subcommands
```
FraPazGal, line 11: I would say checking the `--help` flags does not really add enough value, given that we are already checking the binaries themselves exist.

It's true the comparison is not 1:1, but I think checking a flag and, specifically, a specific output (prone to be changed by upstream) introduces more flakiness than value. Additionally, you are already using the `trivy` binary in the `trivy-scanner` check.

I think we can remove these tests altogether or at least substitute them with ones that test the core functionalities of the asset (like the scanning of an image by trivy below). WDYT @migruiz4 @joancafom?

FraPazGal (commented):
Thanks for the PR @migruiz4! I added a comment to discuss a concern we've had in the past, please check it whenever you can.

P.S: careful with the PR description 😄

.vib/trivy/goss/trivy.yaml
migruiz4: I removed both the `trivy --help` and `wire --help` tests.

The `trivy --help` does not provide any additional value considering we are running a trivy scan below. 
The `wire --help` test was removed because we are already checking the binary is present in the `check-binaries` test.

FraPazGal (approved):
LGTM!

PR #30041, opened 2023-04-12
[bitnami/keycloak] remove dp password posting, when debug mode is enabled

robson90:
What my change does?
It removes the log output of dp-password when debug on image is set

What are the limitations?
It is not possible to see the used password in the logs anymore.

Description of the change

It removes the log output of dp-password when debug on image is set.

Benefits

It is no longer possible to see the db-password inside the log output.

Possible drawbacks

It is harder to retrieve the password for debug purposes.

Additional information

Why Im changing this?
At my current project, we want to live the GitOps approach. So I used Bitnami Sealed Secrets to seal my DB-password. Then I found the real decrypted password in the log file. So I searched why and how this happened. I found the deleted line in the shell script.

If a suspect enables the debug mode in git repository and can see the logs, the password is stolen. It is also possible that the db-password is sent to some observeability tool...

If ur an admin and really need to debug, just ssh into the container and retrieve the pw there.

I decided against masking the password.

If there are any questions, feel free to ask.

### DIFF
--- a/bitnami/keycloak/21/debian-11/rootfs/opt/bitnami/scripts/libkeycloak.sh
+++ b/bitnami/keycloak/21/debian-11/rootfs/opt/bitnami/scripts/libkeycloak.sh
@@ -135,7 +135,6 @@ keycloak_configure_database() {
     if [[ "${KEYCLOAK_DATABASE_VENDOR}" == "postgresql" ]]; then
         keycloak_conf_set "db" "postgres"
         keycloak_conf_set "db-username" "$KEYCLOAK_DATABASE_USER"
-        keycloak_conf_set "db-password" "$KEYCLOAK_DATABASE_PASSWORD"
         keycloak_conf_set "db-url" "jdbc:postgresql://${KEYCLOAK_DATABASE_HOST}:${KEYCLOAK_DATABASE_PORT}/${KEYCLOAK_DATABASE_NAME}?currentSchema=${KEYCLOAK_DATABASE_SCHEMA}${jdbc_params}"
     else
         keycloak_conf_set "db" "$KEYCLOAK_DATABASE_VENDOR"


jotamartos (changes requested):
This is not a valid PR. If you remove this line, you won't be configuring the database password and the app will fail to connect to the database when using a docker-compose file like this one 

```
version: '2'
services:
  postgresql:
    image: docker.io/bitnami/postgresql:11
    environment:
      - POSTGRESQL_USERNAME=bn_keycloak
      - POSTGRESQL_DATABASE=bitnami_keycloak
      - POSTGRESQL_PASSWORD=password123
    volumes:
      - 'postgresql_data:/bitnami/postgresql'

  keycloak:
    image: docker.io/bitnami/keycloak:21
    environment:
      - KEYCLOAK_DATABASE_PASSWORD=password123
    depends_on:
      - postgresql
    ports:
      - "80:8080"

volumes:
  postgresql_data:
    driver: local
```

If you test it, it will fail with this error message 

```
keycloak_1    | 2023-04-25 13:24:41,088 WARN  [org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator] (JPA Startup Thread: keycloak-default) HHH000342: Could not obtain connection to query metadata: org.postgresql.util.PSQLException: The server requested password-based authentication, but no password was provided by plugin null
keycloak_1    |         at org.postgresql.core.v3.AuthenticationPluginManager.lambda$withEncodedPassword$0(AuthenticationPluginManager.java:110)
keycloak_1    |         at org.postgresql.core.v3.AuthenticationPluginManager.withPassword(AuthenticationPluginManager.java:81)
```

jotamartos:
Well, the other user is modifying the same file. So will close this PR in favor of the other one. 

Thanks

PR #30098, opened 2023-04-13
[bitnami/azure-cli] Add VIB tests

migruiz4:
### Description of the change

The main objective of this PR is to publish our Bitnami azure-cli container using VMware Image Builder. In order to do that, several changes are included:

- Increasing the existing test coverage of the asset by adding Goss tests.
- Update verify and publish VIB pipeline's definitions.

### Benefits

- Ensuring higher quality of the container catalog.
- Increased pool of assets completely handled by VMware Image Builder.

### Possible drawbacks

Automated tests could introduce additional flakiness to the CI/CD.

### Applicable issues

NA

PR #30099, opened 2023-04-13
[bitnami/consul] Add VIB tests

migruiz4:
### Description of the change

The main objective of this PR is to publish our Bitnami Consul container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA


.vib/consul/goss/goss.yaml
```diff
@@ -0,0 +1,12 @@
+gossfile:
+  # Goss tests exclusive to the current container
+  ../../consul/goss/consul.yaml: {}
```
FraPazGal, line 3: Similarly to https://github.com/bitnami/containers/pull/30098, I don't see any added value having a `binary --help` test when there is already `binary --version` one. It doesn't hurt, but I'm in favour of not having it if is not valuable. In this case, we'll be able to remove the `consul.yalm` as well.

Besides that, the PR LGTM; I'll wait for the VIB trigger removal.

FraPazGal (approved):
LGTM!

PR #30395, opened 2023-04-15
[bitnami/mastodon] Fix mastodon s3

### LINKED ISSUE #30394: Incorrect variable referenced?
### Name and Version

bitnami/mastodon:4.1.2

amd64

1. enable s3
2. pass in all required s3 variables

I expect to connect to AWS S3.

### What do you see instead?

The log reveals `/opt/bitnami/scripts/libmastodon.sh: line 36: !1: unbound variable`. What I found is that there must be some required S3-related parameter that isn't being passed in, but this line https://github.com/bitnami/containers/blob/675a58fa3d49d30e1d6e5654b41921c0271c9f6b/bitnami/mastodon/4/debian-11/rootfs/opt/bitnami/scripts/libmastodon.sh#L36 is not written correctly, in that it doesn't test that the variable exists first. Failure doesn't reveal what the variable *is*, so this makes it harder to diagnose.

However, I noticed that here https://github.com/bitnami/containers/blob/675a58fa3d49d30e1d6e5654b41921c0271c9f6b/bitnami/mastodon/4/debian-11/rootfs/opt/bitnami/scripts/libmastodon.sh#L114 the variable referenced does not exist. There's *ALIAS_HOST variables, but nothing that ends in ALIAS. Aliasing is also unnecessary, but this block makes it mandatory.

### Additional information

_No response_

lavahot:
### Description of the change

Fixing problem described at #30394

### Benefits

Fixes some S3 bugs

### Possible drawbacks

Might not be a complete fix. Specifically there might be additional S3 fixes.

### Applicable issues

- fixes #30394


bitnami/mastodon/4/debian-11/rootfs/opt/bitnami/scripts/mastodon-env.sh
```diff
@@ -247,7 +248,7 @@ MASTODON_AWS_ACCESS_KEY_ID="${MASTODON_AWS_ACCESS_KEY_ID:-"${AWS_ACCESS_KEY_ID:-
 export MASTODON_AWS_ACCESS_KEY_ID="${MASTODON_AWS_ACCESS_KEY_ID:-}"
 export AWS_ACCESS_KEY_ID="$MASTODON_AWS_ACCESS_KEY_ID"
 export MASTODON_CFG_AWS_ACCESS_KEY_ID="$MASTODON_AWS_ACCESS_KEY_ID"
-export MASTODON_S3_REGION="${MASTODON_S3_REGION:-us-east-1}"
+export MASTODON_S3_REGION="${MASTODON_S3_REGION:-"${S3_REGION:-us-east-1}"
```
dgomezleon, line 251: There is a typo with a missing `}"` in this line. Could you please fix it?

dgomezleon (changes requested):
Thanks for contributing. There is a minor typo that needs to be fixed.

bitnami/mastodon/4/debian-11/rootfs/opt/bitnami/scripts/libmastodon.sh
```diff
@@ -33,7 +33,7 @@ mastodon_validate() {
         error_code=1
     }
     check_empty_value() {
-        if is_empty_value "${!1}"; then
+        if [[ -v "$1" ]] && is_empty_value "${!1}"; then
```
dgomezleon, line 36: Why did you add it? I don't consider it necessary.

dgomezleon (changes requested):

bitnami/mastodon/4/debian-11/rootfs/opt/bitnami/scripts/libmastodon.sh
lavahot: So this is to prevent the case that I ran into where the variable wasn't just unset, but was undefined. The line following would be unreachable and it would be very difficult to determine which variable was being tested. Maybe `check_empty_value` is the wrong place for this semantically, but it should be checked somewhere.

I just realized that this check is actually wrong. If the variable is undefined, it will do nothing. It should be:
```suggestion
        if [[ ! -v "$1" ]] || is_empty_value "${!1}"; then
```
This will shortcut to the error we want on the next line if the variable is unset. And if it is set, but empty, it will also go to the error on the next line. And if it's set and not empty nothing will happen.

bitnami/mastodon/4/debian-11/rootfs/opt/bitnami/scripts/mastodon-env.sh
lavahot: Sorry about that. I thought I had that right.

```suggestion
export MASTODON_S3_REGION="${MASTODON_S3_REGION:-"${S3_REGION:-us-east-1}"}"
```

bitnami/mastodon/4/debian-11/rootfs/opt/bitnami/scripts/libmastodon.sh
dgomezleon: `[[ ! -v "$1" ]]` will be always true. Take into account that all the variables are set (at least to an empty value) in [mastodon-env.sh](https://github.com/bitnami/containers/pull/30395/files#diff-4327d3a68798929b5596005ef5ccb2a1de317d317ba61d4c8bfad83637552375). So it is only necessary to check if empty with `-z`. That's what `is_empty_value` does.

lavahot: That's my point. As it stands, it's not true *now*. It's a sanity check because someone made a typo that is fixed in this PR. It prevents someone from making a mistake that's hard to recover from in the future.

lavahot: Specifically I'm referring to the `MASTODON-S3-ALIAS` unset variable on line 114. Someone wrote the name wrong and it failed poorly because the error only states that the variable is unset, not what the variable is. To debug this I had to go to source and find a variable that didn't make sense. If I had this test I would have known exactly what I was looking for immediately.

dgomezleon: I see your point. However, this would imply changing a lot of files since `check_empty_value` is used many times for all the solutions. Also, as the name indicates, it is supposed to check if a variable is empty. Another issue is that `-v` flag is not recognized by BASH <= 3, and could produce errors.

Due to this, we prefer not to add this validation and show the error in case of a typo.

lavahot: Thats fair enough, I guess. Don't need to upend the world. I'll remove it.

bitnami/mastodon/4/debian-11/rootfs/opt/bitnami/scripts/libmastodon.sh
```diff
@@ -33,7 +33,7 @@ mastodon_validate() {
         error_code=1
     }
     check_empty_value() {
-        if is_empty_value "${!1}"; then
+        if [[ ! -v "$1" ]] || is_empty_value "${!1}"; then
```
lavahot, line 36:
```suggestion
        if is_empty_value "${!1}"; then
```

dgomezleon (approved):
LGTM. Thanks for your contribution :)

PR #30808, opened 2023-04-18
[bitnami/minio] Fix the image startup problem

PuppetA17:
Image startup scans all the first-level directories of bucket through mc ls to determine whether bucket exists or skips. In a normal scenario, this is not a problem, but when there are a large number of files in the first-level directory of bucket, it will become a disaster. In my scene, there are millions or even tens of millions of files in a first-level directory under bucket. When the image is restarted, it will take up a lot of disk io. If the underlying use is file storage such as nas, this will be more intolerable

### Applicable issues

- fixes #


CeliaGMqrz (approved):
Thanks for your contribution! LGTM!

PR #30858, opened 2023-04-19
[bitnami/discourse] Add VIB tests

migruiz4:
### Description of the change

The main objective of this PR is to publish our Bitnami Discourse container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA


.vib/discourse/goss/discourse.yaml
```diff
@@ -0,0 +1,17 @@
+command:
```
FraPazGal, line 1: I would add several checks related to the logic executed at compilation time:

- Check the global npm package `terser` installed
- Check Passenger native support works is properly installed

.vib/discourse/goss/vars.yaml
```diff
@@ -0,0 +1,27 @@
+binaries:
+  - wait-for-port
+  - python
+  - ruby
+  - node
+  - brotli
+  - psql
```
FraPazGal, line 7: I would add bundle, which is used as a discourse cli.

FraPazGal (commented):
Thanks PR @migruiz4! I added a few comments, please review them whenever you can.

.vib/discourse/goss/discourse.yaml
migruiz4: I added the `terser` node module test:
```
  check-terser-modules:
    exec: npm list -g
    exit-status: 0
    stdout:
      - terser
```

Is it okay or would it be enough if we simply add it to the list of checked binaries (as it should be in the PATH)? Maybe both?

FraPazGal: I think the node module test is great 👍

FraPazGal (approved):

PR #30864, opened 2023-04-19
[bitnami/chartmuseum] Add VIB tests

corico44:
### Description of the change

The main objective of this PR is to publish our Bitnami Chartmuseum container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### Additional information

The files dedicated to permissions for Internal TLS for CentOS are not checked since they are used for a specific distro and therefore, it is not a common behavior for all distros:
```
# Fix for CentOS Internal TLS
if [[ -f /etc/pki/tls/certs/ca-bundle.crt ]]; then
    chmod g+w /etc/pki/tls/certs/ca-bundle.crt
fi

if [[ -f /etc/pki/tls/certs/ca-bundle.trust.crt ]]; then
    chmod g+w /etc/pki/tls/certs/ca-bundle.trust.crt
fi
```
Code from postunpack.sh file.

### DIFF
--- a/.vib/chartmuseum/goss/chartmuseum.yaml
+++ b/.vib/chartmuseum/goss/chartmuseum.yaml
@@ -4,11 +4,6 @@ group:
 user:
   harbor:
     exists: true
-file:
-  /bitnami/data:
-    exists: true
-    mode: "0775"
-    filetype: directory
 command:
   # Ensure a set of directories exist and the non-root user has write privileges to them
   check-directories-exist-with-user:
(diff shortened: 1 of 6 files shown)


.vib/chartmuseum/goss/chartmuseum.yaml
```diff
@@ -0,0 +1,16 @@
...
+  harbor:
+    exists: true
+user:
+  harbor:
+    exists: true
+file:
+  /bitnami/data:
+    exists: true
+    mode: "0775"
+    filetype: directory
```
FraPazGal, line 11: If there is now need to check the owner/group of the path we should use the goss templates instead of making this file bigger.

FraPazGal (dismissed):
LGTM!

FraPazGal (approved):

PR #31115, opened 2023-04-20
[bitnami/fluentd] Add VIB tests

corico44:
### Description of the change

The main objective of this PR is to publish our Bitnami Fluentd container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### DIFF
--- a/.vib/fluentd/goss/fluentd.yaml
+++ b/.vib/fluentd/goss/fluentd.yaml
@@ -6,5 +6,9 @@ command:
     exec: ls -l /opt/bitnami/fluentd/gems/fluent-plugin-detect-exceptions-*/lib/fluent/plugin/*.rb 2>/dev/null | grep ".*-rwxrwxr-x"
     exit-status: 0
   check-plugins:
-    exec: ls /opt/bitnami/fluentd/gems/fluent-plugin-concat-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-detect-exceptions-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-elasticsearch-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-grafana-loki-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-kafka-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-kubernetes_metadata_filter-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-multi-format-parser-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-prometheus-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-rewrite-tag-filter-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-s3-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-systemd-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-record-modifier-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/oj-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/json-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluentd-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/activesupport-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/elasticsearch-xpack-* 2>/dev/null
+    exec: gem list --local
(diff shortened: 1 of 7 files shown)


.vib/fluentd/goss/fluentd.yaml
```diff
@@ -0,0 +1,10 @@
+command:
+  check-no-pem-files:
+    exec: find /opt/bitnami/fluentd -name "*.pem" | grep .
+    exit-status: 1
+  check-fluent-plugin-detect-exceptions:
+    exec: ls -l /opt/bitnami/fluentd/gems/fluent-plugin-detect-exceptions-*/lib/fluent/plugin/*.rb 2>/dev/null | grep ".*-rwxrwxr-x"
+    exit-status: 0
```
joancafom, line 7: This test might end up producing false positives: if there are multiple files but only one of them matches the permissions, it will still succeed.

.vib/fluentd/goss/fluentd.yaml
```diff
@@ -0,0 +1,10 @@
...
+  check-plugins:
+    exec: ls /opt/bitnami/fluentd/gems/fluent-plugin-concat-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-detect-exceptions-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-elasticsearch-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-grafana-loki-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-kafka-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-kubernetes_metadata_filter-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-multi-format-parser-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-prometheus-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-rewrite-tag-filter-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-s3-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-systemd-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluent-plugin-record-modifier-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/oj-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/json-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/fluentd-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/activesupport-* 2>/dev/null && ls /opt/bitnami/fluentd/gems/elasticsearch-xpack-* 2>/dev/null
```
joancafom, line 9: This is valid but not an elegant approach IMHO.

1. Instead of a plain `ls`, you should check if the CLI does offer a similar functionality and prefer it. In this case, it seems that `fluent-gem` does so.
2. When you have a list or an iterable collection for which you want to perform the same check, you should use `go-template` loops to improve readability
3. Again, if you have a fixed list or collection related to test data, the right place to declare it is in `vars.yaml`. You can then reference it here, which is much cleaner 😁

joancafom (commented):

.vib/fluentd/goss/fluentd.yaml
corico44: How would you recommend transforming the test so that these false positives don't occur?

joancafom: Maybe a good approach would be checking the opposite case: if one of the files does not match, then produce an error. But it just an idea :)

.vib/fluentd/goss/vars.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - fluent-plugin-prometheus
+    - fluent-plugin-rewrite-tag-filter
+    - fluent-plugin-s3
+    - fluent-plugin-systemd
+    - fluent-plugin-record-modifier
+    - oj
+    - json
+    - fluentd
+    - activesupport
+    - elasticsearch-xpack
```
joancafom, line 32: This is okay, but I would select a subset of the most relevant ones. Again, this is subjective, so feel free to disregard my comment if you believe all of them are relevant.

bitnami/fluentd/1/debian-11/docker-compose.yml
```diff
@@ -2,6 +2,7 @@ version: '2'
 
 services:
   fluentd:
+    # New change
```
joancafom, line 5: Remove the trigger and I'll be happy to approve the PR :)

.vib/fluentd/goss/vars.yaml
corico44: I believe that all packages should be reviewed even if there are several to review.

joancafom (approved):

PR #31174, opened 2023-04-21
[bitnami/gitea] Add VIB tests

corico44:
### Description of the change

The main objective of this PR is to publish our Bitnami Gate container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### DIFF
--- a/.vib/gitea/goss/gitea.yaml
+++ b/.vib/gitea/goss/gitea.yaml
@@ -56,4 +56,4 @@ command:
     exec: timeout --preserve-status 5 gitea web
     exit-status: 0
     stdout:
-      - "/Starting.*web.*server/"
+      - "/Starting.*Web.*server/"
--- a/.vib/gitea/goss/gitea.yaml
+++ b/.vib/gitea/goss/gitea.yaml
@@ -53,7 +53,7 @@ file:
     filetype: file
 command:
   check-app-run:
-    exec: gitea web
+    exec: timeout --preserve-status 5 gitea web
     exit-status: 0
     stdout:
-      - "Starting new Web server"
+      - "Starting new"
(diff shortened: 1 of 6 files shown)


.vib/gitea/goss/gitea.yaml
```diff
@@ -0,0 +1,65 @@
...
+    mode: "0775"
+    owner: gitea
+    group: root
+    filetype: directory
+  /opt/bitnami/gitea/custom/conf/app.ini.template:
+    exists: true
+    mode: "0664"
+    owner: gitea
+    group: root
+    filetype: file
```
FraPazGal, line 59: Given we are checking the `app.ini` file below, I think we can remove this template check.

.vib/gitea/goss/gitea.yaml
FraPazGal, line 1: I would include a check testing Gitea as an asset. The go-to test I would say that is spinning up the Gitea web server and checking its stdout.

FraPazGal (commented):
Thanks for the PR! I left some comments as a review.

.vib/gitea/goss/gitea.yaml
```diff
@@ -0,0 +1,59 @@
...
+command:
+  check-app-run:
+    exec: timeout --preserve-status 5 gitea web
+    exit-status: 0
+    stdout:
+      - "Starting new"
```
FraPazGal, line 59: Checking the commands stdout, I think we may need to use a more specific string to check. Maybe something like `/Starting.*web.*server/` could work here, wdyt?

FraPazGal (approved):

PR #31490, opened 2023-04-24
[bitnami/jenkins] Add VIB tests

corico44:
### Description of the change

The main objective of this PR is to publish our Bitnami Jenkins container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### Additional information

The `jenkins-plugin-manager.jar` file is not checked with the command provided in the build recipe (`java -jar /opt/bitnami/jenkins/jenkins-plugin-manager.jar --war /opt/bitnami/jenkins/jenkins. war --plugin-file /opt/bitnami/jenkins/plugins.txt -d /opt/bitnami/jenkins/plugins`) as it requires the `java -jar /opt/bitnami/jenkins/jenkins.war` command to be executed first and we should force it to stop so that we can then run the `jenkins-plugin-manager.jar` command. As we only want to test that the file works, with the `java -jar /opt/bitnami/jenkins/jenkins-plugin-manager.jar` command we can already verify that it works.

### DIFF
--- a/.vib/jenkins/goss/jenkins.yaml
+++ b/.vib/jenkins/goss/jenkins.yaml
@@ -2,11 +2,6 @@ command:
   check-app-jar:
     exec: java -jar /opt/bitnami/jenkins/jenkins-plugin-manager.jar
     exit-status: 0
-  check-app-version:
-    exec: java -jar /opt/bitnami/jenkins/jenkins.war --version
-    exit-status: 0
-    stdout:              
-      - {{ .Env.APP_VERSION }}
 file:
   /opt/bitnami/jenkins/plugins.txt:
     exists: true
--- a/.vib/jenkins/goss/jenkins.yaml
+++ b/.vib/jenkins/goss/jenkins.yaml
@@ -2,10 +2,3 @@ command:
   check-app-jar:
     exec: java -jar /opt/bitnami/jenkins/jenkins-plugin-manager.jar
     exit-status: 0
-file:
-  /opt/bitnami/jenkins/plugins.txt:
-    exists: true
-    contains:
-      - "swarm"
-      - "matrix-auth"
-      - "authorize-project"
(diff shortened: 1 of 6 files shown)


.vib/jenkins/goss/jenkins.yaml
```diff
@@ -0,0 +1,16 @@
...
+    exit-status: 0
+    stdout:              
+      - {{ .Env.APP_VERSION }}
+file:
+  /opt/bitnami/jenkins/plugins.txt:
+    exists: true
+    contains:
+      - "swarm"
+      - "matrix-auth"
+      - "authorize-project"
```
joancafom, line 16: This is where the plugins to download are defined, but I would rather check they are actually present in the plugins folder

.vib/jenkins/goss/jenkins.yaml
```diff
@@ -0,0 +1,16 @@
+command:
+  check-app-jar:
+    exec: java -jar /opt/bitnami/jenkins/jenkins-plugin-manager.jar
+    exit-status: 0
+  check-app-version:
+    exec: java -jar /opt/bitnami/jenkins/jenkins.war --version
+    exit-status: 0
+    stdout:              
+      - {{ .Env.APP_VERSION }}
```
joancafom, line 9: I think you can safely remove this and use the `check-app-version.yaml` test instead, with `java -jar /opt/bitnami/jenkins/jenkins.war` as the command

.vib/jenkins/goss/jenkins.yaml
joancafom: Alternatively, you can also use the `jenkins-plugin-manager.jar` jar file to assert this

joancafom (commented):
Thanks for your PR! I left you some comments as a review 🚀

joancafom (approved):

PR #31514, opened 2023-04-24
[bitnami/keycloak] Add VIB tests

corico44:
### Description of the change

The main objective of this PR is to publish our Bitnami Keycloak container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### Additional information

The part of code in the build recipe where execution permissions are given to different libraries isn't checked as it depends on the specific architecture being used:
```
    // Some libraries in Keycloak don't have execution permissions and ldd warns about that
    const arch = this.be.target.platform.arch === 'arm64' ? 'aarch64' : 'x86_64';
    const paths = [
      `modules/system/layers/base/org/apache/activemq/artemis/journal/main/lib/linux-${arch}/*`,
      'modules/system/layers/base/org/wildfly/openssl/main/lib/*/*.so',
    ];
    paths.forEach((p) => {
      const libraries = glob.sync(path.join(
        this.prefix,
        p
      ));
      this.logger.debug('Granting execution permisions in the following libraries:', libraries.join(', '));
      libraries.forEach((lib) => fs.chmodSync(lib, 0o754));
    });
  }
```

### DIFF
--- a/.vib/keycloak/goss/keycloak.yaml
+++ b/.vib/keycloak/goss/keycloak.yaml
@@ -61,9 +61,3 @@ file:
     exists: true
     contains:
       - "<module xmlns=\"urn:jboss:module:1.0\" name=\"org.postgresql.jdbc\">"
-command:
-  check-app-version:
-    exec: /opt/bitnami/keycloak/bin/kc.sh --version
-    exit-status: 0          
-    stdout:              
-      - {{ .Env.APP_VERSION }}
--- a/.vib/keycloak/goss/keycloak.yaml
+++ b/.vib/keycloak/goss/keycloak.yaml
@@ -60,11 +60,7 @@ file:
   /opt/bitnami/keycloak/modules/system/layers/keycloak/org/postgresql/jdbc/main/module.xml:
     exists: true
     contains:
-      - "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
       - "<module xmlns=\"urn:jboss:module:1.0\" name=\"org.postgresql.jdbc\">"
-      - "<resource-root path=\"postgresql-"
-      - "<module name=\"javax.api\"/>"
-      - "<module name=\"javax.transaction.api\"/>"
 command:
   check-app-version:
     exec: /opt/bitnami/keycloak/bin/kc.sh --version
--- a/.vib/keycloak/goss/keycloak.yaml
+++ b/.vib/keycloak/goss/keycloak.yaml
@@ -67,7 +67,7 @@ file:
       - "<module name=\"javax.transaction.api\"/>"
 command:
   check-app-version:
-    exec: /opt/bitnami/keycloak/bin/kc.sh --version | grep "Keycloak" | grep -m 1 -o "[0-9]*\.[0-9]*\.[0-9]"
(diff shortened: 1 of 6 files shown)


.vib/keycloak/goss/vars.yaml
```diff
@@ -0,0 +1,7 @@
+binaries:
+  - java
+  - wait-for-port
```
FraPazGal, line 3: We could add `kc.sh` here as well.

.vib/keycloak/goss/vars.yaml
```diff
@@ -0,0 +1,7 @@
...
+directories:
+  - paths:
+      - /opt/bitnami/keycloak/modules/system/layers/keycloak/org/postgresql/jdbc/main
```
FraPazGal, line 6: Given that you are already checking the `module.xml` file inside this dir, I think we can remove this check.

.vib/keycloak/goss/keycloak.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - "<module xmlns=\"urn:jboss:module:1.0\" name=\"org.postgresql.jdbc\">"
+      - "<resource-root path=\"postgresql-"
+      - "<module name=\"javax.api\"/>"
+      - "<module name=\"javax.transaction.api\"/>"
+command:
+  check-app-version:
+    exec: /opt/bitnami/keycloak/bin/kc.sh --version
+    exit-status: 0          
+    stdout:              
+      - {{ .Env.APP_VERSION }}
```
FraPazGal, line 73: I think we can use the `check-app-version` template using:

```
version:
  bin_name: kc.sh
  flag: --version
```

.vib/keycloak/goss/keycloak.yaml
```diff
@@ -0,0 +1,73 @@
...
+    owner: keycloak
+    filetype: directory
+  /opt/bitnami/keycloak/modules/system/layers/keycloak/org/postgresql/jdbc/main/module.xml:
+    exists: true
+    contains:
+      - "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
+      - "<module xmlns=\"urn:jboss:module:1.0\" name=\"org.postgresql.jdbc\">"
+      - "<resource-root path=\"postgresql-"
+      - "<module name=\"javax.api\"/>"
+      - "<module name=\"javax.transaction.api\"/>"
```
FraPazGal, line 67: As we are adding the whole block at the same time after the compilation, we can reduce the `contains` check to just one of these strings, probably one containing a psql mention.

FraPazGal (commented):
Nice PR @corico44, I have a added some comments in my review. Please check them whenever possible.

.vib/keycloak/goss/keycloak.yaml
FraPazGal: I think you forgot removing this code 😅 Otherwise, the PR LGTM!

FraPazGal (commented):
The changes LGTM @corico44, please ping me once the VIB trigger is removed

agomezmoron (dismissed):

FraPazGal (approved):

PR #31574, opened 2023-04-25
[bitnami/keycloak] Redact db password from debug log

### LINKED ISSUE #31067: [bitnami/keycloak] Hide db password in debug log
### Name and Version

bitnami/keycloak:21

amd64

1. Set environment variable `BITNAMI_DEBUG` to "true"
2. Look at Keycloak logs

Should not output the db password at all

### What do you see instead?

Keycloak logs db password: `Setting db-password to '...`

### Additional information

This line is to blame I think: https://github.com/bitnami/containers/blob/01d5e9b62dceaed6e886da469bcaa9b926ea692d/bitnami/keycloak/21/debian-11/rootfs/opt/bitnami/scripts/libkeycloak.sh#L138

pckhoi:
### Description of the change

Display `_redacted_` in place of db password in debug log.

### Benefits

- Avoid unintentionally leaking db password to logs.

### Possible drawbacks

I don't think there are any.

### Applicable issues

- fixes #31067

### DIFF
--- a/bitnami/keycloak/21/debian-11/rootfs/opt/bitnami/scripts/libkeycloak.sh
+++ b/bitnami/keycloak/21/debian-11/rootfs/opt/bitnami/scripts/libkeycloak.sh
@@ -106,7 +106,7 @@ keycloak_conf_set() {
     local -r value="${2:-}"
     # Redact sensitive values before outputting to debug log
     local redacted_value="${value}"
-    if [[ "${key}" == "db-password" ]]; then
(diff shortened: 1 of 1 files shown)


fmulero (commented):
Great @pckhoi!! good catch! Thanks a lot for your contribution.

Your changes looks good to me but I have found more values that can be redacted, for example:
* `https-key-store-password`
* `https-trust-store-password`
* `spi-truststore-file-password`

Could you take them into account in your change?

pckhoi:
@fmulero I have add those keys and I believe they are all that need to be redacted?

fmulero (approved):
Brilliant!! That's enough from my side. Thank you so much for your contribution.

PR #31812, opened 2023-04-26
Include Nashorn js engine in keycloak image

vruum:
### Description of the change
After upgrading the image to use java 17 (which no longer includes the nashorn JS engine), users using authentication providers that depends on a js engine being present, found that their providers no longer worked.

This commit Includes a separate buildstep in the keycloak dockerfile that basically runs a maven build that depends on nashorn and copies the resulting dependencies to the keycloak image. 
The reason for using a maven build is basically to make it easier to keep track of, and include, any transient dependencies (mostly just asm stuff)

### Benefits

Users can use the image to run keycloak providers without having to bother with copying the nashorn jar and dependencies into the container

### Applicable issues

https://github.com/bitnami/charts/issues/15659

Signed-off-by: Peter Holtegaard Nielsen <peterholtegaard@gmail.com>


carrodher:
Thanks for your contribution! Unfortunately, it is difficult for us to accept this PR since the _Dockerfile_ structure is generic across Bitnami containers.
In this case, our suggestion is to include those changes in your local repo and build the container image from there with all the required changes to work for your use case, but not adding those changes by default for all the users

PR #31971, opened 2023-04-28
[bitnami/jasperreports] Only enable remote management on default webapps if default webapps are installed (#31969).

### LINKED ISSUE #31969: Tomcat container fails to startup with TOMCAT_INSTALL_DEFAULT_WEBAPPS="no"
### Name and Version

bitnami/tomcat

amd64

1. Run `docker run -e TOMCAT_INSTALL_DEFAULT_WEBAPPS="no" -e TOMCAT_ENABLE_AUTH="no" -e TOMCAT_ALLOW_REMOTE_MANAGEMENT="no" bitnami/tomcat:9.0`
2. It will crash after the log line "Skipping deployment of default webapps"

The log line `** tomcat setup finished! **` should show and after that it should start up Tomcat

### What do you see instead?

```bash
$ docker run -e TOMCAT_INSTALL_DEFAULT_WEBAPPS="no" -e TOMCAT_ENABLE_AUTH="no" -e TOMCAT_ALLOW_REMOTE_MANAGEMENT="no" bitnami/tomcat:9.0
tomcat 12:02:16.99
tomcat 12:02:17.00 Welcome to the Bitnami tomcat container
tomcat 12:02:17.00 Subscribe to project updates by watching https://github.com/bitnami/containers
tomcat 12:02:17.00 Submit issues and feature requests at https://github.com/bitnami/containers/issues
tomcat 12:02:17.00
tomcat 12:02:17.00 INFO  ==> ** Starting tomcat setup **
tomcat 12:02:17.02 INFO  ==> Configuring port numbers
tomcat 12:02:17.03 INFO  ==> Ensuring Tomcat directories exist
tomcat 12:02:17.03 INFO  ==> Skipping deployment of default webapps
```

The exit-code (0 expected):
```
$ echo $?
1
```

### Additional information

[This commit](https://github.com/bitnami/containers/commit/64fc796b989ad7e2f7b445b303c18d6a67e99196) introduced the issue.

In `libtomcat.sh` (called from entrypoint.sh in this case) it used to first check the `TOMCAT_ALLOW_REMOTE_MANAGEMENT` flag before it would attempt to enable the remote connections for each applications. Now it will always do that for `examples` and `docs`.
In case the user (me in this case) disables the installation of default web apps (`TOMCAT_INSTALL_DEFAULT_WEBAPPS="no"`, [this line](https://github.com/bitnami/containers/commit/64fc796b989ad7e2f7b445b303c18d6a67e99196#diff-b0aa04da63101c75c14f517ef9df1ee703eb249bfa5497562ce3c451f1ecb365R149) will return and quits the entrypoint.sh script with an error code.

MartinKanters:
### Description of the change

This change ensures that remote management will not be enabled on the default webapps (examples, docs, manager and host-manager) if `TOMCAT_INSTALL_DEFAULT_WEBAPPS` is disabled. It [currently fails to start up the container](https://github.com/bitnami/containers/issues/31969) if that's the case. 

### Benefits

It fixes the bug.

### Possible drawbacks

Theoretically, enabling remote management was possible on the default webapps, even if TOMCAT_INSTALL_DEFAULT_WEBAPPS was false. I don't think it's sensible to attempt this, but I might be wrong. That is why I moved the code inside the `TOMCAT_INSTALL_DEFAULT_WEBAPPS==true` if condition.
There are more possibilities to fix this, obviously, but I thought this was the cleanest.

### Applicable issues

- fixes #31969

### Additional information

I verified the tomcat:9 image by building and running it locally. (Related PR for Tomcat: https://github.com/bitnami/containers/pull/31970)

### DIFF
--- a/bitnami/jasperreports/8/debian-11/rootfs/opt/bitnami/scripts/libtomcat.sh
+++ b/bitnami/jasperreports/8/debian-11/rootfs/opt/bitnami/scripts/libtomcat.sh
@@ -201,18 +201,18 @@ EOF
         if is_boolean_yes "$TOMCAT_INSTALL_DEFAULT_WEBAPPS"; then
             info "Deploying Tomcat from scratch"
             cp -rp "$TOMCAT_BASE_DIR"/webapps_default/* "$TOMCAT_WEBAPPS_DIR"
+
+            # These applications have been enabled for historical reasons, and do not pose any security threat
+            tomcat_enable_application examples
+            tomcat_enable_application docs
+            if is_boolean_yes "$TOMCAT_ALLOW_REMOTE_MANAGEMENT"; then
+                # These applications should not be enabled by default, for security reasons
+                info "Enabling remote connections for manager and host-manager applications"
+                tomcat_enable_application manager
+                tomcat_enable_application host-manager
+            fi
         else
             info "Skipping deployment of default webapps"
-        fi
-
-        # These applications have been enabled for historical reasons, and do not pose any security threat
-        tomcat_enable_application examples
-        tomcat_enable_application docs
-        if is_boolean_yes "$TOMCAT_ALLOW_REMOTE_MANAGEMENT"; then
-            # These applications should not be enabled by default, for security reasons
-            info "Enabling remote connections for manager and host-manager applications"
-            tomcat_enable_application manager
-            tomcat_enable_application host-manager
         fi
     fi
 }


aoterolorenzo:
Hi @MartinKanters,

What would happen when TOMCAT_INSTALL_DEFAULT_WEBAPPS is set to false, and one of more non-default webapps are added to the system. This does not disable the manager for them too?

Just trying to look for an edge case to ensure this need

MartinKanters:
> Hi @MartinKanters,
> 
> What would happen when TOMCAT_INSTALL_DEFAULT_WEBAPPS is set to false, and one of more non-default webapps are added to the system. This does not disable the manager for them too?
> 
> Just trying to look for an edge case to ensure this need

Hi, I understand that example, docs, manager and host-manager are the default-webapps that a user either wants to install or not (based on the TOMCAT_INSTALL_DEFAULT_WEBAPPS flag). I don't completely follow what you mean with:
> This does not disable the manager for them too?
The manager is a separate webapp just like any (non-default) webapp that a user can install on tomcat.

If you are concerned about whether users can still allow remote management on their custom webapps if they set TOMCAT_INSTALL_DEFAULT_WEBAPPS to false. Yes, I think they can, because currently this flag (TOMCAT_ALLOW_REMOTE_MANAGEMENT) only activates remote management on the default-webapps.

aoterolorenzo (approved):
@MartinKanters thank you for your contribution!

PR #32281, opened 2023-05-01
[bitnami/airflow] Merging layers into one layer

alekseyolg:
Merging layers will reduce the final size of the image, and reducing the output to stdout will improve readability when assembling the image

### Applicable issues

- fixes #


aoterolorenzo:
Hi @alekseyolg,

Actually, our _published_ docker images are already single-layered, since we apply a `--squash` at the build step before publishing. So although here you can see here a layered Dockerfile, the _published_ result has one only.

image

In case you need to reproduce this building from the Dockerfile, you can also try the [--squash](https://docs.docker.com/engine/reference/commandline/build/#squash) feature by yourself

PR #32665, opened 2023-05-04
[bitnami/mongodb] Create replicaset keyFile before mongo starts

### LINKED ISSUE #16386: [bitnami/mongodb] Chart cannot deploy if Value 'configuration' file is provided with a 'security.keyFile'
### Name and Version

bitnami/mongodb 13.9.4

None

1. Deploy the chart with a MongoDB configuration file (see values section).

```yaml
storage:
  dbPath: /bitnami/mongodb/data/db
    journal:
      enabled: true
    directoryPerDB: false
systemLog:
  destination: file
  quiet: false
  logAppend: true
  logRotate: reopen
  path: /opt/bitnami/mongodb/logs/mongodb.log
  verbosity: 0
net:
  port: 27017
  unixDomainSocket:
    enabled: true
    pathPrefix: /opt/bitnami/mongodb/tmp
  ipv6: false
  bindIpAll: true
  tls:
    allowConnectionsWithoutCertificates: true
replication:
  replSetName: rs0
  enableMajorityReadConcern: true
processManagement:
  fork: false
  pidFilePath: /opt/bitnami/mongodb/tmp/mongodb.pid
setParameter:
  enableLocalhostAuthBypass: false
security:
  authorization: enabled
  keyFile: /opt/bitnami/mongodb/conf/keyfile
```

It seems that the ReplicaSet keyfile is created after MongoDB is started, but it's not clear why that happens to me. I believe the ReplicaSet keyfile should be created early in the initialization process.

### What do you see instead?

The pod would not start and instead threw the error below. Apologies, I had to write this error out by hand (it is not a copy+paste), so there may be a typo.
```
about to fork child process, waiting until server is ready for connections.
forked process: 78
ERROR: child process failed, exited with 1
to see additional information in this output, start without the "--fork" option.
```

### Additional information

Potentially related to https://github.com/bitnami/charts/issues/16385, my goal was to set the MongoDB setting of `net.tls.allowConnectionsWithoutCertificates`. In an attempt to do this:
1. I coped the contents of `/opt/bitnami/mongodb/conf/mongodb.conf` from within the container
2. I added these contents to the Chart `values.yaml`, under the `configuration` field.

In trying to get this working, I deleted the namespace to test a fresh deployment with `values.yaml` that were working.

To troubleshoot this (long story short) I put a `sleep 1000000` in the `setup.sh` before launching Mongo, entered the pod, and manually ran `mongod` until I found it complaining that the file `/opt/bitnami/mongodb/conf/keyfile` does not exist. This file was set in the `values.yaml` under the `configuration` field, in the MongoDB configuration file: `security.keyFile`.

### LINKED ISSUE #32663: [bitnami/mongodb] Replicaset keyfile is made after Mongo starts
### Name and Version

bitnami/mongodb 13.9.4

amd64

Deploy the [Bitnami MongoDB helm chart](https://github.com/bitnami/charts/) with a `.values.configuration` set containing a pointer in the `security.keyFile` field to the `/opt/bitnami/mongodb/conf/keyfile` file. 

Mongo deploys successfully

### What do you see instead?

Pod fails because `keyFile` does not exist when mongo is started on first run.

### Additional information

Related Helm Chart issue: https://github.com/bitnami/charts/issues/16385.

I plan on making a PR to address this.

rrileyca:
### Description of the change

This change creates the MongoDB Replicaset keyFile before `mongod` starts. 

### Benefits

This is needed because if an externally mounted file references the default keyFile that is made at path `/opt/bitnami/mongodb/conf/keyfile`, MongoDB will fail to start due to the non-existant file.

### Possible drawbacks

To my understanding, none.

### Applicable issues

- fixes bitnami/charts#16386
- fixes bitnami/containers#32663


mdhont (changes requested):
These changes need to be tested and reviewed internally, in any case could you fix the indentation?

rrileyca:
@mdhont done

dtrts:
Hello, 

I have been working with this container recently and have applied a work around which also creates and configures the keyfile before `mongod` is started. This PR would definitely help. 

I noticed that you have moved `mongodb_set_listen_all_conf` to before the initial startup. I was thinking this has a (miniscule) risk of letting `mongod` accept external commands before the initial users had been created. 
The option here is to leave it where it is so that the first time `mongod` is run it is bound only to local host.


----

(In my scenario i am passing `--transitionToAuth` in as an extra flag which requires authorisation to be enabled. To stop mongod failing on the first run through I enable auth through `mongodb_set_keyfile_conf`. Since `enableLocalhostAuthBypass` is still enabled for the initilization the root user is still created despite auth being required at that point)

rrileyca:
Good point @dtrts. I've made the suggested change, and left the listen-all configuration where it was.

rrileyca:
Bump

rrileyca:
@mdhont is the review underway or scheduled?

mdhont:
The solution is failing in our internal tests, and it needs some further investigation to find the cause.

rrileyca:
@mdhont thank you for the update.

mdhont (approved):
Lgtm!

mdhont:
@rrileyca, the license-header-linter job is failing, which is checking if some files contain a needed header. As this has recently been added could you sync your fork and do a rebase on the main branch?

rrileyca:
@mdhont Done! Thanks for reviewing.

PR #29156, opened 2023-04-04
[bitnami/drupal] Add VIB tests

migruiz4:
### Description of the change

The main objective of this PR is to publish our Bitnami Drupal container using VMware Image Builder. In order to do that, several changes are included:

- Increasing the existing test coverage of the asset by adding Goss tests.
- Update verify and publish VIB pipeline's definitions.

### Benefits

- Ensuring higher quality of the container catalog.
- Increased pool of assets completely handled by VMware Image Builder.

### Possible drawbacks

Automated tests could introduce additional flakiness to the CI/CD.

### Applicable issues

NA

PR #32689, opened 2023-05-05
[bitnami/magento] Add VIB tests

corico44:
### Description of the change

The main objective of this PR is to publish our Bitnami Magento container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### DIFF
--- a/.vib/magento/goss/goss.yaml
+++ b/.vib/magento/goss/goss.yaml
@@ -8,4 +8,5 @@ gossfile:
   ../../common/goss/templates/check-broken-symlinks.yaml: {}
   ../../common/goss/templates/check-ca-certs.yaml: {}
   ../../common/goss/templates/check-directories.yaml: {}
+  ../../common/goss/templates/check-sed-in-place.yaml: {}
   ../../common/goss/templates/check-spdx.yaml: {}
--- a/.vib/magento/goss/goss.yaml
+++ b/.vib/magento/goss/goss.yaml
@@ -1,0 +1,12 @@
+gossfile:
+  # Goss tests exclusive to the current container
+  ../../magento/goss/magento.yaml: {}
+  # Load scripts from .vib/common/goss/templates
+  ../../common/goss/templates/check-binaries.yaml: {}
+  ../../common/goss/templates/check-broken-symlinks.yaml: {}
+  ../../common/goss/templates/check-ca-certs.yaml: {}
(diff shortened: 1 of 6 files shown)


.vib/magento/goss/vars.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - composer
+directories:
+  - mode: "0775"
+    paths:
+      - /opt/bitnami/magento
+files:
+  - paths:
+      # Validate that a supported web server is configured
+      - /opt/bitnami/scripts/apache-env.sh
+      - /opt/bitnami/scripts/libapache.sh
```
FraPazGal, line 18: I'm not sure we need to check these files, as they would always be present as part of the apache nami module. We want to add a check for the postunpack logic that goes by:
```
ensure_web_server_app_configuration_exists "magento" --type php \
    --apache-move-htaccess no
```
In that case, we may want to check there is indeed no htaccess file created by apache at `/opt/bitnami/apache/conf/vhosts/htaccess/magento-htaccess.conf`

.vib/magento/goss/goss.yaml
FraPazGal, line 1: I think it would be valuable to add the template `check-apache-libphp.yaml` created by @migruiz4 at https://github.com/bitnami/containers/pull/29156

Additionally, I think you missed including the `sed-in-place.yaml` template.

FraPazGal (commented):
I just added a couple of comments, please review when you have some time @corico44

.vib/magento/goss/magento.yaml
```diff
@@ -0,0 +1,44 @@
+group:
```
FraPazGal, line 1: Reviewing possibles check for the asset itself, I think checking the magento cli works (beyond the version check) could be useful. I think we can test the modules feature by enabling one of them, WDYT? An example of this could be:

```
$ magento module:enable Magento_Store
```

.vib/magento/goss/goss.yaml
corico44: If I add the `sed-in-place.yaml `test, I get the following error:
```
Command: check-sed-in-place: exit-status:
Expected
    <int>: 1
to equal
    <int>: 0
```

FraPazGal (commented):
The PR looks ready to be merged once you remove the VIB trigger @corico44

FraPazGal (approved):

PR #32704, opened 2023-05-05
[bitnami/containers] Revisit the script to retrieve additional packages

fmulero:
### Description of the change

Revisit the script to retrieve additional packages from compilation recipe. Several changes were applied

### Benefits

Reduce number of calls to S3. Avoid use of temporal files. Keep similar code as internal projects.

### Possible drawbacks

None identified.


jotamartos (approved):
LGTM

PR #33351, opened 2023-05-10
[bitnami/gotrue] Add VIB tests

andresbono:
### Description of the change

The main objective of this PR is to publish our Bitnami GoTrue container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### DIFF
--- a/.vib/gotrue/goss/vars.yaml
+++ b/.vib/gotrue/goss/vars.yaml
@@ -1,0 +1,6 @@
+binaries:
+  - gotrue
+files:
+  - paths:
+      - /opt/bitnami/gotrue/schema.sql
+root_dir: /opt/bitnami
--- a/.vib/gotrue/goss/vars.yaml
+++ b/.vib/gotrue/goss/vars.yaml
@@ -1,6 +1,3 @@
 binaries:
   - gotrue
-files:
-  - paths:
-      - /opt/bitnami/gotrue/schema.sql
 root_dir: /opt/bitnami
--- a/.vib/gotrue/goss/goss.yaml
+++ b/.vib/gotrue/goss/goss.yaml
@@ -1,0 +1,9 @@
+gossfile:
+  # Load scripts from .vib/common/goss/templates
+  ../../common/goss/templates/check-binaries.yaml: {}
+  ../../common/goss/templates/check-broken-symlinks.yaml: {}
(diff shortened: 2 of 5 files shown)


andresbono:
Tests passing here: ✅ https://github.com/bitnami/containers/actions/runs/4936839246/jobs/8824869201

.vib/gotrue/goss/vars.yaml
```diff
@@ -0,0 +1,6 @@
+binaries:
+  - gotrue
+files:
+  - paths:
+      - /opt/bitnami/gotrue/schema.sql
```
migruiz4, line 5: I'm not sure we should be checking this file considering it is an upstream file and does not really depend on us.

I would instead suggest implementing a `command` test to check something like `find /opt/bitnami/gotrue -name "*.sql" | wc -l` is not 0 to ensure 'migrations' are present.

.vib/gotrue/goss/vars.yaml
```diff
@@ -0,0 +1,6 @@
...
+root_dir: /opt/bitnami
```
migruiz4, line 6: As it is not possible to obtain the `gotrue` version, I would suggest implementing a `gotrue help` test to ensure the main binary is executable.

migruiz4 (changes requested):
Could you please take a look at my comments?

.vib/gotrue/goss/vars.yaml
andresbono: Right, the `gotrue version` command returns a commit id, we cannot test it against `APP_VERSION`. Should we add the help test even if we know the binary is executable because it is ran in the supabase chart by default?

migruiz4: I would say so. Although downstream the `bitnami/chart` test will run the `gotrue` binary and will test more extensively its functionalities, testing here a `--version` test (or, in its absence, a `--help` test) allows us to asses that the binary is present and is executable and does not return an error before releasing the image.

Although, I would suggest not setting a `stdout` condition with many precise matches to avoid errors caused by intended changes in upstream.

migruiz4:
Looks good to me! Please remove the triggers and I will approve the PR

migruiz4 (approved):
LGTM!

PR #33366, opened 2023-05-10
[bitnami/joomla] Add VIB tests

andresbono:
### Description of the change

The main objective of this PR is to publish our Bitnami Joomla! container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### DIFF
--- a/.vib/joomla/goss/joomla.yaml
+++ b/.vib/joomla/goss/joomla.yaml
@@ -4,8 +4,10 @@ command:
     exit-status: 0
   # Placeholders in SQL installation files should have been substituted
   sql-placeholders:
-    exec: grep "#__" /opt/bitnami/joomla/installation/sql/mysql/*.sql
-    exit-status: 1
+    exec: grep "#__" /opt/bitnami/joomla/installation/sql/mysql/*.sql | wc -l
+    exit-status: 0
+    stdout:
+      - /^0/
 file:
   # Joomla! should have been configured based on build-time defaults
   /opt/bitnami/joomla/configuration.php:
@@ -18,7 +20,7 @@ file:
     exists: true
     filetype: file
     contains:
-      - /^memory_limit/
+      - "memory_limit = 256M"
   # HTTP vhost should have been properly rendered
   /opt/bitnami/apache/conf/vhosts/joomla-vhost.conf:
     exists: true
--- a/.vib/joomla/goss/joomla.yaml
+++ b/.vib/joomla/goss/joomla.yaml
@@ -1,0 +1,48 @@
+command:
+  check-app-version:
+    exec: grep {{ .Env.APP_VERSION }} /opt/bitnami/joomla/administrator/manifests/files/joomla.xml
+    exit-status: 0
+  # Placeholders in SQL installation files should have been substituted
+  sql-placeholders:
+    exec: grep "#__" /opt/bitnami/joomla/installation/sql/mysql/*.sql
+    exit-status: 1
+file:
+  # Joomla! should have been configured based on build-time defaults
+  /opt/bitnami/joomla/configuration.php:
+    exists: true
+    filetype: file
+    contains:
+      - "public $db = 'bitnami_joomla'"
+  # PHP settings should have been configured
+  /opt/bitnami/php/etc/php.ini:
+    exists: true
+    filetype: file
+    contains:
+      - /^memory_limit/
+  # HTTP vhost should have been properly rendered
+  /opt/bitnami/apache/conf/vhosts/joomla-vhost.conf:
+    exists: true
+    filetype: file
+    contains:
+      - /DocumentRoot.*/opt/bitnami/joomla/
+      - "^/administrator$"
+  # HTTPs vhost should have been properly rendered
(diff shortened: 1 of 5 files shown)


andresbono:
Tests passing:
- ✅ https://github.com/bitnami/containers/actions/runs/4938553563/jobs/8828477321
- ✅ https://github.com/bitnami/containers/actions/runs/4938553563/jobs/8828477495

.vib/joomla/goss/joomla.yaml
```diff
@@ -0,0 +1,48 @@
...
+    exists: true
+    filetype: file
+    contains:
+      - "public $db = 'bitnami_joomla'"
+  # PHP settings should have been configured
+  /opt/bitnami/php/etc/php.ini:
+    exists: true
+    filetype: file
+    contains:
+      - /^memory_limit/
```
FraPazGal, line 21: Should we would add here the the value of `memory_limit` expected? Could we have this config param set with other value at some point? I ask because joomla's postunpack logic will add or replace the config values when using `php_conf_set()`

.vib/joomla/goss/joomla.yaml
```diff
@@ -0,0 +1,48 @@
+command:
+  check-app-version:
+    exec: grep {{ .Env.APP_VERSION }} /opt/bitnami/joomla/administrator/manifests/files/joomla.xml
+    exit-status: 0
+  # Placeholders in SQL installation files should have been substituted
+  sql-placeholders:
+    exec: grep "#__" /opt/bitnami/joomla/installation/sql/mysql/*.sql
+    exit-status: 1
```
FraPazGal, line 8: This is a bit of a nitpick, but I think we should try to have our commands return an `exit-status=0` when possible. In this case, we could maybe something like:

```suggestion
    exec: if [ $(grep "#__" /opt/bitnami/joomla/installation/sql/mysql/*.sql | wc -l) -gt 0 ]; then exit 1; fi
    exit-status: 0
```

The trade-off is increasing the logic complexity, so I don't have any strong thoughts on this. Another option a bit cleaner could be:

```suggestion
    exec: grep "#__" /opt/bitnami/joomla/installation/sql/mysql/*.sql | wc -l
    exit-status: 0
    stdout:
      - /^0/
```

.vib/joomla/goss/vars.yaml
```diff
@@ -0,0 +1,18 @@
+binaries:
+  - apachectl
+  - mariadb
+  - php
+  - render-template
+files:
+  - paths:
+      - /opt/bitnami/apache/modules/libphp.so
+      - /opt/bitnami/joomla/.htaccess
+      - /opt/bitnami/joomla/configuration.php
```
FraPazGal, line 10: You are already doing a `file` test on this file, so you can drop it from this list. Actually, it is more of a need given that Goss' logic will overwrite one of the 2 checks (the same happens if you have 2 `command` tests with the same name).

.vib/joomla/goss/vars.yaml
```diff
@@ -0,0 +1,18 @@
+binaries:
+  - apachectl
+  - mariadb
+  - php
+  - render-template
+files:
+  - paths:
+      - /opt/bitnami/apache/modules/libphp.so
```
FraPazGal, line 8: This is nice, given that we do not have a libphp container. As such we need to test libphp-related logic in those containers that include said components. That said, there is already a common goss template on the works called [check-apache-libphp.yaml](https://github.com/bitnami/containers/pull/29156/files#diff-2a8ebe41b0c46f27150d4084b037555c1b5700c4488459f7e0c7514e8671204f) 😁 That should simplify these repetitive checks.

Once @migruiz4 merges the related PR I think you can add it to your `goss.yaml` file.

FraPazGal (commented):
Really nice PR @andresbono! I added a few comments as a review, please check them when you have some time.

.vib/joomla/goss/joomla.yaml
andresbono: 👍 I have added the expected value that comes from `PHP_DEFAULT_MEMORY_LIMIT`:

https://github.com/bitnami/containers/blob/8338dbb4337cc1999cd73480d781e5d6699e6989/bitnami/joomla/4/debian-11/rootfs/opt/bitnami/scripts/joomla-env.sh#L117

If the value of the env-var is ever changed, the test will need to be updated. Is that what you mean and the expected behavior, right?

.vib/joomla/goss/vars.yaml
andresbono: Interesting! You are right, the count of tests didn't change after I removed this line.

andresbono:
Tests passing:
- ✅ https://github.com/bitnami/containers/actions/runs/4979530909/jobs/8911253503
- ✅ https://github.com/bitnami/containers/actions/runs/4979530909/jobs/8911253606

.vib/joomla/goss/joomla.yaml
FraPazGal: Perfect! Yes, I think that is a valid tradeoff in this case.

FraPazGal (approved):
Thanks for the PR! LGTM!

PR #33628, opened 2023-05-12
[bitnami/parse] Add VIB tests

marcosbc:
### Description of the change

The main objective of this PR is to publish our Bitnami Parse container using VMware Image Builder. In order to do that, several changes are included:

* Increasing the existing test coverage of the asset by adding Goss tests.
* Update verify and publish VIB pipeline's definitions.

### Benefits

* Ensuring higher quality of the container catalog.
* Increased pool of assets completely handled by VMware Image Builder.

### Possible drawbacks

### Applicable issues

N/A


.vib/parse/goss/vars.yaml
```diff
@@ -0,0 +1,18 @@
+binaries:
+  - node
+  - parse-server
+  - python
+  - mongosh
+directories:
+  - mode: "0775"
+    paths:
+      - /bitnami/parse
```
FraPazGal, line 9: I think you missed adding here the `/opt/bitnami/parse/tmp` dir.

.vib/parse/goss/vars.yaml
```diff
@@ -0,0 +1,18 @@
+binaries:
+  - node
+  - parse-server
```
FraPazGal, line 3: I would maybe check the permission changes done at the postunpack stage to parse binaries affect this `parse-server`, wdyt?

.vib/parse/goss/vars.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - mongosh
+directories:
+  - mode: "0775"
+    paths:
+      - /bitnami/parse
+      - /opt/bitnami/parse
+      - /opt/bitnami/parse/lib
+      - /opt/bitnami/parse/logs
+      - /opt/bitnami/parse/public_html
+      - /opt/bitnami/parse/views
```
FraPazGal, line 14: Unless I'm mistaken, we are not explicitly setting the permissions to the directories added at compilation stage (lib, views, and public_html). I think we could move them to a separate array without the `mode` check.

.vib/parse/goss/vars.yaml
marcosbc: You're right, I've added it.

.vib/parse/goss/vars.yaml
marcosbc: I don't think it is needed. If not, `which parse-server` would not return any result. I think it is enough to ensure that it is executable, which is achievable with the `which` command (and in the end is what we are trying to achieve with the permissions change).

.vib/parse/goss/vars.yaml
marcosbc: Yes we are:

```
for dir in "$PARSE_BASE_DIR" ...; do
    ...
    configure_permissions_ownership "$dir" -d "775" -f "664" -u "$PARSE_DAEMON_USER" -g "root"
done
```

This block will recursively change the permissions in the folder. With this test, we are checking that those permissions are set properly. Note that those folders are created during the compilation.

FraPazGal: You are right, we normally do not check subfolders of a recursive permissions change but I agree that, as we already need to check them due to the compilation logic, we can also check the permissions.

FraPazGal (approved):
LGTM!

PR #34270, opened 2023-05-17
[bitnami/fluxcd-helm-controller] Add VIB tests

andresbono:
### Description of the change

The main objective of this PR is to publish our Bitnami Flux Helm Controller container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### DIFF
--- a/.vib/fluxcd-helm-controller/goss/fluxcd-helm-controller.yaml
+++ b/.vib/fluxcd-helm-controller/goss/fluxcd-helm-controller.yaml
@@ -1,0 +1,6 @@
+command:
+  check-helm-controller-help:
+    exec: helm-controller --help
+    exit-status: 2
+    stderr:
+      - "Usage of helm-controller:"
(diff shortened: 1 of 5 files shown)


.vib/fluxcd-helm-controller/goss/fluxcd-helm-controller.yaml
```diff
@@ -0,0 +1,6 @@
+command:
+  check-custom-daemon-help:
+    exec: helm-controller --help
+    exit-status: 2
```
andresbono, line 4: This might become flaky, I wonder if we just should remove this test.

joancafom: The rule of thumb here should be to check the command executes correctly (`exit-status: 0`), preferring that over specific output checks (which also may become flaky if the app changes).

This case is strange, I'd say that we should:

1. Remove the `exit-status: 2` check. We don't really care what the actual number is (other than 0)
2. Leave the `stderr` check. If this is eventually modified, we will update it.

andresbono:
Tests passing: https://github.com/bitnami/containers/actions/runs/5005863114/jobs/8970391372

joancafom (dismissed):

joancafom (approved):

PR #34293, opened 2023-05-17
[bitnami/drupal] (feat) Allows custom install directory for Drupal. Needed for some customizations

scottrigby:
### Description of the change

Allows users to customize the Drupal install directory by passing `DRUPAL_BASE_DIR` as a build arg.

Passing `DRUPAL_BASE_DIR` as a build arg also has two other effects:
- Sets `DRUPAL_BASE_DIR` env var. Note this was already an existing, undocumented env var set by `drupal-env.sh`, and used for the same purpose.
- Sets `APP_BASE_DIR` env var with the same value. This is for shared libs – libapache and libpersistence – that should not need to know about Drupal-specific vars, only that there is an `app` `install_dir`, and `document_root`.

Examples:

Building a custom image:
```console
docker build . --build-arg DRUPAL_BASE_DIR=/opt/bitnami/drupal/docroot
```

Using docker compose:
```diff
--- a/bitnami/drupal/9/debian-11/docker-compose.yml
+++ b/bitnami/drupal/9/debian-11/docker-compose.yml
@@ -10,7 +10,11 @@ services:
     volumes:
       - 'mariadb_data:/bitnami/mariadb'
   drupal:
-    image: docker.io/bitnami/drupal:9
+    # image: docker.io/bitnami/drupal:9
+    build:
+      context: .
+      args:
+        DRUPAL_BASE_DIR: /opt/bitnami/drupal/docroot
     ports:
       - '80:8080'
       - '443:8443'
```

### Benefits

Currently, users must install Drupal in `/opt/bitnami/drupal`. However, a common pattern for Drupal installations assumes certain files and directories live one directory up from the Drupal installation dir (composer.json, files, config). This change allows users to specify their Drupal install location.

### Possible drawbacks

Will not work with certain paths (for example `/`, as there is no parent directory). However, it will work with any dir below `/opt/bitnami` or custom paths such as `/foo/bar/baz`.

A user could also, in theory, specify an existing dir with conflicting file names once Drupal is unpacked there, which would be possible with an otherwise customized installation.

Users will need to know what they are doing to specify a custom Drupal base directory.

### Applicable issues

- fixes #

N/A. I could open an issue though if that would help.

### DIFF
--- a/bitnami/drupal/10/debian-11/Dockerfile
+++ b/bitnami/drupal/10/debian-11/Dockerfile
@@ -44,6 +44,19 @@ RUN apt-get autoremove --purge -y curl && \
     apt-get clean && rm -rf /var/lib/apt/lists /var/cache/apt/archives
 RUN chmod g+rwX /opt/bitnami
 
+# Optionally override DRUPAL_BASE_DIR to install Drupal in a custom directory
+ARG DRUPAL_BASE_DIR=/opt/bitnami/drupal
+ENV DRUPAL_BASE_DIR=$DRUPAL_BASE_DIR \
+    # For shared libs that don't know about Drupal vars
+    APP_BASE_DIR=$DRUPAL_BASE_DIR
+RUN \
+  if [ "${DRUPAL_BASE_DIR}" != '/opt/bitnami/drupal'  ]; then \
+    tmp=$(mktemp -d) && \
+    mv /opt/bitnami/drupal "${tmp}" && \
+    mkdir -p "$(dirname ${DRUPAL_BASE_DIR})" && \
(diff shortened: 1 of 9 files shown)


andresbono:
Hi @scottrigby, thank you very much for taking the time to contribute with this PR.

I'm hesitant to add a new dimension in the complexity of this container by allowing users to customize the install dir. It may cover some use cases like yours but it would worsen the general maintainability of the container image.

I wonder if it would be more convenient for those specific users that need Drupal to be installed in a custom path to extend the image to fit specific needs: [Drupal packaged by Bitnami: Customize this image](https://github.com/bitnami/containers/tree/main/bitnami/drupal#customize-this-image).

As always, we are monitoring the interests of our users. If more of them are willing to have this implemented, we will reopen the discussion.

PR #34297, opened 2023-05-17
[bitnami/mongodb] Configure mongodb hostname during initialization

dtrts:
### Description of the change

This change enabled the hostname to be configured for the initialization of the server. These steps include: 
- The creation of the root user
- The creation of additional users 
- The setup and configuration of a replicaset

### Benefits

When TLS is enabled through the `*_EXTRA_FLAGS` options it is enabled during initialization and it is common for the certificate to not include `127.0.0.1` in the IP list. 

This change allows us to provide a host which works with TLS. 

### Possible drawbacks

Misconfiguration could lead to configuring the wrong server. 

If attempting to use a hostname which resolves to an external IP then the mongo server will have to bind to that IP before initialization. This puts the server at risk of attack during setup. (Most likely for only a few seconds?) 

### Applicable issues

This issue is for the `charts` repo but does describe the issue here and the workarounds taken to fix it. 
- https://github.com/bitnami/charts/issues/16719

### Additional information

Have a great day!


corico44:
Hello @dtrts,

We are going to review this logic internally as we want to further investigate the behavior of the field that you propose to modify with a variable. We will notify you in this PR when there is any news. 
Thank you very much for the contribution!

bitnami/mongodb/README.md
```diff
@@ -579,6 +579,8 @@ Before starting the cluster you need to generate PEM certificates as required by
 
 Another option would be to use letsencrypt certificates; the required configuration steps for that scenario are left as an exercise for the user and are beyond the scope of this README.
 
+The hostname used during initialization can be configured using `MONGODB_LOCALHOST_NAME`. Use this to provide a vaild domain for your certificate. Preferably this hostname should be routed back to `127.0.0.1` through the hosts file. If not possible then progress can be made bybinding the mongo server to the correct ip, or all ips, in the configuration.
```
dtrts, line 582:
```suggestion
The hostname used during initialization can be configured using `MONGODB_LOCALHOST_NAME`. Use this to provide a valid domain for your certificate. Preferably this hostname should be routed back to `127.0.0.1` through the hosts file. If not possible then progress can be made by binding the MongoDB server to the correct IP, or all IPs, in the `mongodb.conf` configuration file.
https://www.mongodb.com/docs/manual/reference/configuration-options/#mongodb-setting-net.bindIp
```

dtrts:
Thank you @corico44. 

There's more detail in the issue about why I have gone for this approach. 

I have just thought of another option where support is added for different `*_EXTRA_FLAGS` or config file between initialization and the starting of the externally accessed server. 

Have a great weekend!

dtrts:
Hello @corico44 

I've been working with TLS and just wanted to highlight another sticking point. 

With TLS enabled the function `mongodb_is_mongodb_started` fails. It defaults to `get_mongo_hostname`, which in my case is using the `MONGODB_ADVERTISED_HOSTNAME`. 

During initialization I don't have this hostname IP Bound and so it reports an error. 

It could use the MONGODB_LOCALHOST_NAME by default, and we should create a separate check to ensure that the mongo server is accessible through the advertised hostname (i.e. through the internet)

Just some thoughts! Hope you're having a great week.

rrileyca:
I fully agree with this PR as using the FQDN for a Replica Set that uses an X.509 certificate signed by a public CA is the proper way forward. 

For example, the Bitnami Helm chart will break currently with all of the connections it makes to `127.0.0.1`. Both the [livenessProbe](https://github.com/bitnami/charts/blob/8245b39a2f87b9967dbac238bc585f276156a528/bitnami/mongodb/templates/common-scripts-cm.yaml#L37-L37) and [readinessProbe](https://github.com/bitnami/charts/blob/8245b39a2f87b9967dbac238bc585f276156a528/bitnami/mongodb/templates/common-scripts-cm.yaml#L21-L21) in the Bitnami Mongodb Helm chart attmpet to connect to `127.0.0.1`, making the chart impossible to use with PKI/ACME out-of-the-box.

rrileyca:
Bump

carrodher:
This issue has been inactive for a long time and is being closed as part of routine maintenance. Due to other higher-priority work, we weren’t able to investigate this further.

Thanks for your understanding.

PR #34395, opened 2023-05-18
[bitnami/kubectl] Add VIB tests

andresbono:
### Description of the change

The main objective of this PR is to publish our Bitnami Kubectl container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### DIFF
--- a/.vib/kubectl/goss/vars.yaml
+++ b/.vib/kubectl/goss/vars.yaml
@@ -1,0 +1,11 @@
+binaries:
+  - kubectl
+  - jq
+directories:
+  - mode: "0775"
+    paths:
+      - /.kube
+root_dir: /opt/bitnami
+version:
+  bin_name: kubectl
+  flag: version --client
--- a/.vib/kubectl/goss/vars.yaml
+++ b/.vib/kubectl/goss/vars.yaml
@@ -1,10 +1,5 @@
 binaries:
   - kubectl
-  - jq
-directories:
-  - mode: "0775"
-    paths:
-      - /.kube
 root_dir: /opt/bitnami
 version:
   bin_name: kubectl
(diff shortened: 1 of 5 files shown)


.vib/kubectl/goss/vars.yaml
```diff
@@ -0,0 +1,11 @@
+binaries:
+  - kubectl
+  - jq
```
andresbono, line 3: jq is installed via system packages, not sure if it is worth checking it here. WDYT?

FraPazGal: We don't usually check system pkgs installed via Dockerfile with the exception of ca-certificates, which enables https. The reason for this is that testing Dockerfile logic means we would be checking one of the following:

1. Base image distro
2. Docker logic

Both of them are external to the asset itself and are expected to work as expected and thus should not require explicit testing in our side.

That said, this is the current approach but please share your thoughts on this.

.vib/kubectl/goss/vars.yaml
```diff
@@ -0,0 +1,11 @@
...
+directories:
+  - mode: "0775"
+    paths:
+      - /.kube
```
FraPazGal, line 7: I couldn't find our logic modifying this directory, could you point me to it?

FraPazGal (commented):
Thanks for the PR @andresbono! I added a couple of comments as review, please check them when you can.

.vib/kubectl/goss/vars.yaml
andresbono: Thank you! I didn't have strong thoughts on this. The current approach makes sense to me 👍

FraPazGal (approved):
LGTM!

PR #34545, opened 2023-05-19
[bitnami/chartmuseum] Deprecate chartmuseum container

mdhont:
Removing the chartmuseum VIB tests as the ChartMuseum container is no longer maintained by upstream and has been deprecated.


corico44 (approved):
LGTM!

PR #34547, opened 2023-05-19
[bitnami/clickhouse] Add VIB tests

FraPazGal:
### Description of the change

The main objective of this PR is to publish our Bitnami Clickhouse container using VMware Image Builder. In order to do that, several changes are included:

* Increasing the existing test coverage of the asset by adding Goss tests.
* Update verify and publish VIB pipeline's definitions.

### Benefits

* Ensuring higher quality of the container catalog.
* Increased pool of assets completely handled by VMware Image Builder.

### Possible drawbacks

### Applicable issues

N/A

### DIFF
--- a/.vib/clickhouse/goss/clickhouse.yaml
+++ b/.vib/clickhouse/goss/clickhouse.yaml
@@ -1,7 +1,9 @@
 command:
   check-clickhouse-server:
-    exec: timeout --preserve-status 5 clickhouse-server
+    exec: timeout --preserve-status 5 clickhouse-server || true
     exit-status: 0
+    stdout:
+      - Ready for connections
 file:
   /var/lib/clickhouse/tmp:
     exists: true
--- a/.vib/clickhouse/goss/clickhouse.yaml
+++ b/.vib/clickhouse/goss/clickhouse.yaml
@@ -1,6 +1,6 @@
 command:
   check-clickhouse-server:
-    exec: timeout --preserve-status 5 clickhouse-server || true
+    exec: timeout --preserve-status 5 clickhouse-server 2>&1 || true
     exit-status: 0
     stdout:
       - Ready for connections
--- a/.vib/clickhouse/goss/clickhouse.yaml
+++ b/.vib/clickhouse/goss/clickhouse.yaml
@@ -1,0 +1,27 @@
+command:
+  check-clickhouse-server:
+    exec: timeout --preserve-status 5 clickhouse-server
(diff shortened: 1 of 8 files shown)


.vib/clickhouse/goss/clickhouse.yaml
```diff
@@ -0,0 +1,29 @@
+command:
+  check-clickhouse-server:
+    exec: timeout --preserve-status 5 clickhouse-server 2>&1 || true
+    exit-status: 0
```
joancafom, line 4: After our last discussion, we agreed to avoid using `|| true` when possible and prefer to explicitly check the status. Could you please update it here?

.vib/clickhouse/goss/clickhouse.yaml
```diff
@@ -0,0 +1,29 @@
...
+    stdout:
+      - Ready for connections
+file:
+  /var/lib/clickhouse/tmp:
+    exists: true
+    filetype: symlink
+    linked-to: /opt/bitnami/clickhouse/tmp
+  /opt/bitnami/clickhouse/logs/clickhouse.log:
```
joancafom, line 12: I would additionally check that `/opt/bitnami/clickhouse/logs/clickhouse_error.log` exists and is linked to `stderr`

.vib/clickhouse/goss/clickhouse.yaml
```diff
@@ -0,0 +1,29 @@
...
+    exists: true
+    filetype: symlink
+    linked-to: /dev/stdout
+  /opt/bitnami/clickhouse/etc/config.xml:
```
joancafom, line 16: I would also check that `users.xml` is updated properly to pick up the password from the env var:

https://github.com/bitnami/containers/blob/435b7670131950bc31c7b373b796efe471434146/bitnami/clickhouse/22.8/debian-11/rootfs/opt/bitnami/scripts/clickhouse/postunpack.sh#LL60C144-L60C153

joancafom (commented):

.vib/clickhouse/goss/clickhouse.yaml
FraPazGal: In this case, the `|| true` is necessary as the command output varies from `0` and `143` seemingly at random. Increasing the timeout doesn't solve this and thus I decided tu force the `exit-status=0`. I'll a comment on this test.

joancafom (approved):

PR #35280, opened 2023-05-25
[bitnami/matomo] added optional env variables to set noreply name and address

derkoenigeu:
### Description of the change

added the possibility to set the config values for noreply_email_address and noreply_email_name in the General section of the matomo config via new environment variables

new variables:
- MATOMO_NOREPLY_NAME
- MATOMO_NOREPLY_ADDRESS

### Benefits

easy way to set the mentiones values right at the start of the container inestead of having to deal with it afterwards.
makes automation of this process way faster and easier.

### Possible drawbacks

none

### DIFF
--- a/bitnami/matomo/4/debian-11/rootfs/opt/bitnami/scripts/libmatomo.sh
+++ b/bitnami/matomo/4/debian-11/rootfs/opt/bitnami/scripts/libmatomo.sh
@@ -204,6 +204,15 @@ EOF
             ini-file set -s "mail" -k "encryption" -v "$MATOMO_SMTP_PROTOCOL" "$MATOMO_CONF_FILE"
         fi
 
+        # Optional noreply name and address
+        if ! is_empty_value "$MATOMO_NOREPLY_NAME"; then
+            ini-file set -s "General" -k "noreply_email_name" -v "$MATOMO_NOREPLY_NAME" "$MATOMO_CONF_FILE"
+        fi
+
+        if ! is_empty_value "$MATOMO_NOREPLY_ADDRESS"; then
+            ini-file set -s "General" -k "noreply_email_address" -v "$MATOMO_NOREPLY_ADDRESS" "$MATOMO_CONF_FILE"
+        fi
+
         info "Persisting Matomo installation"
         persist_app "$app_name" "$MATOMO_DATA_TO_PERSIST"
     else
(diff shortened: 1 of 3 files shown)


bitnami/matomo/4/debian-11/rootfs/opt/bitnami/scripts/libmatomo.sh
```diff
@@ -204,6 +204,15 @@ EOF
...
         fi
 
+        # Optional noreply name and address
+        if ! is_empty_value "$MATOMO_NOREPLY_NAME"; then
+            ini-file set -s "General" -k "noreply_email_name" -v "$MATOMO_NOREPLY_NAME" "$MATOMO_CONF_FILE"
+        fi
+
+        if ! is_empty_value "$MATOMO_NOREPLY_ADDRESS"; then
+            ini-file set -s "General" -k "noreply_email_address" -v "$MATOMO_NOREPLY_ADDRESS" "$MATOMO_CONF_FILE"
+        fi
```
andresbono, line 214: Does it make sense to add these `ini-file set` inside the `if` block above? The one for the SMTP configuration. Do the `noreply_email_name` and `noreply_email_address` make any effect if the SMTP configuration is not set?

andresbono (commented):
Thank you very much for your contribution! We will be reviewing and testing it to have it merged.

I left a question in `libmatomo.sh` that needs clarification before proceeding.

Thanks again, @derkoenigeu!

bitnami/matomo/4/debian-11/rootfs/opt/bitnami/scripts/libmatomo.sh
derkoenigeu: I have tested the matomo configuration inside the current matomo container and you are right. The `noreply_email_name` and `noreply_email_address` won't have any effect since the default way for matomo to send mails is sendmail which doen't exist in this container anyway.
So I have made the change and updated the pr

andresbono: Thank you! We will do some additional testing and I might push additional changes to your branch regarding the environment variables. After that, the PR should be good to go.

andresbono (approved):
Thank you again for your contribution!

andresbono:
@derkoenigeu, your changes are now available in `bitnami/matomo:4.14.2-debian-11-r18` (https://github.com/bitnami/containers/pull/36435).

PR #35294, opened 2023-05-25
Specify the name of a database during init scripts

### LINKED ISSUE #35096: [bitnami/mongodb] Is there a way to specify the name of a database to be used for creation scripts?
### Name and Version

bitnami/mongodb:latest

amd64

Is there any environment variable to allows us to specify the name of a database to be used for creation scripts in `/docker-entrypoint-initdb.d/*.js`? just like `mongo` image which has `MONGO_INITDB_DATABASE` env.

_No response_

### What do you see instead?

I set these environment variables but didn't work:

- `MONGODB_DATABASE`
- `MONGODB_EXTRA_DATABASES`

I also saw this [line](https://github.com/bitnami/containers/blob/main/bitnami/mongodb/6.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb.sh#L1545) but couldn't find out a way to pass database name to `mongodb_execute_print_output` function.

### Additional information

_No response_

guoard:
### Description of the change

Add `MONGODB_INITDB_DATABASE` environment variable that allows us to specify the name of a database to be used for creation scripts in `/docker-entrypoint-initdb.d/*.js[gz]`.

### Applicable issues

- fixes #35096


Mauraza:
Hi @guoard,

Did you check [this part or the README.md](https://github.com/bitnami/containers/tree/main/bitnami/mongodb#creating-a-user-and-database-on-first-run)? 
Did you try with that variables?

guoard:
Hi @Mauraza, thank you for your review and time.

> Did you try with that variables?

I attempted it, but it also didn't work.

These environment variables only create a user with access to the designated database. When we place a `*.js` file in the `/docker-entrypoint-initdb.d` directory, they are inserted into the database using `mongosh`.

As per the MongoDB shell [documentation](https://www.mongodb.com/docs/mongodb-shell/connect/#connect-to-a-specific-database) on connecting to a specific database, if you fail to specify a database in your URI path or command while using `mongosh`, the connection will be established with the test database.

Mauraza:
Hi @guoard,

Did you check the environment variable `MONGODB_EXTRA_DATABASES`?

guoard:
> Did you check the environment variable `MONGODB_EXTRA_DATABASES`?

Yes, for the same reason I mentioned above, it didn't work either.

Mauraza:
If that is true, you don't need to create a new environment variable you can fix the environment variable (MONGODB_DATABASE and MONGODB_EXTRA_DATABASES) used for that.
Could you update the PR with this change?

Mauraza:
Hi @guoard,

I think it is better to use an init script and in the script create the database.

PR #35335, opened 2023-05-26
[bitnami/wildfly] Add VIB tests

corico44:
### Description of the change

The main objective of this PR is to publish our Bitnami Wildfly container using VMware Image Builder. In order to do that, several changes are included:

### Benefits

### Possible drawbacks

### Applicable issues

NA

### DIFF
--- a/.vib/wildfly/goss/vars.yaml
+++ b/.vib/wildfly/goss/vars.yaml
@@ -1,5 +1,6 @@
 binaries:
   - java
+  - standalone.sh
 directories:
   - mode: "0775"
     paths:
@@ -8,8 +9,6 @@ directories:
       - /opt/bitnami/wildfly/domain
       - /opt/bitnami/wildfly/standalone
       - /bitnami/wildfly/standalone/data
-  - paths:
-      - /opt/bitnami/wildfly/modules/system/layers/base
 files:
   - mode: "0664"
     paths:
--- a/.vib/wildfly/goss/vars.yaml
+++ b/.vib/wildfly/goss/vars.yaml
@@ -1,0 +1,20 @@
+binaries:
+  - java
+directories:
+  - mode: "0775"
+    paths:
+      - /opt/bitnami/wildfly
+      - /home/wildfly
+      - /opt/bitnami/wildfly/domain
+      - /opt/bitnami/wildfly/standalone
+      - /bitnami/wildfly/standalone/data
+  - paths:
+      - /opt/bitnami/wildfly/modules/system/layers/base
+files:
+  - mode: "0775"
+    paths:
+      - /opt/bitnami/wildfly/standalone.conf
+root_dir: /opt/bitnami
+version:
+  bin_name: standalone.sh
+  flag: --version
--- a/.vib/wildfly/goss/vars.yaml
+++ b/.vib/wildfly/goss/vars.yaml
@@ -11,7 +11,7 @@ directories:
   - paths:
       - /opt/bitnami/wildfly/modules/system/layers/base
 files:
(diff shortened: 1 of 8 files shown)


.vib/wildfly/goss/vars.yaml
```diff
@@ -0,0 +1,20 @@
+binaries:
+  - java
```
FraPazGal, line 2: I would also check the most important Wildfly's binaries; at least `standalone.sh`.

.vib/wildfly/goss/wildfly.yaml
```diff
@@ -0,0 +1,12 @@
...
+    exists: true
+    filetype: symlink
+    linked-to: /opt/bitnami/wildfly/standalone/deployments
+  /opt/bitnami/wildfly/modules/system/layers/base/org/wildfly/openssl/main/lib/macosx-x86_64/libwfssl.dylib:
+    exists: false
+command:
+  # The command must give exit-status 1 to verify that there is no library that doesn't have those permissions
+  check-permission-libraries:
+    exec: ls -al /opt/bitnami/wildfly/modules/system/layers/base/org/wildfly/openssl/main/lib/*/*.so 2>/dev/null | grep -v -e "-rwxr-xr--" && ls -al /opt/bitnami/wildfly/modules/system/layers/base/org/apache/activemq/artemis/journal/main/lib/linux-x86_64/libartemis-native-64.so 2>/dev/null | grep -v -e "-rwxr-xr--"
+    exit-status: 1
```
FraPazGal, line 12: The permission modifications done at compilation time on this libraries are done to avoid issues with `ldd`. Given we already have `check-linked-libraries.yaml`, I think we can avoid checking the permissions here.

.vib/wildfly/goss/vars.yaml
```diff
@@ -0,0 +1,20 @@
...
+directories:
+  - mode: "0775"
+    paths:
+      - /opt/bitnami/wildfly
+      - /home/wildfly
+      - /opt/bitnami/wildfly/domain
+      - /opt/bitnami/wildfly/standalone
+      - /bitnami/wildfly/standalone/data
+  - paths:
+      - /opt/bitnami/wildfly/modules/system/layers/base
```
FraPazGal, line 12: Are we creating this directory in our compilation or postunpack logic? I couldn't find where are we interacting with it.

.vib/wildfly/goss/wildfly.yaml
FraPazGal, line 1: It could be useful to check the JDBC connectors for both MariaDB and PostgreSQL are present, as we are adding them in the compilation stage. wdyt?

Additionally, given wildfly uses java and we are adding as a component, we could check any of the wildfly `*.sh` scripts use our java. E.g.: check for `JAVA_HOME: /opt/bitnami/java` in `jboss-cli.sh version` or `JAVA: /opt/bitnami/java/bin/java` in `standalone.sh --help`

FraPazGal (commented):
Nice PR @corico44! I added a few comments, please review them at your leisure.

.vib/wildfly/goss/vars.yaml
corico44: True. In our compilation we are using it here:
```
  get trivySkipDirs() {
    return ['modules/system/layers/base'];
  }
```
But we are already checking it here:
```
  /opt/bitnami/wildfly/modules/system/layers/base/org/wildfly/openssl/main/lib/macosx-x86_64/libwfssl.dylib:
    exists: false
```
So it isn't necessary.

FraPazGal (approved):

PR #35366, opened 2023-05-27
[bitnami/rabbitmq]: failure to start rabbitmq due to missing `curl` to load community plugins

alanwu4321:
### Description of the change

Fix the Dockerfile to prevent `curl` removal since community plugins depend on `curl` to load https://github.com/bitnami/containers/blob/a618a6d5acb2567bf5296b3f9a81f4dadd56d78a/bitnami/rabbitmq/3.11/debian-11/rootfs/opt/bitnami/scripts/librabbitmq.sh#L514 

### Benefits

Solve the `curl: command not found` error when `RABBITMQ_COMMUNITY_PLUGINS` is enabled

```yaml
        - name: RABBITMQ_COMMUNITY_PLUGINS
          value: >-
            https://github.com/rabbitmq/rabbitmq-delayed-message-exchange/releases/download/3.11.1/rabbitmq_delayed_message_exchange-3.11.1.ez
```

image

### Applicable issues

- fixes #

### DIFF
--- a/bitnami/rabbitmq/3.11/debian-11/Dockerfile
+++ b/bitnami/rabbitmq/3.11/debian-11/Dockerfile
@@ -35,8 +35,7 @@ RUN mkdir -p /tmp/bitnami/pkg/cache/ && cd /tmp/bitnami/pkg/cache/ && \
       tar -zxf "${COMPONENT}.tar.gz" -C /opt/bitnami --strip-components=2 --no-same-owner --wildcards '*/files' && \
       rm -rf "${COMPONENT}".tar.gz{,.sha256} ; \
     done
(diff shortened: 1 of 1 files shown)


rafariossaa:
Hi,
  I am sorry but `curl` is not installed in our images in order to reduce the attack surface. If you need it, you could build your own image to add it using bitnami's as base image.

csteeg:
But as @alanwu4321 mentions, the init scripts of bitnami uses curl to install plugins, this update breaks all rabbitmq instances that have the RABBITMQ_COMMUNITY_PLUGINS environment variable set. Quite a massive change to break this functionality

PR #35696, opened 2023-05-29
Removal of the curl cleanup line

### LINKED ISSUE #35694: [bitnami/rabbitmq] Removal of curl in Dockerfile of container rabbitmq version 3.11.16-2
### Name and Version

bitnami/rabbitmq:3.11.16-debian-11-r5

amd64

1. Run container with community-plugins to download
2. Watch the container crash since it can't download plugins using curl, which is now absent

Download the community plugin and start up.

### What do you see instead?

/opt/bitnami/scripts/librabbitmq.sh: line 514: curl: command not found

### Additional information

This is where curl is needed on startup: https://github.com/bitnami/containers/blob/main/bitnami/rabbitmq/3.11/debian-11/rootfs/opt/bitnami/scripts/librabbitmq.sh#L514 (same on 3.9 and 3.10)

This is a problem since "rabbitmq:3.11.16-debian-11-r5" but is also present on "rabbitmq:3.11.16-debian-11-r6".
Here is the problematic commit: (On the added L38) https://github.com/bitnami/containers/commit/a618a6d5acb2567bf5296b3f9a81f4dadd56d78a

Impacts all the deployments using the latest version of the bitnami/rabbitmq helm chart using community-plugins.

ysma500:
### Description of the change

Keeping curl installed in the rabbitmq images since it's used in runtime to install "RabbitMQ - community-plugins" (as documented in linked issue).

### Benefits

Fixes https://github.com/bitnami/containers/issues/35694

### Possible drawbacks

I don't how and where the [vmware-bot](https://github.com/bitnami/containers/commit/a618a6d5acb2567bf5296b3f9a81f4dadd56d78a) is updating those Dockerfile from, my fix may only hold until it does an other run at it.

Also, the removal of curl in the containers was maybe due to security issues that I would bring back with this change.

### Applicable issues

- fixes #35694

### DIFF
--- a/bitnami/rabbitmq/3.10/debian-11/Dockerfile
+++ b/bitnami/rabbitmq/3.10/debian-11/Dockerfile
@@ -35,8 +35,7 @@ RUN mkdir -p /tmp/bitnami/pkg/cache/ && cd /tmp/bitnami/pkg/cache/ && \
       tar -zxf "${COMPONENT}.tar.gz" -C /opt/bitnami --strip-components=2 --no-same-owner --wildcards '*/files' && \
       rm -rf "${COMPONENT}".tar.gz{,.sha256} ; \
     done
-RUN apt-get autoremove --purge -y curl && \
-    apt-get update && apt-get upgrade -y && \
+RUN apt-get update && apt-get upgrade -y && \
     apt-get clean && rm -rf /var/lib/apt/lists /var/cache/apt/archives
 RUN chmod g+rwX /opt/bitnami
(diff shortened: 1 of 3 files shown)


carrodher:
Thanks, we already re-added `curl` in our internal test & release pipeline which is in charge of releasing new versions of the container image as well as updating the source code present in this repo.

For instance, in https://github.com/bitnami/containers/pull/35981 or https://github.com/bitnami/containers/pull/35980 the specific line removing `curl` was removed.

PR #36733, opened 2023-06-07
[bitnami/pgbouncer] Make hba file path configurable

gtato:
### Description of the change

The change makes hba file configurable in the pgbouncer config file. 
This means that the user can provide an environment variable specifying the path of the hba file, and this will be included in the config file, when the auth_type is hba.

### Benefits

Currently it is not straightforward to specify the hba file if the auth_type is set to hba. Having a hba file in place is important for the security of the database behind pgbouncer. 

### Applicable issues

- fixes #


jotamartos (approved):
Thank you for taking the time to create this PR! We will integrate these changes and release a new version of the container soon.

jotamartos:
A new version of the container was just released 

https://github.com/bitnami/containers/pull/36804

jotamartos:
Just forgot to mention that the README will be also updated with the new env var during the next release.

PR #37540, opened 2023-06-14
[bitnami/pgbouncer] Add reserve_pool_timeout to available environment variables

maxzintel:
### Description of the change

Adds reserve pool timeout to PgBouncer's list of environment variables.

### Benefits

By default, PgBouncer sets this value to 5 seconds. For some use-cases, this is too high. By adding the environment variable to the Bitnami container we make it easier for users to set this to what they want.


aoterolorenzo (approved):
LGTM! Thanks for the contribution @maxzintel

PR #37556, opened 2023-06-14
Change spaces in nginx.conf file

aver33:
### Description of the change
Change spaces in nginx.conf file

### Benefits

This is necessary for logging systems e.g. fluentd. Logging systems cannot parsed logs because config uses odd space. For example, nginx-ingress-controller is not using odd space and its parsing is ok.


CeliaGMqrz (approved):
Hi @aver33 

Thanks for your contribution! LGTM

PR #35992, opened 2023-05-31
[bitnami/redis-cluster] Update redis-cluster-env.sh

byErikas:
Adding file reading support for missing enviroment variable `REDISCLI_AUTH`

### Description of the change
- Added `REDISCLI_AUTH` to `redis_cluster_env_vars` array inside `/opt/bitnami/scripts/redis-cluster-env.sh`
- Added an `export REDIS_CLI` command to the same script as above to keep it set

### Benefits
- Allows for the ability to use `REDISCLI_AUTH` with a file instead of a variable
- Provides consistency when used with `REDIS_PASSWORD_FILE`

### Possible drawbacks
- Worth noting I do not know if this is omitted on purpose or just got forgotten, if omitted on purpose this may have breaking changes

### Applicable issues
- fixes #35982

PR #37778, opened 2023-06-15
[redis-cluster] fix REDISCLI_AUTH  to allow clusters with ALLOW_EMPTY_PASSWORD

### LINKED ISSUE #37757: [bitnami/redis-cluster] Cluster no longer allows having no password for the default user
### Name and Version

bitnami/redis-cluster:6.2.12-debian-11-r19 (the most recent 7.0 as well)

arm64

1.) set up a redis-cluster using `ALLOW_EMPTY_PASSWORD`
```yaml
version: '2'
services:

  redis-cluster-1:
    image: docker.io/bitnami/redis-cluster:6.2.12-debian-11-r19
    environment:
      ALLOW_EMPTY_PASSWORD: yes
      REDIS_NODES: redis-cluster-0 redis-cluster-1 redis-cluster-2 redis-cluster-3 redis-cluster-4 redis-cluster-5

  redis-cluster-2:
    image: docker.io/bitnami/redis-cluster:6.2.12-debian-11-r19
    environment:
      ALLOW_EMPTY_PASSWORD: yes
      REDIS_NODES: redis-cluster-0 redis-cluster-1 redis-cluster-2 redis-cluster-3 redis-cluster-4 redis-cluster-5

  redis-cluster-3:
    image: docker.io/bitnami/redis-cluster:6.2.12-debian-11-r19
    environment:
      ALLOW_EMPTY_PASSWORD: yes
      REDIS_NODES: redis-cluster-0 redis-cluster-1 redis-cluster-2 redis-cluster-3 redis-cluster-4 redis-cluster-5

  redis-cluster-4:
    image: docker.io/bitnami/redis-cluster:6.2.12-debian-11-r19
    environment:
      ALLOW_EMPTY_PASSWORD: yes
      REDIS_NODES: redis-cluster-0 redis-cluster-1 redis-cluster-2 redis-cluster-3 redis-cluster-4 redis-cluster-5

  redis-cluster-5:
    image: docker.io/bitnami/redis-cluster:6.2.12-debian-11-r19
    environment:
      ALLOW_EMPTY_PASSWORD: yes
      REDIS_NODES: redis-cluster-0 redis-cluster-1 redis-cluster-2 redis-cluster-3 redis-cluster-4 redis-cluster-5

  redis-cluster-0:
    image: docker.io/bitnami/redis-cluster:6.2.12-debian-11-r19
    depends_on:
      - redis-cluster-1
      - redis-cluster-2
      - redis-cluster-3
      - redis-cluster-4
      - redis-cluster-5
    environment:
      ALLOW_EMPTY_PASSWORD: yes
      REDIS_CLUSTER_REPLICAS: 1
      REDIS_NODES: redis-cluster-0 redis-cluster-1 redis-cluster-2 redis-cluster-3 redis-cluster-4 redis-cluster-5
      REDIS_CLUSTER_CREATOR: yes
      BITNAMI_DEBUG: 1
```

2.) run `docker compose up`

The cluster should successfully boot up (this is what happens when we swap out the image for `bitnami/redis-cluster:6.2.12-debian-11-r18`)

### What do you see instead?

The cluster fails to boot up with auth errors
```
binami-redis-cluster-redis-cluster-0-1  | AUTH failed: ERR AUTH <password> called without any password configured for the default user. Are you sure your configuration is correct?
binami-redis-cluster-redis-cluster-0-1  | Node redis-cluster-0 not ready, waiting for all the nodes to be ready...
```

### Additional information

I expect it is changed due to https://github.com/bitnami/containers/pull/35992 . I think the inclusion of `export REDISCLI_AUTH="${REDISCLI_AUTH:-}"` which will define the variable `REDISCLI_AUTH` as the empty string if it is not defined.

I believe the empty string is read by the redis cli [here](https://github.com/redis/redis/blob/unstable/src/redis-cli.c#L2973-L2975) and treated as auth instead of no auth (which having it undefined would do)

I can make a change that will only export that variable if it is defined (as opposed to exporting the empty string) if we think that is the best approach to still support "ALLOW_EMPTY_PASSWORD"

ewilliamson-patreon:
### Description of the change
The `REDISCLI_AUTH` flag had begun to be exported in https://github.com/bitnami/containers/pull/35992 . 

If `REDISCLI_AUTH` is not defined it was still exported as an empty string (which [redis-cli interpretes as the auth](https://github.com/redis/redis/blob/unstable/src/redis-cli.c#L2973-L2975 )). 

By exporting as an empty string (with `ALLOW_EMPTY_PASSWORD`)  we hit the issue of 
```
binami-redis-cluster-redis-cluster-0-1  | AUTH failed: ERR AUTH <password> called without any password configured for the default user. Are you sure your configuration is correct?
binami-redis-cluster-redis-cluster-0-1  | Node redis-cluster-0 not ready, waiting for all the nodes to be ready...
```

This was a regression from before where it was not defined and `ALLOW_EMPTY_PASSWORD` would work fine.

In this change I make it possible to set `REDISCLI_AUTH` if you want, but keep the value unset as the default instead of the empty string. 

### Benefits

We are still able to support `ALLOW_EMPTY_PASSWORD` for development purposes.

### Possible drawbacks

A bit more complexity as `REDISCLI_AUTH` behaves differently by default that the other environment variables. But that is necessary given the semantics of how it is used

### Applicable issues

- fixes #37757


mdhont:
Thank you for your PR, we need to test this solution on our internal systems, I will update this thread with further information.

PR #38448, opened 2023-06-21
[bitnami/attu,milvus] chore: :construction_worker: Add VIB integration

javsalgar:
### Description of the change

Add VIB files for Milvus and Attu. Milvus is only for amd64 and Attu for ARM and AMD64
### Applicable issues

- fixes #

### DIFF
--- a/.vib/attu/vib-publish.json
+++ b/.vib/attu/vib-publish.json
@@ -1,0 +1,123 @@
+{
+  "context": {
+    "resources": {
+      "url": "{VIB_ENV_CONTAINER_URL}",
+      "path": "{VIB_ENV_PATH}"
+    },
+    "runtime_parameters": "Y29tbWFuZDogWyJ0YWlsIiwgIi1mIiwgIi9kZXYvbnVsbCJd"
+  },
+  "phases": {
+    "package": {
+      "actions": [
+        {
+          "action_id": "container-image-package",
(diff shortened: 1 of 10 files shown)


.vib/attu/vib-publish.json
```diff
@@ -0,0 +1,123 @@
+{
+  "context": {
+    "resources": {
+      "url": "{VIB_ENV_CONTAINER_URL}",
+      "path": "{VIB_ENV_PATH}"
+    },
+    "runtime_parameters": "Y29tbWFuZDogWyJ5YXJuIiwgInN0YXJ0OnByb2QiXQ=="
```
FraPazGal, line 7: For now, and following our container testing philosophy, all of our container tests use the same `runtime_parameters`. That is:

```
command: ["tail", "-f", "/dev/null"]
```

It's true this has some drawbacks, but given the limitations to test more complex containers, it was decided to have the same "initialisation" logic for all of our test suites.

In this case, the start command should be executed as a Goss test. You can check `supabase-studio` for a similar scenario:

https://github.com/bitnami/containers/blob/5bce30a0d14919827a899b047a2d33746af164d1/.vib/supabase-studio/goss/supabase-studio.yaml#L1-L8

FraPazGal (approved):
LGTM!

Issue #38803, opened 2023-06-24
[bitnami/gitea] setup does not have ssh-keygen for configuring access keys


christhomas:
### Name and Version

bitnami/gitea

### What architecture are you using?

amd64

### What steps will reproduce the bug?

From your local machine, run the following:
- This does not work: `docker run -it bitnami/gitea ssh-keygen`
- This works: `docker run -it gitea/gitea ssh-keygen`

ssh-keygen is important when you need to setup access keys in your profile, so without this tool, gitea seems to fail and I think that should work

### What is the expected behavior?

That I can run ssh-keygen from the console, as given in the above instructions. 

### What do you see instead?

In gitea, when adding my public key, I have the following error
```
Cannot verify your SSH key: SSHKeyGenParsePublicKey: fail to parse public key: exec: "ssh-keygen": executable file not found in $PATH -
```

javsalgar:
Hi!

Thank you so much for the PR! The team will take a look

PR #38805, opened 2023-06-24
[bitnami/gitea] add ssh-client to the packages installed so I can use ssh-keygen to my account

christhomas:
This pull request fixes this issue: https://github.com/bitnami/containers/issues/38803

The problem is that in gitea there is a functionality that will allow you to add ssh public keys to your account, the problem is when you try this functionality with the original code, it does not have the ssh-client package installed and therefore you will get a problem saying that "ssh-keygen" is not in the $PATH.

So I added it to the packages installed. I tried it locally and also I deployed this image into a kubernetes cluster and then added my key to my account without any issue.

### DIFF
--- a/bitnami/gitea/1/debian-11/Dockerfile
+++ b/bitnami/gitea/1/debian-11/Dockerfile
@@ -19,7 +19,7 @@ ENV HOME="/" \
 COPY prebuildfs /
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 # Install required system packages and dependencies
-RUN install_packages acl ca-certificates curl git procps
+RUN install_packages acl ca-certificates curl git procps ssh-client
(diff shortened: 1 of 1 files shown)


javsalgar:
Hi!

christhomas:
Why is this not merged?

carrodher:
It was automatically closed due to the inactivity after the previous comment asking to sign the DCO

christhomas:
I don't actually know what the DCO is, how to fix it, or even sign one. So what am I supposed to do now?

carrodher:
There is an action failing that is in charge of checking the DCO (Developer Certificate of Origin), see https://github.com/bitnami/containers/pull/38805/checks?check_run_id=14526772275. There you can see some instructions to fix the issue.

PR #39867, opened 2023-07-04
[bitnami/redis-cluster] Revert "Update redis-cluster-env.sh (#35992)"

joancafom:
### Description of the change

#35992 introduced a regression in which the cluster can't be correctly created when no password is specified. Setting `REDISCLI_AUTH=` is not interpreted in the same way by `redis-cli` than not setting the variable.

This reverts commit 064b810ca7b6703bccc74b1f4da799fb27905cd9.

### Benefits

The cluster can be created again when no password is specified.

### Possible drawbacks

`REDISCLI_AUTH` will not be able to be declared using a file, which was the original scope of the PR. We will internally evaluate this again.

### Applicable issues

- Relates to https://github.com/bitnami/charts/issues/17338

### Additional information

NA


fevisera (approved):
LGTM!

PR #41416, opened 2023-07-19
[bitnami/*] Create a TESTING.md document

FraPazGal:
### Description of the change

This PR adds the testing documentation for our `bitnami/containers` catalog, detailing how the current strategy works and the steps to create a new test suite.


.vib/common/goss/templates/check-ca-certs.yaml
```diff
@@ -2,8 +2,9 @@
 # SPDX-License-Identifier: APACHE-2.0
 
 ########################
-# Checks the ca-certificates package
-# is installed accesing an HTTPS URL
+# Checks the ca-certificates package is installed accesing an HTTPS URL
+# Needed vars:
+# - None
```
joancafom, line 7: Please, create a separate PR for these files

TESTING.md
```diff
@@ -0,0 +1,293 @@
...
+        ...
+      ]
+    }
+  }
+}
+```
+
+This guide will focus on the `verify` phase section, of which there are some things to remark on:
+
+* For the testing of containers, VIB will take the built container in the previous `package` phase and include it in a basic Helm chart template composed of a deployment and service template. Beware VIB will only be taking into account the docker image `ENTRYPOINT/CMD` command of the image. Consequently, this both simplifies and limits the configurability of the template chart and container image tested underneath.
```
joancafom, line 88:
```suggestion
* For the testing of containers, VIB will take the container built in the previous `package` phase and include it in a basic Helm chart template composed of a deployment and service template. Beware VIB will only be taking into account the docker image `ENTRYPOINT/CMD` command of the image. Consequently, this both simplifies and limits the configurability of the template chart and container image tested underneath.
```

joancafom, line 88: I'd separate this into two points: one that explains that a chart template is used and another one for the entrypoint.

```suggestion
* For the testing of containers, VIB will take the built container in the previous `package` phase and include it in a basic Helm chart template composed of a deployment and service template. 

* VIB does only allow to modify the `ENTRYPOINT/CMD` of the image (through `runtime_parameters`). Consequently, this both simplifies and limits the configurability of the template chart and container image tested underneath.
```

TESTING.md
```diff
@@ -0,0 +1,293 @@
...
+
+* A container's testing phase will usually include a single `goss` testing action, followed by additional security-related actions.
+
+### vib-verify.json vs vib-publish.json
+
+Going back to what we explained in the introduction, two different events will trigger the test's execution. The following two files are associated with those events respectively:
```
joancafom, line 94:
```suggestion
Going back to what we explained in the introduction, two different events will trigger the tests's execution. The following two files are associated with those events respectively:
```

TESTING.md
```diff
@@ -0,0 +1,293 @@
...
+
+* The `vib-verify.json` pipeline definition file will be used to verify the changes proposed in a PR.
+* The `vib-publish.json` file will instead define the pipeline launched when the proposed changes are merged to `main`.
+
+Both files define what VIB should do when they are triggered and thus tweaking the files allows to define different action policies depending on the event that was fired. Nevertheless, it was decided that the verification process should be identical in both cases. Therefore, the `verify` section in `vib-verify.json` and `vib-verify.json` files must coincide.
```
joancafom, line 99:
```suggestion
Both files define what VIB should do when they are triggered and thus tweaking them allows to define different action policies depending on the event that was fired. Nevertheless, it was decided that the verification process should be identical in both cases. Therefore, the `verify` section in `vib-verify.json` and `vib-verify.json` files must coincide.
```

joancafom (commented):
Thanks for the PR! I have not finished the review yet, but here you have some comments I've noted down until now

TESTING.md
```diff
@@ -0,0 +1,291 @@
...
+* ❌ Manually finishing the container initialization to run functional tests
+* ✅ Checking the image filesystem (created dirs existence, changed permissions, etc) related to the compilation/postunpack stages
+* ❌ Verifying bash logic performed at `libwordpress.sh`, as we'll run Goss before that logic is executed
+* ✅ Testing binaries' existence and usability
+
+Something of note is the equality of scope for the whole container catalog. Though some apps do not require additional containers to run and can be fully initialised as they are, we will be using the same testing setup for every container.
+
+### Runtime parameters
+
+As of now, and linking with the scope definition we saw previously, the `runtime_parameters` field is only used to "stop" a container initialization logic after its `postunpack` has been executed. The `runtime_parameters` value is a base64 encoded string going by `command: ["tail", "-f", "/dev/null"]` and is the same for every VIB pipeline defined in `bitnami/containers`.
```
joancafom, line 121:
```suggestion
As of now, and linking with the scope definition we saw previously, the `runtime_parameters` field is only used to "stop" a container's initialization logic after its `postunpack` has been executed. The `runtime_parameters` value is a base64 encoded string going by `command: ["tail", "-f", "/dev/null"]` and is the same for every VIB pipeline defined in `bitnami/containers`.
```

TESTING.md
```diff
@@ -0,0 +1,291 @@
...
+
+* [ ] The optional `app.yaml` file is where custom tests created just for the related app verification are included.
+* [ ] The `goss.yaml` file should include the list of used GOSS templates as well as the `app.yaml` file (if necessary).
+* [ ] The `vars.yaml` file should include every variable used in the templates.
+* [ ] The `testfiles` folder should include any external files used in the `app.yaml` tests.
+
+Not every suite will be composed of the same tests, as it will depend on the type of application, its Dockerfile, and the used compilation/configuration logic. The list below details each pillar that should be checked when creating a test suite as well as when to use some of the most common templates:
+
+* Dockerfile check:
+  * Does it include nami modules which contain binaries? If so, use the `check-binaries.yaml` GOSS template.
```
joancafom, line 226: I don't think this terminology is as clear as to us for external contributors. 

```suggestion
  * Does it include bitnami sub-modules which contain binaries? If so, use the `check-binaries.yaml` GOSS template.
```

TESTING.md
```diff
@@ -0,0 +1,291 @@
...
+  * Does it include the ca-certificates package? If so, use the `check-ca-certificates.yaml` template.
```
joancafom, line 227:
```suggestion
  * Does it include the `ca-certificates` package? If so, use the `check-ca-certificates.yaml` template.
```

TESTING.md
```diff
@@ -0,0 +1,291 @@
...
+* [ ] Due to GOSS limitations, there can't be two tests with the same name or checking the same file/directory. In those cases, only one of them will be run.
+* [ ] For clarity purposes, the vars needed for the GOSS templates shouldn't be used in the custom tests at `app.yaml`.
+* [ ] Tests should not rely on system packages (e.g. `curl`). Favor built-in GOSS primitives instead.
+* [ ] Prefer checking the exit status of a command rather than looking for a specific output. This will avoid most of the potential flakiness.
+
+### Run GOSS locally
+
+Sometimes it is of interest to run the tests locally, for example during development. Though there may be different approaches, you may follow the steps below to execute the tests locally:
+
+1. Download the [GOSS binary for Linux AMD64](https://github.com/goss-org/goss/releases/)
```
joancafom, line 270:
```suggestion
1. Download the [GOSS binary for Linux](https://github.com/goss-org/goss/releases/)
```

TESTING.md
FraPazGal: You are right, I'll change this to the one we actually use nowadays; _bitnami components_.

TESTING.md
FraPazGal: Good catch, I'll also modify the `amd64`reference below.

joancafom (approved):

PR #41765, opened 2023-07-21
Notify failing CD pipelines in Slack

carrodher:
### Description of the change

Use the same approach as the one used in bitnami/charts to notify failing CD pipelines in Slack

### DIFF
--- a/.github/workflows/cd-pipeline.yml
+++ b/.github/workflows/cd-pipeline.yml
@@ -1,7 +1,7 @@
 # Copyright VMware, Inc.
 # SPDX-License-Identifier: APACHE-2.0
 
-name: '[CI/CD] CD Pipelin'
+name: '[CI/CD] CD Pipeline'
 run-name: "${{ format('[CI/CD] CD Publish {0}',github.event.workflow_run.display_title) }}"
 on: # rebuild any PRs and main branch changes
   workflow_run:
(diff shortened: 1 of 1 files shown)


.github/workflows/cd-pipeline.yml
```diff
@@ -1,4 +1,7 @@
-name: '[CI/CD] CD Pipeline Run'
+# Copyright VMware, Inc.
+# SPDX-License-Identifier: APACHE-2.0
+
+name: '[CI/CD] CD Pipelin'
```
FraPazGal, line 4:
```suggestion
name: '[CI/CD] CD Pipeline'
```

.github/workflows/cd-pipeline.yml
```diff
@@ -87,3 +90,45 @@ jobs:
...
+      - vib-publish
+      - update-index
+    if: always()
+    name: Notify unsuccessful CD run
+    steps:
+      - name: Notify in Slack channel
+        if: ${{ needs.vib-publish.result != 'success' || needs.update-index.result != 'success' }}
+        uses: slackapi/slack-github-action@v1.23.0
+        with:
+          channel-id: ${{ secrets.CD_SLACK_CHANNEL_ID }}
```
FraPazGal, line 107: Both this envar and `CD_SLACK_BOT_TOKEN` are set at the repo level for `bitnami/charts`, so we may need to add them in this repo still (I can't access the settings and confirm).

FraPazGal (commented):
Thanks for the PR @carrodher! I added a couple of minor comments, please check them whenever you can.

.github/workflows/cd-pipeline.yml
carrodher: Thanks for your review, both should be added, see 
Screenshot 2023-07-21 at 17 49 02

PR #41788, opened 2023-07-22
Fix access log not following specified format

fzyzcjy:
### Description of the change

From my experiments, it seems that things like http_x_forwarded_for is never printed using the container. It seems to be caused by missing to specify the log format name ("main") in access_log command. I will double check and update here.

The nginx counterpart does have `main`: https://github.com/fzyzcjy/containers/blob/4523a19c498fcbd854d4079207b58761d881874f/bitnami/nginx/1.25/debian-11/rootfs/opt/bitnami/nginx/conf/nginx.conf#L18

### Applicable issues

- fixes #


jotamartos (approved):
Thanks for taking the time to create this PR. Our pipeline will take care of updating the container with this change.

fzyzcjy:
You are welcome!

jotamartos:
Version [1.21.4-2-debian-11-r5](https://github.com/bitnami/containers/pull/42114) of the container contains these changes

PR #43135, opened 2023-08-01
[bitnami/kafka] Allow setting ssl.client.auth separately for inter-broker, controller, and client listeners

Rablet:
### Description of the change

The kafka container already has different settings for client auth for client, inter-broker, and controller listeners but it currently uses the same setting for all listeners.

This change allows users to set it separately for inter-broker, controller, and client listeners.

### Benefits

This allows ssl client auth for different categories of listeners

### Possible drawbacks

An even better change might be to make it configurable per listener (for example, you might have a CLIENT listener which doesn't need it, and an EXTERNAL listener which does need it). This change keeps it simple by retaining the existing settings.

### Applicable issues

N/A

### Additional information

N/A

### DIFF
--- a/bitnami/kafka/3.5/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
+++ b/bitnami/kafka/3.5/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
@@ -978,7 +978,13 @@ kafka_initialize() {
             listener_lower="$(echo "$listener" | tr '[:upper:]' '[:lower:]')"
 
             if [[ "$protocol" = "SSL" || "$protocol" = "SASL_SSL" ]]; then
-                kafka_server_conf_set "listener.name.${listener_lower}.ssl.client.auth" "$KAFKA_TLS_INTER_BROKER_AUTH"
+                if [[ "$listener" = "${KAFKA_CFG_INTER_BROKER_LISTENER_NAME:-INTERNAL}" ]]; then
+                    kafka_server_conf_set "listener.name.${listener_lower}.ssl.client.auth" "$KAFKA_TLS_INTER_BROKER_AUTH"
+                elif [[ "${KAFKA_CFG_CONTROLLER_LISTENER_NAMES:-CONTROLLER}" =~ $listener ]]; then
+                    kafka_server_conf_set "listener.name.${listener_lower}.ssl.client.auth" "$KAFKA_TLS_CONTROLLER_AUTH"
+                else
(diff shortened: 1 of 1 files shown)


bitnami/kafka/3.5/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
```diff
@@ -978,7 +978,13 @@ kafka_initialize() {
             listener_lower="$(echo "$listener" | tr '[:upper:]' '[:lower:]')"
 
             if [[ "$protocol" = "SSL" || "$protocol" = "SASL_SSL" ]]; then
-                kafka_server_conf_set "listener.name.${listener_lower}.ssl.client.auth" "$KAFKA_TLS_INTER_BROKER_AUTH"
+                if [[ "$listener" = "${KAFKA_CFG_INTER_BROKER_LISTENER_NAME:-INTERNAL}" ]]; then
+                    kafka_server_conf_set "listener.name.${listener_lower}.ssl.client.auth" "$KAFKA_TLS_INTER_BROKER_AUTH"
+                elif [[ "${KAFKA_CFG_CONTROLLER_LISTENER_NAMES:-CONTROLLER}" =~ $listener ]]; then
```
jotamartos, line 983: Please use the same the syntax than in the first if

```suggestion
                elif [[ "$listener" = "${KAFKA_CFG_CONTROLLER_LISTENER_NAMES:-CONTROLLER}" ]]; then
```

jotamartos (commented):
Please take a look at my comment and I'll take care of updating our systems once these changes are merged.

bitnami/kafka/3.5/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
Rablet: > Please use the same the syntax than in the first if

I used that format in order to be consistent with how the controller listener is detected on line 995:
`elif [[ "${KAFKA_CFG_CONTROLLER_LISTENER_NAMES:-CONTROLLER}" =~ $listener ]]; then`
My assumption is it was done that way because the config is a comma separated list rather than just a single name.

In light of that, would you still like me to change it?

Rablet:
Looks like this was implemented in a different way recently:
`KAFKA_TLS_<uppercase_controller_listener_name>_CLIENT_AUTH: Configures mTLS authentication method for kafka control plane communications. Allowed values: required, requested, none.`
Closing this.

PR #43871, opened 2023-08-07
[bitnami/moodle] Update README.md to include information about debug logging

SoraMakes:
### Description of the change

add information about debug logging to the README.md file

### Benefits

Users and Developers can greatly benefit of additional logging. In my case the installation failed in a test environment where the admin password was set to "admin". Enabling the additional debug logging revealed the error, but the switch to enable that is not documented.

### Possible drawbacks

None

### Applicable issues

None


migruiz4 (approved):
Thank you, looks good to me

PR #44732, opened 2023-08-15
[bitnami/openldap:2.6] Additional env vars

### LINKED ISSUE #1039: Add flag to force clear text password hash
Hi,

First thanks for this high quality container!

I have a very similar issue as described [here](https://github.com/osixia/docker-openldap/issues/208#issuecomment-536370016). In short I'd like to set a ppolicy that forces password hash even if they are sent in clear text by the user, and the client is not respecting the `ldap password modify extended operation`.   
In the posted link  they solve it by adding  `olcPPolicyHashCleartext` attribute and set to true. 

I've naively tried to make a similar modification as described in the last comment but failed (not an ldap expert), probably the DN is different, I get this error:
```bash
$ ldapadd -Q -Y EXTERNAL -H ldapi:/// -f  _file.ldif
modifying entry "cn=module,cn=config"
ldap_modify: No such object (32)
        matched DN: cn=config
```
Having a flag to turn on this feature would be very nice. Do you have any suggestions on how to achieve this?

Thanks!

Horstexplorer:
#### Added environment variables
`LDAP_REQUIRE_TLS`
Configures whether connections must use TLS to connect. Can only be applied with LDAP_ENABLE_TLS active. Defaults to `no`.

`LDAP_PASSWORD_HASH`
Configures the hash to be used when generating user passwords. Must be one of {SSHA}, {SHA}, {SMD5}, {MD5}, {CRYPT}, and {CLEARTEXT}. Defaults to **{SSHA}**.

`LDAP_CONFIGURE_PPOLICY`
Enables the ppolicy module and creates an empty configuration. Defaults to **no**

`LDAP_PPOLICY_USE_LOCKOUT`
Whether bind attempts to locked accounts will always return an error. Will only be applied with `LDAP_CONFIGURE_PPOLICY` active. Default: **no**

`LDAP_HASH_CLEARTEXT`
Whether plaintext passwords should be hashed automatically. Will only be applied with `LDAP_CONFIGURE_PPOLICY` active. Default: **no**

#### Modified behaviour
`LDAP_ALLOW_ANON_BINDING`
Regarding the note on bind_anon provided by the openldap configuration reference, disabling this env var will now also enable the requirement for authentication.
 > Disabling the anonymous bind mechanism does not prevent anonymous access to the directory. ... one should instead specify "require authc".

Modified the initial configuration sequence so that tls settings will be applied after everything else to avoid the configuration to fail after `LDAP_REQUIRE_TLS` has been enabled.

### Benefits

Adds more accessible configuration options.

### Possible drawbacks

The default configurations for the newly added env vars have been selected to ensure that OpenLDAP maintains continuity with its previous behavior.
Modifications done to `LDAP_ALLOW_ANON_BINDING` as well as the initial configuration sequence did appear to have an discernible impact.

### Applicable issues

Should resolve #1039 and similar.

### Additional information

It might be useful to additionally reconsider the default values of `LDAP_ALLOW_ANON_BINDING` and `LDAP_REQUIRE_TLS` to provide a more secure out of the box configuration.

### DIFF
--- a/bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
+++ b/bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
@@ -622,10 +622,10 @@ ldap_initialize() {
         fi
         # enable tls
         if is_boolean_yes "$LDAP_ENABLE_TLS"; then
-          ldap_configure_tls
-          if is_boolean_yes "$LDAP_REQUIRE_TLS"; then
-            ldap_configure_tls_required
-          fi
+            ldap_configure_tls
+            if is_boolean_yes "$LDAP_REQUIRE_TLS"; then
+                ldap_configure_tls_required
+            fi
         fi
         ldap_stop
     fi
@@ -775,7 +775,7 @@ olcPPolicyHashCleartext: TRUE
 EOF
     debug_execute ldapmodify -Q -Y EXTERNAL -H "ldapi:///" -f "${LDAP_SHARE_DIR}/ppolicy_configuration_hash_cleartext.ldif"
     fi
-  # enable ppolicy_use_lockout
+    # enable ppolicy_use_lockout
     if is_boolean_yes "$LDAP_PPOLICY_USE_LOCKOUT"; then
         info "Enabling ppolicy_use_lockout"
         cat > "${LDAP_SHARE_DIR}/ppolicy_configuration_use_lockout.ldif" << EOF
--- a/bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
+++ b/bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
@@ -403,6 +403,9 @@ dn: cn=config
 changetype: modify
 add: olcDisallows
 olcDisallows: bind_anon
+-
+add: olcRequires
+olcRequires: authc
 EOF
     debug_execute ldapmodify -Y EXTERNAL -H "ldapi:///" -f "${LDAP_SHARE_DIR}/disable_anon_bind.ldif"
 }
--- a/bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
+++ b/bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
@@ -71,8 +71,10 @@ export LDAP_REQUIRE_TLS="${LDAP_REQUIRE_TLS:-no}"
 export LDAP_ULIMIT_NOFILES="${LDAP_ULIMIT_NOFILES:-1024}"
 export LDAP_ALLOW_ANON_BINDING="${LDAP_ALLOW_ANON_BINDING:-yes}"
 export LDAP_LOGLEVEL="${LDAP_LOGLEVEL:-256}"
-export LDAP_HASH_CLEARTEXT="${LDAP_HASH_CLEARTEXT:-no}"
 export LDAP_PASSWORD_HASH="${LDAP_PASSWORD_HASH:-{SSHA}}"
+export LDAP_CONFIGURE_PPOLICY="${LDAP_CONFIGURE_PPOLICY:-no}"
+export LDAP_PPOLICY_USE_LOCKOUT="${LDAP_PPOLICY_USE_LOCKOUT:-no}"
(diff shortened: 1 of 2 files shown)


carrodher:

Horstexplorer:
Thanks for the hint! I hope that the signature is now set correctly.

bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
```diff
@@ -402,6 +407,11 @@ dn: cn=config
 changetype: modify
 add: olcDisallows
 olcDisallows: bind_anon
+
+dn: cn=config
+changetype: modify
+add: olcRequires
+olcRequires: authc
```
juan131, line 414: Why is this change necessary?

bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
```diff
@@ -680,3 +701,109 @@ EOF
...
+        info "Enabling ppolicy_hash_cleartext"
+        cat > "${LDAP_SHARE_DIR}/ppolicy_configuration_hash_cleartext.ldif" << EOF
+dn: olcOverlay={0}ppolicy,olcDatabase={2}mdb,cn=config
+changetype: modify
+add: olcPPolicyHashCleartext
+olcPPolicyHashCleartext: TRUE
+EOF
+    debug_execute ldapmodify -Q -Y EXTERNAL -H "ldapi:///" -f "${LDAP_SHARE_DIR}/ppolicy_configuration_hash_cleartext.ldif"
+    fi
+  # enable ppolicy_use_lockout
```
juan131, line 778: Please indent this comment

bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
```diff
@@ -680,3 +701,109 @@ EOF
...
+ldap_configure_password_hash() {
+    info "Configuring LDAP olcPasswordHash"
+    cat > "${LDAP_SHARE_DIR}/password_hash.ldif" << EOF
+dn: olcDatabase={-1}frontend,cn=config
+changetype: modify
+add: olcPasswordHash
+olcPasswordHash: $LDAP_PASSWORD_HASH
+EOF
+    debug_execute ldapmodify -Y EXTERNAL -H "ldapi:///" -f "${LDAP_SHARE_DIR}/password_hash.ldif"
+}
```
juan131, line 809: Please address the "no newline at end of file" warning

bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
```diff
@@ -606,6 +613,20 @@ ldap_initialize() {
...
+        if is_boolean_yes "$LDAP_CONFIGURE_PPOLICY"; then
+            ldap_configure_ppolicy
+        fi
+        # enable tls
+        if is_boolean_yes "$LDAP_ENABLE_TLS"; then
+          ldap_configure_tls
+          if is_boolean_yes "$LDAP_REQUIRE_TLS"; then
+            ldap_configure_tls_required
+          fi
+        fi
```
juan131, line 629: Indentation in this block is different (it should be 4 spaces)

juan131 (changes requested):
Thanks so much for this awesome contribution! Overall, it looks great
I just added a few comments to address some "cosmetic" things.

bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
Horstexplorer: Although the wording of the README specifies that `LDAP_ALLOW_ANON_BINDING` targets the binding functionality, it seemed to me that this would be mainly used to prevent anonymous access to the directory. While I was configuring my instance I stumbled across [OpenLDAP Configuration Reference, 14.3.1](https://www.openldap.org/doc/admin26/guide.html#%22simple%22%20method) which suggests that simply disallowing bind_anon does not actually prevent anonymous access.
I am unaware under what circumstances this is the case or whether this applies to some data accessible without binding to an identity, but it felt to have the possibility of setting false security expectations to me.
However I could see reason for not including this into `LDAP_ALLOW_ANON_BINDING` but rather have it be a dedicated setting with a note for reference.

juan131 (approved):
LGTM

PR #45066, opened 2023-08-18
[bitnami/postgresql] README was not reflecting the start scripts

dhxgit:
### Description of the change

The start scripts use POSTGRESQL_INITDB_WAL_DIR instead of POSTGRESQL_INITDB_WALDIR.

This just fixes the README.md to reflect the code.

### Benefits

The README.md will no longer be wrong.

### Possible drawbacks

None

### Applicable issues

None (Could also file an issue, but this is a really small and obvious fix I believe)

### DIFF
--- a/bitnami/postgresql/README.md
+++ b/bitnami/postgresql/README.md
@@ -646,12 +646,12 @@ If you are using your custom `postgresql.conf`, you should create (or uncomment)
 Specifying extra initdb arguments can easily be done using the following environment variables:
 
 * `POSTGRESQL_INITDB_ARGS`: Specifies extra arguments for the initdb command. No defaults.
-* `POSTGRESQL_INITDB_WALDIR`: Defines a custom location for the transaction log. No defaults.
+* `POSTGRESQL_INITDB_WAL_DIR`: Defines a custom location for the transaction log. No defaults.
 
 ```console
 docker run --name postgresql \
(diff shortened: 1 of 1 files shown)


Mauraza:
Hi @dhxgit,

Could you be so kind as to tell us why this change?  Could you add some documentation?

dhxgit:
> Hi @dhxgit,
> 
> Could you be so kind as to tell us why this change? Could you add some documentation?

hi, as stated this fixes the documentation(readme), which at the moment does not reflect the underlying code.

atm the documentation says to set the env var named: POSTGRESQL_INITDB_WALDIR

This is wrong as the startup script of the container is checking for POSTGRESQL_INITDB_WAL_DIR.

if you would setup the container according to the readme, you would not be able to set the waldir properly.

see startup script in question here:

https://github.com/bitnami/containers/blob/main/bitnami/postgresql/15/debian-11/rootfs/opt/bitnami/scripts/postgresql-env.sh#L215

I stumbled over this because i was wondering why setting the wal_dir did not work, when i tried to set it up according to the readme.

I hope this clarifies the problem.

fixing it in the readme is the most sensible approach, as fixing the code to make the readme correct would be a bigger change, which would break existing configuration.

Mauraza (approved):
Hi @dhxgit,

You are right, thanks for the contribution!!
LGTM!

PR #46901, opened 2023-09-01
[bitnami/kafka] change zookeeper keystore pem location

toondaey:
### Description of the change

I think this necessary because from experience, I find that I keep having to update my local keystore which somehow became more of a burden

### Benefits

Eliminate the need to update and change zookeeper.keystore.pem every single time.

### Possible drawbacks

None that I can think of.

### Applicable issues

Frustrating need to keep updating local zookeeper.keystore.pem evicted.

### DIFF
--- a/bitnami/kafka/3.2/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
+++ b/bitnami/kafka/3.2/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
@@ -773,7 +773,7 @@ kafka_zookeeper_configure_tls() {
         # Concatenating private key into public certificate file
         # This is needed to load keystore from location using PEM
         keystore_location="${KAFKA_CERTS_DIR}/zookeeper.keypair.pem"
-        cat "${KAFKA_CERTS_DIR}/zookeeper.keystore.pem" "${KAFKA_CERTS_DIR}/zookeeper.keystore.key" > "$keystore_location"
+        cat "${KAFKA_CERTS_DIR}/zookeeper.keystore.{pem,key}" > "$keystore_location"
     fi
 
     kafka_server_conf_set "zookeeper.clientCnxnSocket" "org.apache.zookeeper.ClientCnxnSocketNetty"
--- a/bitnami/kafka/3.2/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
+++ b/bitnami/kafka/3.2/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
@@ -772,8 +772,8 @@ kafka_zookeeper_configure_tls() {
     elif [[ "$KAFKA_ZOOKEEPER_TLS_TYPE" = "PEM" ]] && [[ -f "$KAFKA_CERTS_DIR"/zookeeper.keystore.pem ]] && [[ -f "$KAFKA_CERTS_DIR"/zookeeper.keystore.key ]]; then
         # Concatenating private key into public certificate file
         # This is needed to load keystore from location using PEM
-        cat "$KAFKA_CERTS_DIR"/zookeeper.keystore.pem "$KAFKA_CERTS_DIR"/zookeeper.keystore.key > "$KAFKA_CERTS_DIR"/zookeeper-keypair.pem
-        keystore_location="$KAFKA_CERTS_DIR"/zookeeper-keypair.pem
+        keystore_location="${KAFKA_CERTS_DIR}/zookeeper.keypair.pem"
+        cat "${KAFKA_CERTS_DIR}/zookeeper.keystore.pem" "${KAFKA_CERTS_DIR}/zookeeper.keystore.key" > "$keystore_location"
     fi
 
     kafka_server_conf_set "zookeeper.clientCnxnSocket" "org.apache.zookeeper.ClientCnxnSocketNetty"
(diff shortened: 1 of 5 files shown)


bitnami/kafka/3.2/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
```diff
@@ -772,8 +772,8 @@ kafka_zookeeper_configure_tls() {
     elif [[ "$KAFKA_ZOOKEEPER_TLS_TYPE" = "PEM" ]] && [[ -f "$KAFKA_CERTS_DIR"/zookeeper.keystore.pem ]] && [[ -f "$KAFKA_CERTS_DIR"/zookeeper.keystore.key ]]; then
         # Concatenating private key into public certificate file
         # This is needed to load keystore from location using PEM
-        cat "$KAFKA_CERTS_DIR"/zookeeper.keystore.key >>"$KAFKA_CERTS_DIR"/zookeeper.keystore.pem
-        keystore_location="${KAFKA_CERTS_DIR}/zookeeper.keystore.pem"
+        cat "$KAFKA_CERTS_DIR"/zookeeper.keystore.pem "$KAFKA_CERTS_DIR"/zookeeper.keystore.key > "$KAFKA_CERTS_DIR"/zookeeper-keypair.pem
+        keystore_location="$KAFKA_CERTS_DIR"/zookeeper-keypair.pem
```
andresbono, line 776: What do you think about this minor suggestion?

```suggestion
        keystore_location="${KAFKA_CERTS_DIR}/zookeeper.keypair.pem"
        cat "${KAFKA_CERTS_DIR}/zookeeper.keystore.pem" "${KAFKA_CERTS_DIR}/zookeeper.keystore.key" > "$keystore_location"
```

Could you apply it to the rest of the files?

andresbono (commented):
LGTM, thank you!!  The current implementation would also be a problem if the file is mounted as a read-only secret.

 I just added a minor suggestion about the style. Could you please check it?

toondaey:
> LGTM, thank you!! The current implementation would also be a problem if the file is mounted as a read-only secret.
> 
> I just added a minor suggestion about the style. Could you please check it?

Actually that might not because I've you can use `cat` with a readonly mount into container files. In my local docker setup, I mounted all the other certificates as `ro` save for those of zookeeper because of the modifications.

I've applied and made some changes to the modifications.

bitnami/kafka/3.2/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
```diff
@@ -772,8 +772,8 @@ kafka_zookeeper_configure_tls() {
     elif [[ "$KAFKA_ZOOKEEPER_TLS_TYPE" = "PEM" ]] && [[ -f "$KAFKA_CERTS_DIR"/zookeeper.keystore.pem ]] && [[ -f "$KAFKA_CERTS_DIR"/zookeeper.keystore.key ]]; then
         # Concatenating private key into public certificate file
         # This is needed to load keystore from location using PEM
-        cat "$KAFKA_CERTS_DIR"/zookeeper.keystore.key >>"$KAFKA_CERTS_DIR"/zookeeper.keystore.pem
-        keystore_location="${KAFKA_CERTS_DIR}/zookeeper.keystore.pem"
+        keystore_location="${KAFKA_CERTS_DIR}/zookeeper.keypair.pem"
+        cat "${KAFKA_CERTS_DIR}/zookeeper.keystore.{pem,key}" > "$keystore_location"
```
andresbono, line 776: Unfortunately, [brace expansion](https://www.gnu.org/software/bash/manual/html_node/Brace-Expansion.html) will not work in a quoted string.

In the repository, we tend to favor the style of variable interpolation with curly braces and quoting the whole string, for example `"${var}/foo/bar"`. Although you can find exceptions. In this case, following this style convention would prevent brace expansion from working. So we need:

1)
```shell
cat "$KAFKA_CERTS_DIR"/zookeeper.keystore.{pem,key} > "$keystore_location"
```

or

2)
```shell
cat "${KAFKA_CERTS_DIR}/zookeeper.keystore.pem" "${KAFKA_CERTS_DIR}/zookeeper.keystore.key" > "$keystore_location"
```

I would go with the second option to follow the common style convention, but option one will also work.

andresbono (commented):
> > LGTM, thank you!! The current implementation would also be a problem if the file is mounted as a read-only secret.
> > I just added a minor suggestion about the style. Could you please check it?
> 
> Actually that might not because I've you can use `cat` with a readonly mount into container files. In my local docker setup, I mounted all the other certificates as `ro` save for those of zookeeper because of the modifications.

By "current implementation" I meant the implementation before your PR. I was just highlighting that `zookeeper.keystore.pem` could not be mounted as `ro` because it needed to be writable. With your PR you are also fixing that. Is that correct?

Please, check my last in-line comment. Thank you!

bitnami/kafka/3.2/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
toondaey: Yes, this should fix the current implementation. Apologies for not checking for the common style. I'll stick with the second one as well.

toondaey:
I hope this is ok? I discovered that `KAFKA_KRAFT_CLUSTER_ID` had a typo [here](https://github.com/bitnami/containers/blob/896caff5aedf91ce379effe71ac3b2ca5e9f655f/bitnami/kafka/README.md?plain=1#L210)

bitnami/kafka/3.2/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
andresbono: No problem at all!! Thank you for adding the suggestion.

andresbono (approved):
LGTM! Thank you!

PR #47340, opened 2023-09-04
[bitnami/opensearch] Fix readiness/liveness probe not working if "restEncryption: false" in values.yaml

### LINKED ISSUE #47078: [bitnami/elasticsearch] readiness/liveness probe not working if "restEncryption: false" in values.yaml
### Name and Version

bitnami/elasticsearch:8

amd64

`helm upgrade --install --wait --wait-for-jobs --values values.yaml --version 19.10.8 $release_name elasticsearch/elasticsearch`

values.yaml:
```yaml
security:
  enabled: true
  elasticPassword: "{{ service.attrs_elasticsearch.password }}"
  tls:
    autoGenerated: true
    restEncryption: false
    verificationMode: none
```

The readiness and liveness probes should work correctly regardless of the "restEncryption" setting. When "restEncryption" is set to "false," Elasticsearch should still be responsive to these probes, indicating its health status.

### What do you see instead?

Currently, when "restEncryption" is set to "false," the readiness and liveness probes appear to always fail. This behavior is inconsistent with the expected functionality.

### Additional information

The bug lies here: https://github.com/bitnami/containers/blob/cf811facba3bb35964a72fe81af0937b66920ab5/bitnami/elasticsearch/8/debian-11/rootfs/opt/bitnami/scripts/libelasticsearch.sh#L913C1-L914C1

since the condition at L913 will fail and that way the curl command would not report basic auth, returning 401 instead of 2xx

rabbagliettiandrea:
### Description of the change

The readiness and liveness probes should work correctly regardless of the "restEncryption" setting. When "restEncryption" is set to "false," Opensearch should still be responsive to these probes, indicating its green health status.

### Benefits

This fix the failing health-check (that is preventing opensearch startup if security is on and tlsEncryption is off).
Without the fix, the health-check is failing because misses the basic-auth

### Possible drawbacks

Nothing

### Applicable issues

- fixes #47078

### DIFF
--- a/bitnami/opensearch/1/debian-11/rootfs/opt/bitnami/scripts/libopensearch.sh
+++ b/bitnami/opensearch/1/debian-11/rootfs/opt/bitnami/scripts/libopensearch.sh
@@ -940,7 +940,8 @@ elasticsearch_healthcheck() {
 
     host=$(get_elasticsearch_hostname)
 
-    is_boolean_yes "$DB_ENABLE_SECURITY" && is_boolean_yes "$DB_ENABLE_REST_TLS" && protocol="https" && command_args+=("-k" "--user" "${DB_USERNAME}:${DB_PASSWORD}")
+    is_boolean_yes "$DB_ENABLE_SECURITY" && command_args+=("-k" "--user" "${DB_USERNAME}:${DB_PASSWORD}")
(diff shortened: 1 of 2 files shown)


dgomezleon (approved):

Issue #48237, opened 2023-09-12
Add libpq-dev python3-dev gcc to images


oktayalizada:
### Name and Version

bitnami/airflow-worker

### What is the problem this feature will solve?

psycopg2 could be installed from requirements.txt. Currently, if we try to install it in airflow as a dependency we will get: 
 ```
python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [23 lines of output]
      running egg_info
      creating /tmp/pip-pip-egg-info-24grtjxt/psycopg2.egg-info
      writing /tmp/pip-pip-egg-info-24grtjxt/psycopg2.egg-info/PKG-INFO
      writing dependency_links to /tmp/pip-pip-egg-info-24grtjxt/psycopg2.egg-info/dependency_links.txt
      writing top-level names to /tmp/pip-pip-egg-info-24grtjxt/psycopg2.egg-info/top_level.txt
      writing manifest file '/tmp/pip-pip-egg-info-24grtjxt/psycopg2.egg-info/SOURCES.txt'
      
      Error: pg_config executable not found.
      
      pg_config is required to build psycopg2 from source.  Please add the directory
      containing pg_config to the $PATH or specify the full executable path with the
      option:
      
          python setup.py build_ext --pg-config /path/to/pg_config build ...
```

### What is the feature you are proposing to solve the problem?

It will solve the problem of adding psycopg2 as a dependency in airflow

carrodher:
Thank you for opening this issue and submitting the associated Pull Request. Our team will review and provide feedback. Once the PR is merged, the issue will automatically close.

Your contribution is greatly appreciated!

PR #48258, opened 2023-09-12
[bitnami/airflow] Add libraries libpq-dev python-dev gcc to support installation psycopg2 as dependency

ghost:
### Description of the change

The change introduces three new packages which will be installed in airflow pods to support installation of psycopg2 library from requirements.txt file. 
Currently if we add psycopg2 library to requirements.txt we will get an error:
     Error: pg_config executable not found.

### Benefits

Possibility to use bitnami images and helm chart to install psycopg2 and other libraries which depend on it

### Applicable issues

- fixes #[48237](https://github.com/bitnami/containers/issues/48237)
- fixes #[19242](https://github.com/bitnami/charts/issues/19242)

### DIFF
--- a/.idea/workspace.xml
+++ b/.idea/workspace.xml
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="AutoImportSettings">
+    <option name="autoReloadType" value="SELECTIVE" />
+  </component>
+  <component name="ChangeListManager">
+    <list default="true" id="f34d2b43-b088-4b2b-b62a-d496e022a3f7" name="Changes" comment="" />
+    <option name="SHOW_DIALOG" value="false" />
+    <option name="HIGHLIGHT_CONFLICTS" value="true" />
+    <option name="HIGHLIGHT_NON_ACTIVE_CHANGELIST" value="false" />
+    <option name="LAST_RESOLUTION" value="IGNORE" />
+  </component>
+  <component name="Git.Settings">
+    <option name="RECENT_GIT_ROOT_PATH" value="$PROJECT_DIR$" />
+  </component>
+  <component name="MarkdownSettingsMigration">
+    <option name="stateVersion" value="1" />
+  </component>
+  <component name="ProjectColorInfo"><![CDATA[{
+  "associatedIndex": 5
+}]]></component>
+  <component name="ProjectId" id="2VIZDKu5gQtd7iBH615Ke59PnSm" />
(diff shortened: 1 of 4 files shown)


fevisera:
Hi @oktayalizada,

I have built the image locally, and it appears to be quite large:
```
local/airflow                      latest                        5bc6e03c8bf9   14 minutes ago   3.28GB 
bitnami/airflow                 latest                        8255c5199bb5   3 days ago       1.77GB
```
I believe the intention in https://github.com/bitnami/charts/issues/19242 was to create a fork rather than modify the bitnami airflow images. Adding these additional packages would need additional analysis, including an assessment of image size and the need for these packages. For instance, adding python-dev may include more packages than necessary. 

Please let me know if there are any additional details I may be overlooking.

oktayalizada:
Hi @fevisera,
So you mean airflow will not support installation and usage of psycopg2 library in airflow pods? Can we have maybe a different tag for this specific build (e.g. debian-11-psycopg2)?

fevisera:
Hi @oktayalizada,

Thank you for your question! I did not mean to imply that Airflow will not support the installation and usage of the psycopg2 library for bitnami/airflow. What I intended to convey is that before adding those new packages, it is essential to conduct a more thorough analysis to assess the image size impact and the actual necessity of these packages.

We already have an ongoing internal task to address adding dependencies feature, which has been created for https://github.com/bitnami/containers/pull/48258. As soon as this functionality becomes available, we will notify you here.

Thank you for proposing this change and bringing it to our attention. If you have any more questions or concerns, please feel free to ask. We appreciate your feedback.

alekseyolg:
Any opinion?
If necessary, I can open a new pull request to add the functionality needed in the work, taking into account the changes that have appeared since the creation of this pull request.
Thank you!

juan131:
Hi everyone!

These deps the PR is proposing to add (such as gcc) are required for compilation and they should never be included in the runtime image. We can use these libraries for compiling internally the different packages included in the container but they're useless at runtime.

IMHO this PR should be closed and an issue requesting `psycopg2` should be created instead. That said, we are already including the [psycopg2-binary](https://pypi.org/project/psycopg2-binary) pip package in the container image, see:

```console
$ docker run --rm -it bitnami/airflow bash
(...)
airflow@8275af36d3d5:/$ . /opt/bitnami/airflow/venv/bin/activate
(venv) airflow@8275af36d3d5:/$ pip list | grep psycopg2
psycopg2-binary                           2.9.10
```

PR #48505, opened 2023-09-14
[bitnami/mastodon] fix: correct mismatched variable name

backrunner:
`MASTODON_S3_ALIAS` is not matched the definition in `mastodon-env.sh` (the correct name is `MASTODON_S3_ALIAS_HOST`)

### Description of the change

Match the variable name in `mastodon-env.sh` and `libmastodon.sh`.

### Benefits

Correct the variable name to get rid of the `unbound value` issue.

### Applicable issues

If set `S3_ENABLED` to `true`, because of the mismatched variable name, there's an `unbound value` error.


joancafom (approved):
LGTM! Thanks for the change 🚀

PR #48957, opened 2023-09-18
[bitnami/kibana] Increase linked_libraries.timeout to 60s

migruiz4:
### Description of the change

Increases Kibana linked_libraries test timeout from 20s to 60s.

In my local tests, the linked_libraries test takes about 35 seconds to complete when using the latest 7.x version of Kibana.


andresbono (approved):
LGTM

PR #48962, opened 2023-09-18
[bitnami/harbor-notary-*] Add deprecated notice

Mauraza:
Deprecate notice in harbor-notary-server and harbor-notary-signer


carrodher:
It seems you are mixing two different deprecations: Envoy 1.23 and Harbor Notary

juan131 (approved):
LGTM

Issue #49157, opened 2023-09-19
[bitnami/openldap] ppolicy module should be enabled before adding custom ldifs or create tree


plomosits:
### Name and Version

bitnami/openldap:2.6.6

### What architecture are you using?

arm64

### What steps will reproduce the bug?

1. set environment variables to enable ppolicy
```yaml
env:
  LDAP_CONFIGURE_PPOLICY: 'yes'
  LDAP_PPOLICY_USE_LOCKOUT: 'yes'
  LDAP_PPOLICY_HASH_CLEARTEXT: 'yes'
```
2. create objects with passwords with customLdifFile or create users with passwords

### What is the expected behavior?

Enable ppolicy modules before adding custom LDIFs or building a tree so that this object's passwords are already hashed.

### What do you see instead?

The password of the created objects are in cleartext.

### Additional information

A possible solution could be moving the code, adding custom ldifs or creating a tree after all configurations.

`/opt/bitnami/scripts/libopenldap.sh`
```bash
       # enable tls
        if is_boolean_yes "$LDAP_ENABLE_TLS"; then
            ldap_configure_tls
            if is_boolean_yes "$LDAP_REQUIRE_TLS"; then
                ldap_configure_tls_required
            fi
        fi

        if ! is_dir_empty "$LDAP_CUSTOM_LDIF_DIR"; then
            ldap_add_custom_ldifs
        elif ! is_boolean_yes "$LDAP_SKIP_DEFAULT_TREE"; then
            ldap_create_tree
        else
            info "Skipping default schemas/tree structure"
        fi

        ldap_stop
```

[libopenldap.sh.txt](https://github.com/bitnami/containers/files/12660364/libopenldap.sh.txt)

carrodher:
Thank you for bringing this issue to our attention. We appreciate your involvement! If you're interested in contributing a solution, we welcome you to create a pull request. The Bitnami team is excited to review your submission and offer feedback. You can find the contributing guidelines [here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md).

Your contribution will greatly benefit the community. Feel free to reach out if you have any questions or need assistance.

carrodher:
Thank you for opening this issue and submitting the associated Pull Request. Our team will review and provide feedback. Once the PR is merged, the issue will automatically close.

Your contribution is greatly appreciated!

PR #49271, opened 2023-09-20
#49157 move adding custom ldifs or create tree to the end of the initial configuration section

plomosits:
### Description of the change

As described in #49157 adding ldifs or create tree should run after configuring ppolicy.

### Benefits

Created passwords in objects from the ldifs or the tree will be hashed correctly.


dgomezleon (approved):

PR #49583, opened 2023-09-22
Update OpenLDAP container readme for environment variables.

0xbobm:
### Applicable issues

- fixes #


fmulero (commented):
Thanks a lot @okbobm

Those values are the defaults for `LDAP_ROOT` and `LDAP_ADMIN_DN` so they shouldn't be required. Did you face any issue without them?

okbobm:
I'm suggesting adding those variables into the example even if they are the default (perhaps comment as such), as the docker commands will run successfully the first time however the second time with custom domain values (changing the corresponding values in the Maria docker command line), users will struggle to understand that the variables need to be changed on the docker command for the openldap server implementation.

fmulero (approved):
LGTM

PR #50118, opened 2023-09-27
[bitnami/rclone] Add fuse as dep

Daviey:
rclone supports mounting using fuse, but depends on the fusermount3 command, which is provided by the package fuse3.

More info here:
https://rclone.org/commands/rclone_mount/


Daviey:
@FraPazGal Are you able to review this please?  Thanks

FraPazGal:
Just to give an update on this, it seems the `fuse3` pkg is not available for Ubuntu 18. We are working on some internal changes that should be ready for the start of next week.

Daviey:
Is there something I can do to move this along? Thanks

FraPazGal (approved):
Thanks for the wait @Daviey, we have now solved some internal hiccups and can safely merge this change!

PR #50435, opened 2023-09-29
[bitnami/odoo] Add database filter

LuisValdesZero:
### Description of the change

The current [odoo.conf.tpl](https://github.com/bitnami/containers/blob/97a8d24dca0a93d1c5f23cb4fe332fbeb3615c76/bitnami/odoo/16/debian-11/rootfs/opt/bitnami/scripts/odoo/bitnami-templates/odoo.conf.tpl#L15C14-L15C14) template has by a default a wide range of values for database name.

```
dbfilter = .*
```

By adding the env var `ODOO_DATABASE_FILTER` it is possible to filter a specific regex for the database name.

### Benefits

It will allow to set a specific database name by using a regex

### Additional information

I made some changes on the bitnami helm chart for odoo, I added the [env var](https://github.com/kavichu/bitnami-charts/blob/ffd1aeab9127914c2d61173fcb861a65905c0096/bitnami/odoo/templates/deployment.yaml#L124C14-L124C14)
I was able to deploy this change and it works as expected.
How to use with helm chart:

```
helm upgrade --install -f values.yml \
     --set odooEmail="odoo@example.com" \
     --set odooPassword="password" \
     --set odooSkipInstall="true" \
     --set odooDatabaseFilter="^business_database\$" \
     --namespace default \
     odoo-aratiri ./bitnami-charts/bitnami/odoo
```

### DIFF
--- a/bitnami/odoo/14/debian-11/rootfs/opt/bitnami/scripts/odoo-env.sh
+++ b/bitnami/odoo/14/debian-11/rootfs/opt/bitnami/scripts/odoo-env.sh
@@ -42,6 +42,7 @@ odoo_env_vars=(
     ODOO_DATABASE_NAME
     ODOO_DATABASE_USER
     ODOO_DATABASE_PASSWORD
+    ODOO_DATABASE_FILTER
     SMTP_HOST
     SMTP_PORT
     ODOO_SMTP_PORT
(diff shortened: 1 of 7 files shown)


jotamartos:
Thanks for taking the time to create this PR. Could you please update the README file so users are aware of this option? 

Thanks

jotamartos (approved):

jotamartos:
Thanks for your contribution. A new version of the containers will be available soon.

kavichu:
Hi @jotamartos, thanks guiding me on this.

PR #50838, opened 2023-10-03
[bitnami/wordpress] Add env var to disable XML-RPC

### LINKED ISSUE #50755: [bitnami/wordpress] Add env var to disable XML-RPC
### Name and Version

bitnami/wordpress:latest

Since XML-RPC is a known vulnerability and only kept for backward compatibility, you should be able to disable it straight away for new wordpress instances.

It would be great to have the option to set this directly as a env var, that then maybe will add a rule to the .htaccess file:
```
# Block WordPress xmlrpc.php requests
<Files xmlrpc.php>
order deny,allow
 deny from all
#allow from 123.123.123.123
</Files>
```

There are also alternative solutions like installing a plugin that blocks these requests or adding a filter in a child theme. But I think these solutions are pretty overkill.

kimdre:
### Description of the change

This PR implements a new environment variable to the (Apache) Wordpress image called `WORDPRESS_ENABLE_XML_RPC`, that lets you disable/block access to the Wordpress XML-RPC endpoint at `xmlrpc.php`, as this endpoint is deprecated in favor of the now used Wordpress Rest API and may have security vulnerabilities. 
The current best-practice is to disable access to it alltogether if you don't need it, so you should have the abillity to do so in my opinion.

To disable access to the XML-RPC endpoint you can set `WORDPRESS_ENABLE_XML_RPC=no`.
This will create a new htaccess rule when WordPress is initialized:
```
# Disable the oudated WordPress XML-RPC endpoint to prevent security vulnerabilities.
<Files xmlrpc.php>
Order Allow,Deny
Deny from all
</Files>
```

### Benefits

By blocking the XML-RPC endpoint you now have one thing less to worry about as a site administrator.
There are also some plugins that do the same, or you could use a filter in the child theme. However, for me, using the htaccess file is the best and safest approach.

### Possible drawbacks

Blocking the endpoint could cause problems for users who still use it for access through third-party software.
However, the endpoint has long been deprecated, so further use is already discouraged and it will likely be removed altogether somewhere in the future.

### Applicable issues

- fixes #50755


carrodher:

PR #51077, opened 2023-10-04
[bitnami/wordpress] Add env var to disable XML-RPC

### LINKED ISSUE #50755: [bitnami/wordpress] Add env var to disable XML-RPC

kimdre:
### Description of the change

This PR implements a new environment variable to the (Apache) Wordpress image called `WORDPRESS_ENABLE_XML_RPC`, that lets you disable/block access to the Wordpress XML-RPC endpoint at `xmlrpc.php`, as this endpoint is deprecated in favor of the now used Wordpress Rest API and may have security vulnerabilities. 
The current best-practice is to disable access to it alltogether if you don't need it, so you should have the abillity to do so in my opinion.

To disable access to the XML-RPC endpoint you can set `WORDPRESS_ENABLE_XML_RPC=no`.
This will create a new htaccess rule when WordPress is initialized:
```
# Disable the oudated WordPress XML-RPC endpoint to prevent security vulnerabilities.
<Files xmlrpc.php>
Order Allow,Deny
Deny from all
</Files>
```

### Benefits

By blocking the XML-RPC endpoint you now have one thing less to worry about as a site administrator.
There are also some plugins that do the same, or you could use a filter in the child theme. However, for me, using the htaccess file is the best and safest approach.

### Possible drawbacks

Blocking the endpoint could cause problems for users who still use it for access through third-party software.
However, the endpoint has long been deprecated, so further use is already discouraged and it will likely be removed altogether somewhere in the future.

### Applicable issues

- fixes #50755


bitnami/wordpress/6/debian-11/rootfs/opt/bitnami/scripts/libwordpress.sh
```diff
@@ -226,6 +227,12 @@ wordpress_initialize() {
                 WORDPRESS_DATA_TO_PERSIST+=" ${htaccess_file}"
             fi
         fi
+    else
+        if is_boolean_yes "$WORDPRESS_HTACCESS_OVERRIDE_NONE"; then
+            local htaccess_file="${APACHE_HTACCESS_DIR}/wordpress-htaccess.conf"
+        else
+            local htaccess_file="${WORDPRESS_BASE_DIR}/.htaccess"
+        fi
```
joancafom, line 235: To me, it is confusing to add this part of the code here just so that `htaccess_file` is defined with the correct value. I would rather compute the right value inside the `wordpress_disable_xmlrpc_endpoint` function itself.

bitnami/wordpress/6/debian-11/rootfs/opt/bitnami/scripts/libwordpress.sh
```diff
@@ -583,6 +591,29 @@ if ( !defined( 'WP_CLI' ) ) {
...
+########################
+# Disable access to the WordPress XML-RPC endpoint
+# Globals:
+#   *
+# Arguments:
+#   $1 - path to .htaccess file
+# Returns:
+#   None
+#########################
+wordpress_disable_xmlrpc_endpoint() {
```
joancafom, line 603: Nice! But you have to bear in mind that we also offer the option to use wordpress with nginx (see [bitnami/wordpress-nginx](https://github.com/bitnami/containers/tree/main/bitnami/wordpress-nginx)). The `libwordpress.sh` should have logic that can be used in both implementations, so you should also take that in mind.

joancafom (changes requested):
Thanks for your PR! I left you some comments as a review, could you please check them?

bitnami/wordpress/6/debian-11/rootfs/opt/bitnami/scripts/libwordpress.sh
kimdre: Ok, fixed this.

bitnami/wordpress/6/debian-11/rootfs/opt/bitnami/scripts/libwordpress.sh
kimdre: Well the issue with this is, that nginx doesn't support .htaccess files so the alternative would be to add a nginx config:
```nginx
location = /xmlrpc.php {
    deny all;
}
```

Would this be sufficient?

joancafom: Yes, I think it would be enough 😁

kimdre: Ok, wordpress-nginx now also has support for this optiom. 
I also moved the the code from .htaccess to the apache server config so that the code for the nginx and apache images have a similiar structure.

kimdre:
Now both the Apache and Nginx images support this option and use the same code. This also has the advantage that the option is no longer static in the .htaccess configuration, but can now be toggled by changing the variable value and a container restart.

kimdre:
@joancafom Could you please check the latest changes to my PR if you have time? :)

bitnami/wordpress-nginx/README.md
```diff
@@ -240,6 +240,7 @@ Available environment variables:
 - `WORDPRESS_SKIP_BOOTSTRAP`: Skip the WordPress installation wizard. This is necessary when providing a database with existing WordPress data. Default: **no**
 - `WORDPRESS_AUTO_UPDATE_LEVEL`: Level of auto-updates to allow for the WordPress core installation. Valid values: `major`, `minor`, `none`. Default: **none**
 - `WORDPRESS_ENABLE_REVERSE_PROXY`: Enable WordPress support for reverse proxy headers. Default: **no**
+- `WORDPRESS_ENABLE_XML_RPC`: Enable the WordPress XML-RPC endpoint. Default: **yes**
```
joancafom, line 243: I do believe that this point should be disabled by default. I would change the default value to `no` and add a note in the [Notable Changes](https://github.com/bitnami/containers/blob/main/bitnami/wordpress/README.md#notable-changes) section regarding this change.

If someone is using is using it, they can simply turn the feature back by tuning the value. WDYT?

joancafom (commented):
Thank you so much for your contribution, the approach you followed was brilliant! I added a small comment, could you please take a look at it when you have a chance? 😁

bitnami/wordpress-nginx/README.md
kimdre: Sure, this is fine for me. 👍

joancafom (approved):
LGTM! Thanks for your changes 🚀

PR #51111, opened 2023-10-04
[bitnami/mastodon] Fix initJob.createAdmin is true and the init container never finish

xueshanf:
See https://github.com/bitnami/charts/issues/14485.

If the admin user/email  is already exist, the current code exists with error instead of skipping account creation,  because the testing expects "been taken" string from tootctl command, however tootctl in  the 4.2.0 release that I tested, return different string:

```
res=$(tootctl accounts create "$MASTODON_ADMIN_USERNAME" --email "$MASTODON_ADMIN_EMAIL" "--confirmed" "--role" "Owner")
$ echo $res
Failure/Error: email taken Failure/Error: account.username taken
```

Simply change the testing string to "taken" fix this issue.

### Applicable issues

- fixes #


fevisera (approved):
LGTM! Thank so much for you contribution!

fevisera:
Hi @xueshanf,

I have just realized that issue had already been resolved by #50725 a couple of days prior to you creating this one. For that reason, I'm canceling this issue. Nevertheless, thank you for reporting it!

PR #51192, opened 2023-10-06
[bitnami/openldap] Correct naming and location for some organization units

### LINKED ISSUE #19716: LDAP groups should be in a different OU than users
### Name and Version

bitnami/openldap:2.6

Create an LDAP container with a user and group, why are groups in the same OU as users?

Groups should be under a different ou, not users or people right? E.g.
https://serverfault.com/questions/87079/openldap-posix-groups-account
```
# User primary group
dn: cn=ussergroup,ou=groups,dc=me,dc=com
...
# User account
dn: uid=user,ou=users,dc=me,dc=com
```
Please add a LDAP_GROUP_OU similar to LDAP_USER_DC and use that in the template.
Btw LDAP_USER_DC should probably be renamed to "LDAP_USER_OU", since it defaults to ou=users not a dc?

### What do you see instead?

`dn: cn=mygroup,ou=users,dc=example,dc=org`

### Additional information

this originally came up in https://github.com/bitnami/bitnami-docker-openldap/issues/2

barsikus007:
### Description of the change

Fix of #19716 issue

### Benefits

Correct and predictable location of groups organization unit (E.g. https://hub.docker.com/r/wheelybird/ldap-user-manager image expect to see groups in separate ou than users)

### Possible drawbacks

It could break some configs, which is targeted to previous schema and env variables

### Applicable issues

- fixes #19716


barsikus007:
hey?

carrodher:
Hi, please note this PR is marked as Draft, if it is ready to be reviewed, could you please mark it as _Ready for review_?
Screenshot 2023-10-23 at 08 16 42

barsikus007:
Oh I get it

migruiz4:
Hi @barsikus007,

Thank you very much for your contribution, but I'm worried about how replacing `LDAP_USER_DC` with `LDAP_USER_OU/LDAP_GROUP_OU` would affect existing users.

Do you think you could make this change backward-compatible? For example, `LDAP_USER_DC` could be tagged as deprecated, but if provided legacy behavior will be used and both `LDAP_USER_OU/LDAP_GROUP_OU` will be set to `LDAP_USER_OU="${LDAP_USER_DC} and LDAP_GROUP_OU="${LDAP_USER_DC}"` and current behavior will be used.
 
That way we could warn users and safely remove it in a future release.

migruiz4 (commented):
Hi @barsikus007,

Thank you very much for your contribution, but I'm worried about how this change would affect existing users.

Do you think you could make this change backward-compatible? For example, instead of removing it, you could set `LDAP_USER_OU/LDAP_GROUP_OU` to `LDAP_USER_DC` if it was set and warn users about that environment variable being deprecated in favor of the new ones.

That way, the image functionality won't be affected and users will have time to modify their deployments, so we can safely remove the deprecated setting in a future release.

barsikus007:
Okay, but how to notify users about changes?

bitnami/openldap/2.5/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
```diff
@@ -65,7 +65,8 @@ export LDAP_EXTRA_SCHEMAS="${LDAP_EXTRA_SCHEMAS:-cosine,inetorgperson,nis}"
 export LDAP_SKIP_DEFAULT_TREE="${LDAP_SKIP_DEFAULT_TREE:-no}"
 export LDAP_USERS="${LDAP_USERS:-user01,user02}"
 export LDAP_PASSWORDS="${LDAP_PASSWORDS:-bitnami1,bitnami2}"
-export LDAP_USER_DC="${LDAP_USER_DC:-users}"
+export LDAP_USER_OU="${LDAP_USER_OU:-users}"
+export LDAP_GROUP_OU="${LDAP_GROUP_OU:-groups}"
```
migruiz4, line 69: As in my suggestion, please keep the value `LDAP_USER_DC` but remove its default value.

Then, for `LDAP_USER_OU` and `LDAP_GROUP_OU` it will have the following priorities:
1. Value provided in `LDAP_USER_OU`/`LDAP_GROUP_OU`
2. If no value provided and `LDAP_USER_DC`, use value in `LDAP_USER_DC`.
3. If neither `LDAP_USER_OU`/`LDAP_GROUP_OU` or `LDAP_USER_DC` provided, then use default values `user/group`.

```suggestion
export LDAP_USER_DC="${LDAP_USER_DC:-}"
export LDAP_USER_OU="${LDAP_USER_OU:-${LDAP_USER_DC:-users}}"
export LDAP_GROUP_OU="${LDAP_GROUP_OU:-${LDAP_USER_DC:-groups}}"
```

Then, around line 174, inside `ldap_validate` add the following message:
```
if [[ -n "$LDAP_USER_DC" ]]; then
    warn "The env variable 'LDAP_USER_DC' has been deprecated and will be removed in a future release. Please use 'LDAP_USER_OU' and 'LDAP_GROUP_OU' instead."
fi
```

bitnami/openldap/README.md
```diff
@@ -186,12 +186,13 @@ The Bitnami Docker OpenLDAP can be easily setup with the following environment v
 * `LDAP_CONFIG_ADMIN_PASSWORD_FILE`: Path to a file that contains the LDAP configuration admin user password. This will override the value specified in `LDAP_CONFIG_ADMIN_PASSWORD`. No defaults.
 * `LDAP_USERS`: Comma separated list of LDAP users to create in the default LDAP tree. Default: **user01,user02**
 * `LDAP_PASSWORDS`: Comma separated list of passwords to use for LDAP users. Default: **bitnami1,bitnami2**
-* `LDAP_USER_DC`: DC for the users' organizational unit. Default: **users**
+* `LDAP_USER_OU`: Name for the user's organizational unit. Default: **users**
+* `LDAP_GROUP_OU`: Name for the group's organizational unit. Default: **groups**
```
migruiz4, line 190: Finally, in the README.md, you could also warn about the deprecation:

```
* `LDAP_USER_OU`: Name for the user's organizational unit. Default: **users**
* `LDAP_GROUP_OU`: Name for the group's organizational unit. Default: **groups**
* `LDAP_USER_DC`: DC for the users' organizational unit. **DEPRECATED** Please use `LDAP_USER_OU` and `LDAP_GROUP_OU` instead.
```

migruiz4 (commented):
Please take a look at my suggestions

PR #51280, opened 2023-10-07
[bitnami/gitea] permit to disable users registration

eltorio:
### Description of the change

On Gitea setting `DISABLE_REGISTRATION` in `app.ini` permits to disable users to self register in Gitea after which only admin can create accounts for users.  
This can be easily controlled in Docker with this PR by setting this environment variable. 
* `GITEA_DISABLE_REGISTRATION`: "true"|"false"

### Benefits

Control with environment variable
```toml
[service]
DISABLE_REGISTRATION = true|false
```

### Applicable issues

- fixes #


amalgamm:
Actually this changes are quite useful. Could you merge it?

PR #51282, opened 2023-10-09
[bitnami/*] Do not execute tests for automatic PRs

joancafom:
### Description of the change

At this moment, we are testing changes coming from automated PR twice (once in our internal pipeline and another here in GitHub). This PR introduces changes so that automated PRs are directly merged.

### Benefits

- Automated PRs will be available sooner
- Reduction of computation efforts
- Reduction of transient errors

### Possible drawbacks

None. PRs have been previously verified before they are sent out to GitHub.

### Applicable issues

NA

### Additional information

NA


carrodher (dismissed):

fmulero (dismissed):
Thanks @joancafom

My first idea would have been to add the condition in the get-containers step (and probably change that name) as we currently do for the READMEs, but this approach seems clearer. LGTM

FraPazGal (commented):
I would personally separate the `auto-pr-review` job to a new workflow altogether. Now that it is going to be independent from the verification process (the proper CI), the automatic PR review could be in a different workflow.

FraPazGal (approved):
LGTM!

PR #51499, opened 2023-10-10
[bitnami/symfony] Remove webapp-pack package test

mdhont:
### Description of the change

We are removing the webapp-pack package as it is not included by default and is incompatible with the default installed packages.


dgomezleon (approved):
LGTM

PR #51504, opened 2023-10-10
[bitnami/*] Fix race condition in CI automated pipelines

joancafom:
### Description of the change

Unifies `auto-pr-review.yml` and `license-headers.yml` in the `ci-pipeline.yml` workflow, which solves some race condition problems introduced in #51282

### Benefits

- The auto-approval and merge of an automated PR will only happen when the license header check has run and passed.

### Possible drawbacks

NA

### Applicable issues

NA

### Additional information

NA


FraPazGal (approved):
LGTM!

fmulero (commented):
Be careful with these changes, `license-headers-linter` job is used by the branch protection rules. Please check if those rules are OK after applying the changes.

PR #51506, opened 2023-10-10
[bitnami/ksql] Update goss test for ksql

CeliaGMqrz:
### Description of the change

Update the goss test according to the following branch for KSQL 7.5


FraPazGal (approved):
LGTM!

PR #51862, opened 2023-10-13
fix: Fix typo of priviledged port

CrBoy:
### Description of the change

tcp/1389 is not a privileged port in general.

### Benefits

The descriptions won't confuse users.


dgomezleon (approved):

PR #51872, opened 2023-10-13
[bitnami/postgresql] Enable override the archive_command value in postgresql containers

### LINKED ISSUE #51871: [bitnami/postgresql*] Not posible to define archive_command for incremental backup
### Name and Version

bitnami/postgresql-repmgr

Name and Version
bitnami/postgresql-repmgr

What is the problem this feature will solve?
Right now we have replication but we do not have the ability to define an additional incremental backup in a Volume of choice It will be good either to be able to either:

- provide to the pg_ctl a -c archive_command=' cp % /var/lib/postgresql/backup/${MY_POD_NAME}

- Generate the archive_command value in postgresql.conf based on an environment variable

An additional variable should be added to the environment of the container

During startup when the file is configured it could be placed in the postgresql.conf

I tried to use configmaps for the postgresql.conf, but it won't work as each node needs to have different primary_conninfo value.
At the same time there is no easy way to add this into the code.

I considered using extraArgs, but those are not args to the pg_ctl, but to the command you override it with.

I am going to try to use a initcontainer or a sidecar to modify the postgresql.conf, but it needs to be done before the postgresql starts so it is doing again all the job already done when the file is created by the image.

Alvaro-Campesino:
### Description of the change

This changes enables overriding the value of  `archive_command` int the `postgresql.conf`  when `POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR` is defined as an environment variable

*Default value*: `/bin/true`
*Value when overrided*:  `cp %p ${POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR}/%f`

### Benefits

Now it is possible to store incremental backups in an additional volume. Increasing restore capabilities.

### Applicable issues

- fixes #51871


javsalgar:
Hi!

bitnami/postgresql-repmgr/16/debian-11/rootfs/opt/bitnami/scripts/libpostgresql.sh
```diff
@@ -666,6 +680,7 @@ postgresql_initialize() {
     is_boolean_yes "$create_pghba_file" && is_boolean_yes "$POSTGRESQL_ENABLE_TLS" && [[ -n $POSTGRESQL_TLS_CA_FILE ]] && postgresql_tls_auth_configuration
 
     is_boolean_yes "$create_conf_file" && [[ -n "$POSTGRESQL_SHARED_PRELOAD_LIBRARIES" ]] && postgresql_set_property "shared_preload_libraries" "$POSTGRESQL_SHARED_PRELOAD_LIBRARIES"
+    is_boolean_yes "$create_conf_file" && postgresql_configure_archive_command
```
fmulero, line 683: Is there any reason why this change is not in the rest of files?

bitnami/postgresql-repmgr/11/debian-11/rootfs/opt/bitnami/scripts/libpostgresql.sh
```diff
@@ -619,10 +632,12 @@ postgresql_initialize() {
     chmod go-rwx "$POSTGRESQL_DATA_DIR" || warn "Lack of permissions on data directory!"
 
     is_boolean_yes "$POSTGRESQL_ALLOW_REMOTE_CONNECTIONS" && is_boolean_yes "$create_pghba_file" && postgresql_create_pghba && postgresql_allow_local_connection
-    # Configure port
+        # Configure port
```
fmulero, line 635:
```suggestion
    # Configure port
```

bitnami/postgresql-repmgr/11/debian-11/rootfs/opt/bitnami/scripts/libpostgresql.sh
```diff
@@ -619,10 +632,12 @@ postgresql_initialize() {
...
     postgresql_set_property "port" "$POSTGRESQL_PORT_NUMBER"
     is_empty_value "$POSTGRESQL_DEFAULT_TOAST_COMPRESSION" || postgresql_set_property "default_toast_compression" "$POSTGRESQL_DEFAULT_TOAST_COMPRESSION"
     is_empty_value "$POSTGRESQL_PASSWORD_ENCRYPTION" || postgresql_set_property "password_encryption" "$POSTGRESQL_PASSWORD_ENCRYPTION"
+    # Configure WAL backup with archive_command
+    is_empty_value "$POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR" || postgresql_configure_archive_command
```
fmulero, line 640: The variable `POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR` is not defined in `postgresql-env.sh` Could you add it to that file?

bitnami/postgresql-repmgr/11/debian-11/rootfs/opt/bitnami/scripts/libpostgresql.sh
```diff
@@ -368,6 +368,19 @@ postgresql_create_replication_user() {
...
+# Globals:
+#   POSTGRESQL_*
+# Arguments:
+#   None
+# Returns:
+#   None
+#########################
+postgresql_configure_archive_command() {
+  info "Overriding archive_command default value..."
+  postgresql_set_property "archive_command" "cp %p ${POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR}/%f"
```
fmulero, line 382: Should the `POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR` be persisted? Should we place it under `/bitnami/postgresql` ?

fmulero (changes requested):
Thanks a lot Alvaro for your contribution.

As general comment, could you apply your changes in just one asset (postgresql for example)? That will help us to test your changes, don't worry about the other assets, once the PR is approved, we will apply the changes to all of them.

I also left some comments, could you take them a glance?

bitnami/postgresql-repmgr/11/debian-11/rootfs/opt/bitnami/scripts/libpostgresql.sh
fmulero: In that case we should ensure that the directory exists and has the right permissions. Please take a look to this piece of code: https://github.com/bitnami/containers/blob/934b1fa547a3bf4cf506cb48a185dd236def47b3/bitnami/postgresql-repmgr/11/debian-11/rootfs/opt/bitnami/scripts/libpostgresql.sh#L625-L629

NOTE: In your case the directory can be empty

fmulero (changes requested):
Thanks a lot @Alvaro-Campesino 

Please implement the changes in only one container, use `postgresql` container for example and remove changes in `postgresql-repmgr` and `supabase-postgres` . That will help us to review and test the changes. Please also fix the DCO, we can't not merge changes if they are not properly signed.

Also I left a little comment, please take it a look.

fmulero (changes requested):
Hi @Alvaro-Campesino 

Thanks a lot for applying the changes only to postgresql container.

About my comment [here](https://github.com/bitnami/containers/pull/51872#discussion_r1383081150) I was wrong, if we use a separate volume to store the files, is not needed to create the folders and permissions. Sorry to point you in the wrong direction. Maybe adding a validation to ensure the directory exists would be a good idea in this case.

bitnami/postgresql/16/debian-11/rootfs/opt/bitnami/scripts/libpostgresql.sh
```diff
@@ -614,15 +627,19 @@ postgresql_initialize() {
         ensure_dir_exists "$dir"
         am_i_root && chown "$POSTGRESQL_DAEMON_USER:$POSTGRESQL_DAEMON_GROUP" "$dir"
     done
+    is_empty_value "$POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR" || ensure_dir_exists "$POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR" && am_i_root && chown "$POSTGRESQL_DAEMON_USER:$POSTGRESQL_DAEMON_GROUP" "$POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR"
     am_i_root && find "$POSTGRESQL_DATA_DIR" -mindepth 1 -maxdepth 1 -not -name ".snapshot" -not -name "lost+found" -exec chown -R "$POSTGRESQL_DAEMON_USER:$POSTGRESQL_DAEMON_GROUP" {} \;
     chmod u+rwx "$POSTGRESQL_DATA_DIR" || warn "Lack of permissions on data directory!"
     chmod go-rwx "$POSTGRESQL_DATA_DIR" || warn "Lack of permissions on data directory!"
+    is_empty_value "$POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR" || chmod u+rw "$POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR"
```
fmulero, line 634: Due to the issue mentioned above, something like this could be worth.

```suggestion
    is_empty_value "$POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR" || ! is_file_writable "$POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR" || chmod u+rw "$POSTGRESQL_WAL_ARCHIVE_COMMAND_DIR"

```

fmulero (commented):
Hi @Alvaro-Campesino, I left some comments

Please take them a glance and fix DCO.

Regards

PR #51960, opened 2023-10-18
[bitnami/pgbouncer] Support the max_prepared_statements variable in pgbouncer

davisg123:
### Description of the change

PGBouncer version 1.21.0 has added a new [configuration parameter](https://www.pgbouncer.org/config.html#max_prepared_statements) `max_prepared_statements`. This change allows the parameter to be set via an environment variable

### Benefits

The change allows users to use prepared statements with PGBouncer by exposing the configuration setting


jotamartos (approved):
Changes LGTM! A new version of the container should be available soon with these changes

PR #52049, opened 2023-10-23
[bitnami/appsmith] Update test of rts

Mauraza:
### Description of the change

In the next version, server.js is in `/opt/appsmith/rts/bundle` and doesn't need the required


mdhont (approved):
Lgtm!

PR #52131, opened 2023-10-25
Remove Build Maintenance column and its automation

carrodher:
After changes at https://github.com/bitnami/containers/pull/51282 the revision of automated PRs was removed so changes are directly merged in `main`. Because of that, the _Build Maintenance_ column is not used anymore in containers


fmulero (changes requested):
Thanks a lot @carrodher 

I think you have to remove also this piece of code in the [comments handler](https://github.com/bitnami/containers/blob/3b50e17ecbeb756429b3762e709065785255dcd5/.github/workflows/comments.yml#L26-L34)

fmulero (approved):
LGTM

PR #52210, opened 2023-10-27
[bitnami/moodle] Fix moodle mount points in README

### LINKED ISSUE #19667: [bitnami/moodle] Missing files after Moodle migration: Failed to open stream: No such file or directory in /bitnami/moodle/lib/filestorage/file_system.php on line 263
### Name and Version

bitnami/moodle 4.1.2

None

1. Set up moodle in kubernetes for one domain (without certificate)
2. Change domain names and include a certificate
3. Restart pods
4. All the uploaded files are missing.

_No response_

The expected behavior is that after migration to a new domain everything works as it did on the old domain.

### What do you see instead?

On the new domain, all uploaded files are missing (file not found, 404).

### Additional information

The  only difference between the two instances is that the development version (original) did not have a Let's Encrypt certificate, while the production version (after migration) does have a Let's Encrypt certificate.

rafariossaa:
### Description of the change

Updates the volumes mount points in the README file

### Applicable issues

- fixes [chart #19667](https://github.com/bitnami/charts/issues/19667)

### DIFF
--- a/bitnami/moodle/README.md
+++ b/bitnami/moodle/README.md
@@ -112,6 +112,7 @@ docker run -d --name moodle \
   --env MOODLE_DATABASE_NAME=bitnami_moodle \
   --network moodle-network \
   --volume moodle_data:/bitnami/moodle \
+  --volume moodledata_data:/bitnami/moodledata
   bitnami/moodle:latest
 ```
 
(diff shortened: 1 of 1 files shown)


dgomezleon (approved):
LGTM

PR #52229, opened 2023-10-30
[bitnami/pgbouncer] Document how to connect PgBouncer with N PostgreSQL

### LINKED ISSUE #20874: [bitnami/pgbouncer] Allow multiple DB connections

gongomgra:
### Description of the change

Documents how to connect PgBouncer with N PostgreSQL databases as discussed in [ticket #20874](https://github.com/bitnami/containers/issues/20874).

### Possible drawbacks

Not expected.

### Applicable issues

- fixes #20874


rafariossaa (approved):

dgomezleon (approved):
LGTM

PR #52428, opened 2023-11-07
Enable pw-sha2. This partly implements #52373

### LINKED ISSUE #52373: [bitnami/openldap] Enable pw-sha2 module by default
### Name and Version

bitnami/openldap:2.6

The default password hash algorithms ({SSHA}, {SHA}, {SMD5}, {MD5}, {CRYPT}, and {CLEARTEXT}) provided for use with the LDAP_PASSWORD_HASH option may not provide sufficient password protection, primarily because they rely on variations of SHA1 and MD5.
More suitable algorithms like (S)SHA256 and (S)SHA512 can be accessed via the pw-sha2 module, which should already be included in current Bitnami builds of OpenLDAP.

Enabling the pw-sha2 module by default and recommending or configuring a stronger hash algorithm as the default choice, rather than SSHA, would be a valuable improvement for enhancing the out-of-the-box security of password stored by the OpenLDAP server.

Alternatively, users of the container could be prompted to manually enable the module and customize the password hash algorithm during initialization. However, this approach may be less user-friendly when compared to the convenience of using the LDAP_PASSWORD_HASH environment option or having a stronger default configuration in place.

maxemann96:
### Description of the change
Enables the pw-sha2 openldap module in the openldap container.

### Benefits
There are more secure hashing algorithms enabled.

### Possible drawbacks
Nothing disabled or unloaded. No drawbacks expected.

### Applicable issues
- partly fixes #52373

### Additional information
Although this merge request does not solve the ticket 100%, it should already be integrated as a first step and additional security improvement.

### DIFF
--- a/bitnami/openldap/2.5/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
+++ b/bitnami/openldap/2.5/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
@@ -281,6 +281,16 @@ olcArgsFile: /opt/bitnami/openldap/var/run/slapd.args
 olcPidFile: /opt/bitnami/openldap/var/run/slapd.pid
 
 #
+# Enable pw-sha2 module
+#
+
+dn: cn=module,cn=config
+cn: module
+objectClass: olcModuleList
+olcModulePath: /opt/bitnami/openldap/libexec/openldap
+olcModuleLoad: pw-sha2.so
+
+#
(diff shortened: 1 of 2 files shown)


Mauraza (approved):
Hi @maxemann96,

Thanks for the contribution!! LGTM!! 
This change will be available in the next revision.

mistial-dev:
An alternate approach has been implemented in #70398.

PR #52661, opened 2023-11-10
[SPARK] lang.NullPointerException: invalid null input: name

### LINKED ISSUE #52698: [SPARK] lang.NullPointerException: invalid null input: name
### Name and Version

bitnami/spark:3.5.0-debian-11-r12

amd64

I used the bitnami helm chart to setup spark.  

Then I started an example application from a client:

```
spark-submit \
    --class org.apache.spark.examples.SparkPi \
    --conf spark.kubernetes.container.image=docker.io/bitnami/spark:3.5.0-debian-11-r12 \
    --master k8s://https://<cluster-url> \
    --conf spark.kubernetes.driverEnv.SPARK_MASTER_URL=spark://<spark-cluster-url>:7077 \
    --deploy-mode cluster \
    --conf spark.kubernetes.authenticate.driver.serviceAccountName=spark \
    --name spark-pi \
    --conf spark.kubernetes.executor.podNamePrefix=test \
    --conf spark.kubernetes.executor.request.cores=100m \
    --conf spark.kubernetes.executor.request.memory=1Gi \
    --conf spark.executor.instances=1 \
    local:///opt/bitnami/spark/examples/jars/spark-examples_2.12-3.5.0.jar
```

It starts the driver and after that he starts spinning up executors:

```console
$ kubectl get pods
NAME                               READY   STATUS    RESTARTS      AGE
test-spark-master-0              1/1     Running   0             97m
spark-pi-bdf6ef8bba1672e6-driver   1/1     Running   0             9s
timtest-exec-1                     1/1     Running   0             3s
```

Now, the executor will crash over and over again, being replaced by "-2", "-3", ... untill he finally gives up and exits.  The pods remain in `Error` state.

I was able to get this log from the executor pods:

```
23/11/10 16:06:13 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Exception in thread "main" org.apache.hadoop.security.KerberosAuthException: failure to login: javax.security.auth.login.LoginException: java.lang.NullPointerException: invalid null input: name
	at jdk.security.auth/com.sun.security.auth.UnixPrincipal.<init>(UnixPrincipal.java:71)
	at jdk.security.auth/com.sun.security.auth.module.UnixLoginModule.login(UnixLoginModule.java:134)
	at java.base/javax.security.auth.login.LoginContext.invoke(LoginContext.java:755)
	at java.base/javax.security.auth.login.LoginContext$4.run(LoginContext.java:679)
	at java.base/javax.security.auth.login.LoginContext$4.run(LoginContext.java:677)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:712)
	at java.base/javax.security.auth.login.LoginContext.invokePriv(LoginContext.java:677)
	at java.base/javax.security.auth.login.LoginContext.login(LoginContext.java:587)
	at org.apache.hadoop.security.UserGroupInformation$HadoopLoginContext.login(UserGroupInformation.java:2065)
	at org.apache.hadoop.security.UserGroupInformation.doSubjectLogin(UserGroupInformation.java:1975)
	at org.apache.hadoop.security.UserGroupInformation.createLoginUser(UserGroupInformation.java:719)
	at org.apache.hadoop.security.UserGroupInformation.getLoginUser(UserGroupInformation.java:669)
	at org.apache.hadoop.security.UserGroupInformation.getCurrentUser(UserGroupInformation.java:579)
	at org.apache.spark.deploy.SparkHadoopUtil.createSparkUser(SparkHadoopUtil.scala:70)
	at org.apache.spark.deploy.SparkHadoopUtil.runAsSparkUser(SparkHadoopUtil.scala:61)
	at org.apache.spark.scheduler.cluster.k8s.KubernetesExecutorBackend$.run(KubernetesExecutorBackend.scala:67)
	at org.apache.spark.scheduler.cluster.k8s.KubernetesExecutorBackend$.main(KubernetesExecutorBackend.scala:56)
	at org.apache.spark.scheduler.cluster.k8s.KubernetesExecutorBackend.main(KubernetesExecutorBackend.scala)

	at org.apache.hadoop.security.UserGroupInformation.doSubjectLogin(UserGroupInformation.java:1986)
	at org.apache.hadoop.security.UserGroupInformation.createLoginUser(UserGroupInformation.java:719)
	at org.apache.hadoop.security.UserGroupInformation.getLoginUser(UserGroupInformation.java:669)
	at org.apache.hadoop.security.UserGroupInformation.getCurrentUser(UserGroupInformation.java:579)
	at org.apache.spark.deploy.SparkHadoopUtil.createSparkUser(SparkHadoopUtil.scala:70)
	at org.apache.spark.deploy.SparkHadoopUtil.runAsSparkUser(SparkHadoopUtil.scala:61)
	at org.apache.spark.scheduler.cluster.k8s.KubernetesExecutorBackend$.run(KubernetesExecutorBackend.scala:67)
	at org.apache.spark.scheduler.cluster.k8s.KubernetesExecutorBackend$.main(KubernetesExecutorBackend.scala:56)
	at org.apache.spark.scheduler.cluster.k8s.KubernetesExecutorBackend.main(KubernetesExecutorBackend.scala)
Caused by: javax.security.auth.login.LoginException: java.lang.NullPointerException: invalid null input: name
	at jdk.security.auth/com.sun.security.auth.UnixPrincipal.<init>(UnixPrincipal.java:71)
	at jdk.security.auth/com.sun.security.auth.module.UnixLoginModule.login(UnixLoginModule.java:134)
	at java.base/javax.security.auth.login.LoginContext.invoke(LoginContext.java:755)
	at java.base/javax.security.auth.login.LoginContext$4.run(LoginContext.java:679)
	at java.base/javax.security.auth.login.LoginContext$4.run(LoginContext.java:677)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:712)
	at java.base/javax.security.auth.login.LoginContext.invokePriv(LoginContext.java:677)
	at java.base/javax.security.auth.login.LoginContext.login(LoginContext.java:587)
	at org.apache.hadoop.security.UserGroupInformation$HadoopLoginContext.login(UserGroupInformation.java:2065)
	at org.apache.hadoop.security.UserGroupInformation.doSubjectLogin(UserGroupInformation.java:1975)
	at org.apache.hadoop.security.UserGroupInformation.createLoginUser(UserGroupInformation.java:719)
	at org.apache.hadoop.security.UserGroupInformation.getLoginUser(UserGroupInformation.java:669)
	at org.apache.hadoop.security.UserGroupInformation.getCurrentUser(UserGroupInformation.java:579)
	at org.apache.spark.deploy.SparkHadoopUtil.createSparkUser(SparkHadoopUtil.scala:70)
	at org.apache.spark.deploy.SparkHadoopUtil.runAsSparkUser(SparkHadoopUtil.scala:61)
	at org.apache.spark.scheduler.cluster.k8s.KubernetesExecutorBackend$.run(KubernetesExecutorBackend.scala:67)
	at org.apache.spark.scheduler.cluster.k8s.KubernetesExecutorBackend$.main(KubernetesExecutorBackend.scala:56)
	at org.apache.spark.scheduler.cluster.k8s.KubernetesExecutorBackend.main(KubernetesExecutorBackend.scala)

	at java.base/javax.security.auth.login.LoginContext.invoke(LoginContext.java:850)
	at java.base/javax.security.auth.login.LoginContext$4.run(LoginContext.java:679)
	at java.base/javax.security.auth.login.LoginContext$4.run(LoginContext.java:677)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:712)
	at java.base/javax.security.auth.login.LoginContext.invokePriv(LoginContext.java:677)
	at java.base/javax.security.auth.login.LoginContext.login(LoginContext.java:587)
	at org.apache.hadoop.security.UserGroupInformation$HadoopLoginContext.login(UserGroupInformation.java:2065)
	at org.apache.hadoop.security.UserGroupInformation.doSubjectLogin(UserGroupInformation.java:1975)
	... 8 more
```

It should not fail, it should just start

### Additional information

I figured out through research on the internet, that the executor script tries to find the name of the user and since the user is not defined in `/etc/passwd`, this will fail, resulting in an empty "name" string.  This means that the java command will get an "empty name variable", resulting in the error:
```
lang.NullPointerException: invalid null input: name
```
A lot of users work around this by creating their own image, correcting the problem in a number of different ways, like add a line to `/etc/passwd`. Some even remove the "USER" in the dockerfile and create it as a root container (!!).  

I can imagine you want this fixed at the source (the spark script itself), but this means that until this is fixed, a lot of people are overriding your image by fixing it themselves.  Why don't you add a working workaround now and get it out when they fix it at the source?

TimVerboisCgk:
### Description of the change
I found a bug in the startup procedure of the spark-executor usage of the spark image.  

The bug was located in this piece of code:

```shell
if [ ! $EUID -eq 0 ] && [ -e "$LIBNSS_WRAPPER_PATH" ]; then
    echo "spark:x:$(id -u):$(id -g):Spark:$SPARK_HOME:/bin/false" > "$NSS_WRAPPER_PASSWD"
    echo "spark:x:$(id -g):" > "$NSS_WRAPPER_GROUP"
    echo "LD_PRELOAD=$LIBNSS_WRAPPER_PATH" >> "$SPARK_CONF_DIR/spark-env.sh"
fi
```

this was perfect, unless the fact that `spark-env.sh` has permissions root:root and 644.  This code is executed with user 1001.  So, the first 2 lines of the then clause work perfectly, but the last part doesn't work.

This will result in the fact that the "whoami" command that is executed further down the start-up path will fail, since the user will not have an entry in the /etc/passwd file.  This file is the actively chosen file by whoami, since the LD_PRELOAD variable did not get set due to permission errors.  In that case, the pod will fail if it is started as a "spark executor".

### Code fix
So, what I did is remove this last line in that previous piece of code:

```shell
if [ ! $EUID -eq 0 ] && [ -e "$LIBNSS_WRAPPER_PATH" ]; then
    echo "spark:x:$(id -u):$(id -g):Spark:$SPARK_HOME:/bin/false" > "$NSS_WRAPPER_PASSWD"
    echo "spark:x:$(id -g):" > "$NSS_WRAPPER_GROUP"
fi
```

and add another `if` to the spark-env.sh file:

```shell
# NSS LD_PRELOAD setting
if [ ! $EUID -eq 0 ] && [ -e "$LIBNSS_WRAPPER_PATH" ]; then
    export LD_PRELOAD=$LIBNSS_WRAPPER_PATH
fi
```

The same test is done at runtime, so the same variables will be filled in as in the other `if`.  It then loads the LD_PRELOAD variable in case the `if` succeeds.

### Benefits
The Spark executor pod will succeed during startup and execute the spark application instead of crashing

### Possible drawbacks
Not that I know of

### Applicable issues

- fixes https://github.com/bitnami/containers/issues/52698

### Additional information

None

### DIFF
--- a/bitnami/spark/3.5/debian-11/rootfs/opt/bitnami/scripts/spark-env.sh
+++ b/bitnami/spark/3.5/debian-11/rootfs/opt/bitnami/scripts/spark-env.sh
@@ -95,4 +95,9 @@ export SPARK_METRICS_ENABLED="${SPARK_METRICS_ENABLED:-false}"
 export SPARK_DAEMON_USER="spark"
 export SPARK_DAEMON_GROUP="spark"
 
+# NSS LD_PRELOAD setting
+if [ ! $EUID -eq 0 ] && [ -e "$LIBNSS_WRAPPER_PATH" ]; then
+    export LD_PRELOAD=$LIBNSS_WRAPPER_PATH
+fi
+    
 # Custom environment variables may be defined below
--- a/bitnami/spark/3.5/debian-11/rootfs/opt/bitnami/scripts/spark/entrypoint.sh
+++ b/bitnami/spark/3.5/debian-11/rootfs/opt/bitnami/scripts/spark/entrypoint.sh
@@ -21,7 +21,6 @@ print_welcome_page
 if [ ! $EUID -eq 0 ] && [ -e "$LIBNSS_WRAPPER_PATH" ]; then
     echo "spark:x:$(id -u):$(id -g):Spark:$SPARK_HOME:/bin/false" > "$NSS_WRAPPER_PASSWD"
     echo "spark:x:$(id -g):" > "$NSS_WRAPPER_GROUP"
-    echo "LD_PRELOAD=$LIBNSS_WRAPPER_PATH" >> "$SPARK_CONF_DIR/spark-env.sh"
 fi
 
 if [[ "$1" = "/opt/bitnami/scripts/spark/run.sh" ]]; then


mdhont:
We are going to review this logic internally and we will notify you in this PR when there is any news.
Thank you very much for the contribution!

TimVerboisCgk:
Guys, this bug is real and several people are complaining about it, heaving workarounds installed, they even sometimes rebuild the image themselves.  
To get the helm chart working, I had to rebuild it myself just as the others did. So, I saw that the issue itself was already closed due to lack of activity, but can you review this?  If you don't agree with the logic, please propose another fix, that is also fine by me, but a fix is needed for this.

mdhont:
The env.sh files are automatically generates so any conditional statement would be overwritten there. In the bitnami jenkins [entrypoint.sh](https://github.com/bitnami/containers/blob/851a1f417acd8887f828823df7777373aaf1d28f/bitnami/jenkins/2/debian-11/rootfs/opt/bitnami/scripts/jenkins/entrypoint.sh#L22) we set the nns_wrapper dynamically based on the UID.

```
  # Set nss_wrapper vars only when running as non-root
  # Configure libnss_wrapper based on the UID/GID used to run the container
  # This container supports arbitrary UIDs, therefore we have do it dynamically
  if ! am_i_root; then
      export LNAME="jenkins"
      export LD_PRELOAD="/opt/bitnami/common/lib/libnss_wrapper.so"
      if ! user_exists "$(id -u)" && [[ -f "$LD_PRELOAD" ]]; then
          info "Configuring libnss_wrapper"
          NSS_WRAPPER_PASSWD="$(mktemp)"
          export NSS_WRAPPER_PASSWD
          NSS_WRAPPER_GROUP="$(mktemp)"
          export NSS_WRAPPER_GROUP
          echo "jenkins:x:$(id -u):$(id -g):Jenkins:${JENKINS_HOME}:/bin/false" >"$NSS_WRAPPER_PASSWD"
          echo "jenkins:x:$(id -g):" >"$NSS_WRAPPER_GROUP"
          chmod 400 "$NSS_WRAPPER_PASSWD" "$NSS_WRAPPER_GROUP"
          export HOME="$JENKINS_HOME"
      fi
  fi
```
Perhaps something similar can be done here.

TimVerboisCgk:
@mdhont the code I added to `spark-env.sh` is not overwritten during startup.  This file already contains a lot of default code: exports, for loops and if statements.  Adding one more is not really a problem in my opinion.  

The issue of the problem is here that `spark-env.sh` is not writable at the moment this line is executed at runtime:

```
echo "LD_PRELOAD=$LIBNSS_WRAPPER_PATH" >> "$SPARK_CONF_DIR/spark-env.sh"
```

so, this means it doesn't work.  Or we add something to the `spark-env.sh` file as I did, or we make sure this environment parameters is populated somewhere else.  It can only be populated if this test succeeds:

```bash
if [ ! $EUID -eq 0 ] && [ -e "$LIBNSS_WRAPPER_PATH" ]; then
```

Now, it would be logical to just add it to the `entrypoint.sh` script (with an export) like this:

```bash
if [ ! $EUID -eq 0 ] && [ -e "$LIBNSS_WRAPPER_PATH" ]; then
    echo "spark:x:$(id -u):$(id -g):Spark:$SPARK_HOME:/bin/false" > "$NSS_WRAPPER_PASSWD"
    echo "spark:x:$(id -g):" > "$NSS_WRAPPER_GROUP"
    export LD_PRELOAD=$LIBNSS_WRAPPER_PATH
fi
```

The question is then: why wasn't this done in the first place.  Since everything is always started from the entrypoint.  

This would mean it is only loaded from the `entrypoint.sh`script and everything that is started there, but as far as I can see, this is always the case:

```bash
exec "${CMD[@]}"
```

So, this makes your point, I believe, in adding it to the `entrypoint.sh` script.  The only question remains then: why was it explicitly added to the `spark-env.sh` script.   

Do you know if we can stop adding it to `spark-env.sh`?

TimVerboisCgk:
Guys, 

we cannot just ignore this.  More people are heaving the same problems.  You asked me to get involved and to do a proposition to improve it, but this has been blocked for months now.  We talked about this, we looked at different solutions, but without an approval, we cannot fix this ...

We really want to help you guys and add more improvements to images if this is needed, but if it always takes months before a bug is fixed because there is no activity and we don't get an approval, what is the point of getting involved?  In that case, we can just create wrappers around your images to fix things and this is just a pitty and not what I expected from a community ...

Tim

jcrist1:
I'm running into the issue mentioned here [https://github.com/bitnami/containers/issues/52698]. @mdhont is the only thing blocking this that there is not an implementation that dynamically configures the login credentials? @TimVerboisCgk do y'all have an image published somewhere that addresses this?

TimVerboisCgk:
@jcrist1

I have an image we already use while waiting for the official fix.  This works fine, but hasn't been updates in the past few months. I uploaded it to my dockerhub account, you can try it here:

https://hub.docker.com/r/verboistim/spark

Now, as I stated above, we proposed a fix, but it doesn't have to be this fix.  The point is, we really need A fix, it doesn't matter which.  Since you are heaving the same issues, it just confirms my statement on this.

rafariossaa:
Hi,
  Sorry for the delay, we are a small team and sometime we need to prioritize tasks.

  I am looking on how to integrate it with our current pipeline. We generate the `xxxx-env.sh` files from internal metadata, and code with logic (`if` condition) is not supported by the tooling. 
  I am trying to figure out a way to do a similar approach but that is "compatible" with our tooling.

jcrist1:
Thanks @TimVerboisCgk Can confirm your fix gets me further along (now have link errors with BZip2 libs 🤪 )

PR #53119, opened 2023-11-22
[bitnami/fluentd] Exclude gems path from sed-in-place test

FraPazGal:
### Description of the change

This fixes an ongoing Goss issue in our internal testing pipelines.

### Benefits

None
### Possible drawbacks

None

### Additional information

N/A

PR #53143, opened 2023-11-23
[bitnami/fluentd] Add sed_in_place.exclude_paths

gongomgra:
### Description of the change

Updates `sed_in_place.exclude_paths` array to avoid false positives with third-party scripts like the one below

```
$ grep sed /opt/bitnami/fluentd/gems/ffi-1.16.3/ext/ffi_c/libffi/testsuite/emscripten/build-tests.sh
ls *c | sed 's!\(.*\)\.c!sed -i "s/main/test__\1/g" \0!g' | bash
ls *.c | sed 's/\(.*\)\.c/emcc $CFLAGS -c \1.c -o \1.o /g' | bash
ls *.cc | sed 's/\(.*\)\.cc/em++ $CXXFLAGS -c \1.cc -o \1.o /g' | bash
```

### Benefits

False positives won't prevent future Fluentd releases.

### Possible drawbacks

Not expected.

### DIFF
--- a/.vib/fluentd/goss/vars.yaml
+++ b/.vib/fluentd/goss/vars.yaml
@@ -38,3 +38,7 @@ version:
   bin_name: fluentd
   flag: --version
 root_dir: /opt/bitnami
+# We can't control what third-party devs include in their internal scripts.
+sed_in_place:
+  exclude_paths:
+    - /opt/bitnami/fluentd/gems


gongomgra:
Closed as a duplicated of https://github.com/bitnami/containers/pull/53119

PR #53233, opened 2023-11-27
[bitnami/nginx] Remove deprecated TLS version

d4rklynk:
## TLS 1.0 and TLS 1.1 should be avoided at all cost.

-> https://www.ssl.com/guide/ssl-best-practices/#ftoc-heading-4

I removed TLS 1.0 and TLS 1.1 because it is deprecated and should not be used anymore.


gongomgra (approved):
Thanks for your contribution!

PR #53369, opened 2023-12-04
[bitnami/opensearch] Fix: Set correct `LD_LIBRARY_PATH` for OpenSearch

arturgrigor:
### Description of the change

Set correct `LD_LIBRARY_PATH` for OpenSearch v1 and v2. The `LD_LIBRARY_PATH` has been adjusted to ensure compatibility with both OpenSearch versions.

This change fixes the following runtime error:
> java.lang.UnsatisfiedLinkError:  no opensearchknn_nmslib in java.library.path: /usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib

More details can be found here: https://github.com/opensearch-project/k-NN/issues/508


jotamartos (approved):
LGTM! These changes will be included in the next release of the solution.

Issue #53068, opened 2023-11-21
[bitnami/etcd] Latest etcd image does not work


Kunamatata:
### Name and Version

bitnami/etcd:latest

### What architecture are you using?

None

### What steps will reproduce the bug?

1. Run `docker-compose up etcd` with the following snippet of `docker-compose.yml`
```yaml
  etcd:
    image: bitnami/etcd:3.5.9
    environment:
      ALLOW_NONE_AUTHENTICATION: "no"
      ETCD_ROOT_PASSWORD: "root"
    ports:
      - "2379:2379"
      - "2380:2380"
```

### What is the expected behavior?

It should be able to start. Seems like `MY_STS_NAME` might be the issue?

### What do you see instead?

Output of `docker-compose up etcd`
```
etcd-1  | etcd 01:15:49.42 INFO  ==> 
etcd-1  | etcd 01:15:49.42 INFO  ==> Welcome to the Bitnami etcd container
etcd-1  | etcd 01:15:49.42 INFO  ==> Subscribe to project updates by watching https://github.com/bitnami/containers
etcd-1  | etcd 01:15:49.43 INFO  ==> Submit issues and feature requests at https://github.com/bitnami/containers/issues
etcd-1  | etcd 01:15:49.43 INFO  ==> 
etcd-1  | etcd 01:15:49.43 INFO  ==> ** Starting etcd setup **
etcd-1  | etcd 01:15:49.44 INFO  ==> Validating settings in ETCD_* env vars..
etcd-1  | etcd 01:15:49.44 INFO  ==> Initializing etcd
etcd-1  | etcd 01:15:49.44 INFO  ==> Generating etcd config file using env variables
etcd-1  | etcd 01:15:49.45 INFO  ==> There is no data from previous deployments
etcd-1  | etcd 01:15:49.46 INFO  ==> Obtaining cluster member ID
etcd-1  | etcd 01:15:49.46 INFO  ==> Starting etcd in background
etcd-1  | {"level":"warn","ts":"2023-11-21T01:15:52.533989Z","logger":"etcd-client","caller":"v3@v3.5.10/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0x40003ca700/127.0.0.1:2379","attempt":0,"error":"rpc error: code = FailedPrecondition desc = etcdserver: authentication is not enabled"}
etcd-1  | etcd 01:15:52.54 INFO  ==> Stored member ID: 8e9e05c52164694d
[...]

PR #53782, opened 2023-12-13
[bitnami/etcd] add MY_STS_NAME to etcd-env.sh

Kunamatata:
### Description of the change

This PR adds the `MY_STS_NAME` environment variable to both `etcd-env.sh` to address https://github.com/bitnami/containers/issues/53068


andresbono:
Hi, thank you for your contribution with this PR.

The function that makes use of `MY_STS_NAME` should not even be called if the container is not running on k8s, that is the actual bug.

In parallel to the bug report https://github.com/bitnami/containers/issues/53068, we have addressed the issue. The next release of the etcd container should fix it.

We will post a message there so you can try again when the new container is available. Thank you.

PR #53817, opened 2023-12-14
[bitnami/kafka] embedded jmx exporter agent

### LINKED ISSUE #53816: Add JMX to Prometheus exporter to Kafka image
### Name and Version

bitnami/kafka 26.0.0

See https://github.com/bitnami/charts/issues/21309

I am working on a PR for https://github.com/bitnami/charts/issues/21309 for which I need to include the jmx exporter in the Kafka image. 

_No response_

janvanbesien-ngdata:
### Description of the change

Add the jmx exporter for prometheus in the Kafka container. This allows running it as an embedded java agent in the Kafka broker JVM.

### Benefits

This is a prerequisite for https://github.com/bitnami/charts/issues/21309, see details there.

fixes #53816

### Possible drawbacks

The image is now slightly larger.

### Applicable issues

None

### Additional information

None

### DIFF
--- a/bitnami/kafka/3.6/debian-11/Dockerfile
+++ b/bitnami/kafka/3.6/debian-11/Dockerfile
@@ -31,6 +31,7 @@ RUN mkdir -p /tmp/bitnami/pkg/cache/ ; cd /tmp/bitnami/pkg/cache/ ; \
       "java-17.0.9-11-6-linux-${OS_ARCH}-debian-11" \
       "render-template-1.0.6-4-linux-${OS_ARCH}-debian-11" \
       "kafka-3.6.1-0-linux-${OS_ARCH}-debian-11" \
+      "jmx-exporter-0.20.0-1-linux-${OS_ARCH}-debian-11" \
     ) ; \
     for COMPONENT in "${COMPONENTS[@]}"; do \
       if [ ! -f "${COMPONENT}.tar.gz" ]; then \


juan131:
Hi @janvanbesien-ngdata 

Thanks so much for this contribution. Unfortunately, I'm afraid we cannot accept it. 

Including the JMX exporter as a JAVA agent both increases the Kafka container size & increments the attack surface on the Kafka container. It also goes against the "one container - one process" principle.

Please note some of the missing metrics on standalone mode such as memory & CPU consumption can be retrieved from other metrics exporters (e.g. the [Metrics Server](https://github.com/bitnami/containers/tree/main/bitnami/metrics-server)).

With this said, your contribution is valid from a technical point of view. So please feel free to build your custom image with the changes you're proposing and use it in the Kafka chart customizing the `image.registry`, `image.repository` & `image.tag` properties.

janvanbesien-ngdata:
Thank you for your review. I respectfully disagree, but obviously it is your call. We can indeed make it work with a custom image, but we were trying to avoid having to maintain that. Let me try one last time to give you some arguments pro this approach:

- Yes it increases the Kafka container size (and increments the attack surface), but it adds features. That's a balance you have to make for every new functionality, in this case I'd certainly argue that the benefits outweigh the additional complexity. It even reduces complexity (and attack surface) in some areas as well: there is no longer a need to expose remote JMX on the Kafka broker.
- The proposed approach is very common in Java. It is what https://github.com/prometheus/jmx_exporter suggests themselves. It is also what is done in for example ZooKeeper (including in the bitnami/zookeeper chart, but because it is built-in in ZooKeeper it is less noticeable). 
- The memory and CPU consumption obtained from https://github.com/bitnami/containers/tree/main/bitnami/metrics-server is not what is needed here. The latter is on OS process level, while the prometheus agent gives internal JVM level details (e.g. of the different memory pools in the JVM).

juan131:
Hi @janvanbesien-ngdata 

> there is no longer a need to expose remote JMX on the Kafka broker.

There's no need with the current setup either. Please note containers within a pod share the network namespace, including the IP address and network ports. Therefore, the JMX exporter sidecar can access JMX port via "localhost" without exposing it to the outside.

> The latter is on OS process level, while the prometheus agent gives internal JVM level details (e.g. of the different memory pools in the JVM).

Yes, that's the main disadvantage the standalone option offers: there's no way to obtain such a "fine-grained" information about memory/cpu consumption.

> I respectfully disagree, but obviously it is your call

We really appreciate your feedback. As I said on my previous response, your approach is 100% valid and it definitely has important pros. At this moment, keeping our images minimal & reducing potential vulnerabilities is one of our main priorities and that'a the main reason that sustains my position.

In any case, I already shared this with one of my colleagues but I'm going to open it to the whole team and collect their feedback. In case they align with you, I'm opened to revisit this decision.

OneCricketeer:
I think the strimzi operator uses an init container to download and mount the JMX agent first before starting the broker process. Therefore, the base container doesn't need it. 

Otherwise, you can pull the bitnami container on your own, add the JAR, then update the necessary `ENV KAFKA_OPTS` to include the agent for modifying runtime behavior. This would also apply to other agents like NewRelic, Otel, Datadog, etc.

PR #53960, opened 2023-12-19
Update openldap libopenldap.sh

### LINKED ISSUE #120: ldif in /ldiffs not loaded
**Describe the bug**

I added some ldiff (root / ou / users) in a configmap or in the customLdifFiles value, the files was created in /ldiffs but nothing was done with ldapsearch or with phpldapadmin. When i add the files with ldapadd it's ok. I tried with different openldap versions.

**To Reproduce**
use customLdifFiles:
  00-root.ldif: |-
    dn: dc=example,dc=com
    objectClass: dcObject
    objectClass: organization
    dc: example
    o: Example

**Expected behavior**
ldiff added with customLdifFiles or configmap

### LINKED ISSUE #136: Not compatible with Bitnami/openldap 2.6.6
Hello,

When i try to update the docker image with bitnami/openldap:2.6.6 it's not working correctly.
The Pod up but the replication is not enable. 

In the log i can see that the {0}module din't create correctly. Bitnami in openldap 2.6.6 add a {0}module with a oclModulePath. In this path the sync module is also not present.

module

jp-gouin:
### Description of the change

During Openldap init, this change does the TLS configuration before adding custom LDIFs. 

### Benefits

Using this [helm chart](https://github.com/jp-gouin/helm-openldap), this change fixes a replication issue.
This chart allows users to spin up Openldap in K8S in a high availability mode. In a multi-node scenario, the first one starts without issue, other ones are in a race condition when setting up the tree (`ldap_create_tree`) and ldif `ldap_add_custom_ldifs`) due to the replication. This is an expected behaviour and once the replication kicks in all nodes are running.

By doing the TLS initialisation at the end of `ldap_initialize()` the TLS is not properly configured when Openldap restart.

### Applicable issues

- fixes https://github.com/jp-gouin/helm-openldap/issues/136 
- fixes https://github.com/jp-gouin/helm-openldap/issues/120

### DIFF
--- a/bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
+++ b/bitnami/openldap/2.6/debian-11/rootfs/opt/bitnami/scripts/libopenldap.sh
@@ -646,19 +646,19 @@ ldap_initialize() {
         if is_boolean_yes "$LDAP_ENABLE_SYNCPROV"; then
             ldap_enable_syncprov
         fi
-        if ! is_dir_empty "$LDAP_CUSTOM_LDIF_DIR"; then
-            ldap_add_custom_ldifs
-        elif ! is_boolean_yes "$LDAP_SKIP_DEFAULT_TREE"; then
-            ldap_create_tree
-        else
-            info "Skipping default schemas/tree structure"
-        fi
         # enable tls
         if is_boolean_yes "$LDAP_ENABLE_TLS"; then
             ldap_configure_tls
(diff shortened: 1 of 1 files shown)


FraPazGal (approved):
LGTM, thanks for the contribution @jp-gouin!

jpsn123:
@FraPazGal  why this change do not apply to debian-12, moreover, are there any more pr have been lost.

@jp-gouin 
this mistake waste of developers' hard work. hope to be valued by maintainer.

jpsn123:
it seem to be reverted by stupid bot, #55570

Issue #54028, opened 2023-12-21
MySQL replica when missing master binary logs


michel-silva:
### Name and Version

bitnami/mysql:8.0

### What architecture are you using?

None

### What steps will reproduce the bug?

1 - Master server with data but without all binary files;
2 - Building a slave container;

### What is the expected behavior?

The replication would be fail, but the passwords should working 

### What do you see instead?

After create container the replication fail and I tried to log to MySQL slave to run MySQL dump extract from master, but the passwords doesn’t work.

### Additional information

I don’t be certain that is a bug. But I need create a slave server and didn't it work

michel-silva:
This other issue describe exactly what occur in my slave.

https://github.com/bitnami/containers/issues/34145

carrodher:
Thank you for opening this issue and submitting the associated Pull Request. Our team will review and provide feedback. Once the PR is merged, the issue will automatically close.

Your contribution is greatly appreciated!

PR #54038, opened 2023-12-22
[bitnami/mysql] Feat mysqldump master database

michel-silva:
### Description of the change
Implement mysql dump to cases of missing binary files  

### Benefits
The dump will be executed before replication start. If binary files is missing will not broken replication 

### Possible drawbacks
The master server must be used only this process

### Applicable issues
https://github.com/bitnami/containers/issues/54028


carrodher:
Thanks for your contribution! Could you please check the markdown issue reported by the linter? Thanks

michel-silva:
@carrodher Thank you for your fast reply.
I've resolved the lint messages and removed my commit that doesn't have the signature.

Could you check if it's all right? Let me know if I could make something to merge

michel-silva:
Hi Carlos,

I’m resolved lint message and remove the commit without signature.
Let me know if I could make something to this merge.

Thank you,

Michel Silva
Senior Software Developer
+55 11 9 9887 1808


Em sex., 22 de dez. de 2023 às 08:54, Carlos Rodríguez Hernández <
***@***.***> escreveu:

dgomezleon (changes requested):
Thanks for contributing. I added some suggestions.

michel-silva:
@dgomezleon Thank you for your review. I made the changes, could you check again?

bitnami/mysql/8.0/debian-11/rootfs/opt/bitnami/scripts/libmysql.sh
```diff
@@ -178,6 +178,67 @@ pid_file=${DB_PID_FILE}
...
+
+        FILE_LOCATION="$DB_DATA_DIR/dump_$DB.sql"
+
+        mysqldump --verbose -h "$DB_MASTER_HOST" -P "$DB_MASTER_PORT_NUMBER" -u "$DB_MASTER_ROOT_USER" -p"$DB_MASTER_ROOT_PASSWORD" $DB > $FILE_LOCATION
+
+        info "Finish dump database $DB"
+
+        if [ $? -eq 0 ]; then
+            head -n 15 $FILE_LOCATION
+
```
dgomezleon, line 217:
```suggestion
```
I guess it was added to debug the code.

bitnami/mysql/8.0/debian-11/rootfs/opt/bitnami/scripts/libmysql.sh
```diff
@@ -178,6 +178,67 @@ pid_file=${DB_PID_FILE}
...
+            mysql -u "$DB_MASTER_ROOT_USER" <<EOF 
+    create database if not exists $DB;
+EOF
+
+            info "Start import dump database $DB"
+            mysql_execute $DB < $FILE_LOCATION
+            info "Finish import dump database $DB"
+
+        else
+            info "Fail create dump"
```
dgomezleon, line 229:
```suggestion
            info "Error creating dump"
```

dgomezleon (changes requested):
Thanks for applying the changes. I detected a couple of minor typos. Could you please take a look?

michel-silva:
@dgomezleon Thank you for the new review. The changes have been applied.

bitnami/mysql/8.0/debian-11/rootfs/opt/bitnami/scripts/libmysql.sh
```diff
@@ -178,6 +178,65 @@ pid_file=${DB_PID_FILE}
...
+#   DB_*
+# Arguments:
+#   None
+# Returns:
+#   None
+#########################
+mysql_exec_initial_dump() {
+    info "MySQL dump master data start..."
+    
+    mysql -h "$DB_MASTER_HOST" -P "$DB_MASTER_PORT_NUMBER" -u "$DB_MASTER_ROOT_USER" -p"$DB_MASTER_ROOT_PASSWORD" -e 'RESET MASTER;'
```
dgomezleon, line 193: I have one last question about this. This would execute `RESET MASTER` for every slave. Is this necessary? 
Note that RESET MASTER is already executed in [mysql_initialize() ](https://github.com/bitnami/containers/blob/f81b69b67f0bd59bd3d1a09a3f141938722456f8/bitnami/mysql/8.0/debian-11/rootfs/opt/bitnami/scripts/libmysql.sh#L309). Could you please confirm?

dgomezleon (commented):
Thanks again. Could you please check my last question?

bitnami/mysql/8.0/debian-11/rootfs/opt/bitnami/scripts/libmysql.sh
michel-silva: Hi @dgomezleon.
Sorry about my lazy response.. I was on vacation and I returned in the last days.
Your question is very interesting and makes me think about how replication works. So let me explain my point of view.

I have a master setup and a slave was broken because of a crash on the slave server.. So, I make a reinstallation of the new slave server and when I up the MySQL slave server this [error](https://github.com/bitnami/containers/issues/54028) occurs and this error is because the Mysql master server is missing some binary files. To fix it, I could make a manual mysql dump on master and update the slave, but to do it I must login in recovery mode on MySQL slave because the mysql.user could not be transferred by binary files and we do not have any valid user to login. Analyzing this scenario, I make this approach to make an auto mysql dump on master and import to a slave database.
 
Analyzing your question about the mysql_initialize(), this is only exec on the first up of the container. In my scenario, the reset will not be executed and the mysql master binary files are not modified. To resolve this, I could manually get mysql dump on the master server, make the reset master and run the mysql dump in slave. 

So, when the MySQL master server doesn't have all the binary files, for each slave that is added to the server the reset master will be necessary for the slave to start replicating at the point of the master server stop.

One important point that your question made me think about is: that is very important that all slave servers have the last data present in the master database, when we restart replication after reset master the slave has the same data and does not miss any data. So, I think is important to add one alert to make sure that all slaves have all data updated, before start the setup the new slave.

I don't know if it explains my point of view. But I'm open to having some discussion about this and making the better to the package.

dgomezleon (approved):

mostafa-rz:
@michel-silva Thanks for your PR. I have this problem and your feature will help me a lot. I have two questions:

1. Why you are not using `mysqldump --all-databases --master-data` instead of [searching for the tables](https://github.com/bitnami/containers/blob/293342a68f1896b2eafcc49aa5e92ff18a312e47/bitnami/mysql/8.0/debian-11/rootfs/opt/bitnami/scripts/libmysql.sh#L193)? This is the way mysql suggests on their [website](https://dev.mysql.com/doc/refman/8.0/en/replication-snapshot-method.html). (Please look at _17.1.2.5.1 Creating a Data Snapshot Using mysqldump_)
2. With your current solution the user tables on the master are not copied therefore the slave is not the same as the master. Is this intentional?

mostafa-rz:
There is also another problem. If I increase the number of slaves, the previous slaves will lose the connection to the master because of the `RESET MASTER` flag in this [line](https://github.com/bitnami/containers/blob/293342a68f1896b2eafcc49aa5e92ff18a312e47/bitnami/mysql/8.0/debian-11/rootfs/opt/bitnami/scripts/libmysql.sh#L193). It is also mentioned in mysql [documentation ](https://dev.mysql.com/doc/refman/8.0/en/reset-master.html).

> [RESET MASTER](https://dev.mysql.com/doc/refman/8.0/en/reset-master.html) is not intended to be used while any replicas are running. The behavior of [RESET MASTER](https://dev.mysql.com/doc/refman/8.0/en/reset-master.html) when used while replicas are running is undefined (and thus unsupported), whereas [PURGE BINARY LOGS](https://dev.mysql.com/doc/refman/8.0/en/purge-binary-logs.html) may be safely used while replicas are running.

_**Step to reproduce:**_

1.  Start with 1 slave `docker compose -f docker-compose-replication.yml up`
2. Increase number of slaves `docker-compose -f docker-compose-replication.yml up --detach --scale mysql-master=1 --scale mysql-slave=3`
3. Create a test database in master
4. The test database shown in **_slave-2_** but not in **_slave-1_**

michel-silva:
@mostafa-rz Sorry about my delay response. I will analyse your points and create a new pr to resolve this questions. 
Thank you for your valuation.

michel-silva:
@mostafa-rz this PR fix the issue. 
https://github.com/bitnami/containers/pull/55126/

PR #54384, opened 2024-01-03
Create docker-image.yml

Sehquerry:
### Applicable issues

- fixes #


carrodher:
Thanks for your contribution, unfortunately it can't be merged since our build and test process is already defined and it involves different steps

PR #54799, opened 2024-01-16
[bitnami/*] Remove undesired composes

FraPazGal:
### Description of the change

As of now, we do offer `docker-compose.yml` files for all of our container applications. We are now reviewing this approach to only leave those compose files that are actually useful and removing those that do not comply with the following criteria:

- A compose file is kept if:
   - The application requires a multi-container or complex setup AND
   - Upstream developers explicitly support a compose setup
- A compose file is removed if:
   - The application does NOT require a multi-container or complex setup OR
   - Upstream developers do NOT explicitly support a compose setup OR
   - The container is an auxiliary one to a bigger setup (E.g.: parse-dashboard is used in a Parse setup) OR
   - The application is Kubernetes native OR
   - The application is a CLI/exporter

### Benefits

Trimming of useless and non-supported `docker-compose` files.

### Possible drawbacks

There may be users using a compose file to be removed in this PR. Where necessary, we will be updating the installation instructions to use the `docker` CLI instead of Docker Compose.


fevisera:
There is an additional of `docker-compose` in https://github.com/bitnami/containers/blob/main/CONTRIBUTING.md?plain=1#L53

FraPazGal:
Updated, thanks!!

carrodher (approved):

PR #55276, opened 2024-01-22
Skip setting kafka config from empty environment variables

yves-tutti:
### Description of the change
When generating the kafka config do not set empty configuration settings.

### Benefits
With the change in https://github.com/bitnami/containers/commit/0b8731370e900f34e525e70311751e10747792da for KAFKA_CFG_MESSAGE_MAX_BYTES the server option `message.max.bytes` would otherwise always be set to empty string which is not accepted as a valid value by the kafka server.

The error message with `bitnami/kafka:3.3.2-debian-11-r244` looks like 
```
Exception in thread "main" org.apache.kafka.common.config.ConfigException: Invalid value  for configuration message.max.bytes: Not a number of type INT
```

### Possible drawbacks
Other variables, for which an empty value wasn't a problem before also won't be set anymore to an empty string value.

### Additional information
I could test this only in a very limited use-case and I'm not very familiar with all of it. This change should be tested by others before merging it.

### DIFF
--- a/bitnami/kafka/3.2/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
+++ b/bitnami/kafka/3.2/debian-11/rootfs/opt/bitnami/scripts/libkafka.sh
@@ -816,7 +816,8 @@ kafka_configure_from_environment_variables() {
         done
 
         value="${!var}"
-        kafka_server_conf_set "$key" "$value"
+        # Skip empty variables from kafka-env.sh
+        ! is_empty_value "$value" && kafka_server_conf_set "$key" "$value"
     done
 }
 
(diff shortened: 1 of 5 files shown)


carrodher:

yves-tutti:
@carrodher I'll force-push with a new commit that contains the sign-off line.
I didn't see that this is required but I only read this section on the frontpage: https://github.com/bitnami/containers?tab=readme-ov-file#contributing , could we add it there?

epot (approved):

yves-tutti:
To me it seems some funny automation closed the PR, sorry about that.

epot (approved):

Dynee (approved):

jotamartos (approved):
Changes LGTM! Our pipeline will release a new version of the containers soon.

PR #55332, opened 2024-01-24
[bitnami/airflow] Update extra packages to check

mdhont:
### Description of the change

Updating the packages to check as some packages are not available for version 2.8.1

```
  WARNING: apache-airflow 2.8.1 does not provide the extra 'all_dbs'
  WARNING: apache-airflow 2.8.1 does not provide the extra 'cncf.kubernetes'
```

We also install the extra async package (the async classes for gunicorn)  but this no longer shows as a separate package. I'm replacing the missing packages with alternatives to check.

### DIFF
--- a/.vib/airflow/goss/vars.yaml
+++ b/.vib/airflow/goss/vars.yaml
@@ -21,15 +21,14 @@ version:
   bin_name: airflow
   flag: version
 subpackages:
-  - async
   - amazon
   - celery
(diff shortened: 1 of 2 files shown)


FraPazGal (dismissed):
LGTM! Remember to remove the VIB trigger before merging

FraPazGal (approved):

PR #55562, opened 2024-01-29
[bitnami/minio] Update minio-env.sh to allow custom path for persist

Mo0rBy:
…nce directory and updated README.md

### Description of the change

This change allows the `MINIO_DATA_DIR` environment variable to be a configurable value so that the persistence directory can exist at a path inside the container that is not `/bitnami/minio/data`.

### Benefits

This allows users to have a persistence directory that is not the default `bitnami/minio/data` path so that mounting the persistence volume to something like AWS EFS or EBS can be done with naming conventions suited to the user/team.

### Possible drawbacks

This is very small change, I honestly can't think of any apart from something to check while troubleshooting/debugging.

### Applicable issues

My own issue created in `bitnami/charts` > https://github.com/bitnami/charts/issues/21757 (I will be looking at creating a Helm chart PR in the next few days as well, when I get time).

### Additional information

I tested this by running the container image locally, going into the UI and creating a new bucket called `test-bucket` and you can see that when I shell into the container, there is a `test-bucket` directory created at the custom persistence path.

Here is a screenshot of my terminal to prove that the bucket exists in the correct place (inside `custom-data-dir`):

Screenshot from 2024-01-29 22-36-15

Screenshot from 2024-01-29 22-37-20

I would not consider this a limitation but this configuration requires that the user deploying the container must have the container path mount and the environment variables as the same value. This is ONLY when configuring the container to use a custom persistence path, the default `/bitnami/minio/data` can still be used. I have updated the relevant section of the `README.md` which explains.

### DIFF
--- a/bitnami/minio/2024/debian-11/rootfs/opt/bitnami/scripts/minio-env.sh
+++ b/bitnami/minio/2024/debian-11/rootfs/opt/bitnami/scripts/minio-env.sh
@@ -39,6 +39,7 @@ minio_env_vars=(
     MINIO_FORCE_NEW_KEYS
     MINIO_ROOT_USER
     MINIO_ROOT_PASSWORD
+    MINIO_DATA_DIR
 )
 for env_var in "${minio_env_vars[@]}"; do
     file_env_var="${env_var}_FILE"
@@ -60,7 +61,7 @@ export MINIO_CERTS_DIR="/certs"
 export MINIO_LOGS_DIR="${MINIO_BASE_DIR}/log"
 export MINIO_TMP_DIR="${MINIO_BASE_DIR}/tmp"
 export MINIO_SECRETS_DIR="${MINIO_BASE_DIR}/secrets"
-export MINIO_DATA_DIR="/bitnami/minio/data"
+export MINIO_DATA_DIR="${MINIO_DATA_DIR:-/bitnami/minio/data}"
 export MINIO_LOG_FILE="${MINIO_LOGS_DIR}/minio.log"
 export MINIO_PID_FILE="${MINIO_TMP_DIR}/minio.pid"
 export PATH="${MINIO_BIN_DIR}:${PATH}"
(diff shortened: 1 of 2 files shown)


bitnami/minio/2024/debian-11/rootfs/opt/bitnami/scripts/minio-env.sh
```diff
@@ -78,6 +78,7 @@ export MINIO_DISTRIBUTED_MODE_ENABLED="${MINIO_DISTRIBUTED_MODE_ENABLED:-no}"
 export MINIO_DEFAULT_BUCKETS="${MINIO_DEFAULT_BUCKETS:-}"
 export MINIO_STARTUP_TIMEOUT="${MINIO_STARTUP_TIMEOUT:-10}"
 export MINIO_SERVER_URL="${MINIO_SERVER_URL:-$MINIO_SCHEME://localhost:$MINIO_API_PORT_NUMBER}"
+export MINIO_DATA_DIR="${MINIO_DATA_DIR:-/bitnami/minio/data}"
```
FraPazGal, line 81: Minor nitpick; could you move it back to the path section? It's true it could also be considered part of MinIO config but having all the path-related envs grouped seems more intuitive to me.

FraPazGal (commented):
Thanks a lot for the contribution @Mo0rBy! The PR looks good on my side, though I added a minor comment as review. Please check it whenever you can!

bitnami/minio/2024/debian-11/rootfs/opt/bitnami/scripts/minio-env.sh
Mo0rBy: Sure! I was 50/50 on where this should go in here as everything in the "path" section is non-configurable, but this is still a path, but now it is a configurable path!

Mo0rBy: Done

Mo0rBy: @FraPazGal Bumping this

FraPazGal: Perfect, the PR LGTM @Mo0rBy! Let me run a couple of check in our internal pipeline before merging this.

FraPazGal (approved):
LGTM!

PR #55910, opened 2024-02-01
[bitnami/mongodb-sharded] Update libmongodb.sh : mongodb_is_primary_node_up() function : grep check to find if mongod instance changed from secondary to primary

esasidharan:
…heck to find if mongod instance changed from secondary to primary

In the Bitnami script file libmongodb.sh, in mongodb_is_primary_node_up(), the grep chec to see if the MongoDB instance has turned from secondary to primary has bug.

Reason : The problem is in the line grep -q "true" <<<"$result". As part of the $result output, MongoDB gives a connection string which also has the string "true".

Hence even when it is secondary, i.e. output of db.isMaster().ismaster is false, the grep -q "true" <<<"$result" check passes and it proceeds to create the root user. When it tries to create a root user during the secondary state, the root user creation fails. In faster systems, it turns into primary quickly and this bug in the code doesn't matter.

Changing the grep check into a more specific one like the following helps to check if the mongodb instance turns into primary and then the root user gets created successfully and Authentication passes.

grep -q "\[direct: primary\] admin> true" <<<"$result"

### Applicable issues

- fixes #

### DIFF
--- a/bitnami/mongodb-sharded/7.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb.sh
+++ b/bitnami/mongodb-sharded/7.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb.sh
@@ -1052,7 +1052,7 @@ mongodb_is_primary_node_up() {
 db.isMaster().ismaster
 EOF
     )
-    grep -q "\[direct: \S\+\] admin> true" <<<"$result"
+    grep -qE ".*\[direct:\s?\S+\]\s+admin>\s+true" <<<"$result"
 }
 
 ########################
--- a/bitnami/mongodb-sharded/7.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb.sh
+++ b/bitnami/mongodb-sharded/7.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb.sh
@@ -1052,7 +1052,7 @@ mongodb_is_primary_node_up() {
 db.isMaster().ismaster
 EOF
     )
-    grep -q "\[direct: primary\] admin> true" <<<"$result"
+    grep -q "\[direct: \S\+\\] admin> true" <<<"$result"
 }
 
 ########################
--- a/bitnami/mongodb-sharded/7.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb.sh
+++ b/bitnami/mongodb-sharded/7.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb.sh
@@ -1052,7 +1052,7 @@ mongodb_is_primary_node_up() {
 db.isMaster().ismaster
 EOF
     )
-    grep -q "true" <<<"$result"
+    grep -q "\[direct: primary\] admin> true" <<<"$result"
 }
 
 ########################
--- a/bitnami/mongodb-sharded/7.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb.sh
+++ b/bitnami/mongodb-sharded/7.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb.sh
@@ -1052,7 +1052,7 @@ mongodb_is_primary_node_up() {
 db.isMaster().ismaster
 EOF
     )
-    grep -q "\[direct: \S\+\\] admin> true" <<<"$result"
+    grep -q "\[direct: \S\+\] admin> true" <<<"$result"
 }
 
 ########################


gongomgra:
Hi @esasidharan,

I have reproduced the issue you mention here, but the information used to check the availability of the primary node in my case (default docker-compose file without further changes) is slightly different to the one you are trying to check with your updated code

```
# updated code to print "result" information
root@10f405f5a82b:/# . /opt/bitnami/scripts/libmongodb.sh
root@10f405f5a82b:/# BITNAMI_DEBUG=true mongodb_is_primary_node_up mongodb-sharded 27017 root password123
mongodb 11:34:33.48 DEBUG ==> Validating mongodb-sharded as primary node...
mongodb 11:34:34.29 DEBUG ==> Current Mongosh Log ID:   65c0c7c93e264d7638ebcb47
Connecting to:          mongodb://<credentials>@mongodb-sharded:27017/admin?directConnection=true&appName=mongosh+2.1.3
Using MongoDB:          7.0.5
Using Mongosh:          2.1.3

For mongosh info see: https://docs.mongodb.com/mongodb-shell/

[direct: mongos] admin> true
[direct: mongos] admin>
```

Can you provide us with more information on your testing scenario and also update the grep command to check a more generic, yet still valid, string?

esasidharan:
Hi @gongomgra ,
    Thanks for testing and reproducing the failing issue case with default bitnami code.

My setup :
Connecting to:      mongodb://127.0.0.1:27017/admin?directConnection=true&serverSelectionTimeoutMS=2000&appName=mongosh+2.1.1
Using MongoDB:      7.0.5
Using Mongosh:      2.1.1

    In my setup, it's trying to connect to mongod shard instance from the same local pod as the shard instance is running.

    I'm not sure why you are getting the string [direct: mongos], whereas I get the string [direct: primary] in case of primary and [direct: secondary] in case of secondary replica set member. Are you trying to connect to mongod instance from a mongosh pod ? Or is the Mongosh version difference (2.1.3 in your case and 2.1.1 in my case) the reason ?

    Anyways, we can use a more generic grep like following to cover any string after [direct: 

grep -q "\[direct: \S\+\] admin> true" <<<"$result"

    Can you please check with this grep and see if it works for you ?

Thanks,
Sasi

esasidharan:
Hi @gongomgra ,
     My docker-compose.yaml is exactly the same as the one in the bitnami container repository.

https://github.com/bitnami/containers/blob/main/bitnami/mongodb-sharded/7.0/debian-11/docker-compose.yml

Thanks,
Sasi

gongomgra:
Hhi @esasidharan,

Thanks for your message. I got the same message you mentioned with the default docker-compose file. I originally used that file without any change, but I didn't check the messages during the initialization process. Instead, I `docker exec` to one of the nodes, loaded the libraries, and executed the `mongodb_is_primary_node_up` method manually, that's why I was getting another message (from manual execution of the function). 

Having this additional scenario into account (someone may want to run additional checks on their own), I believe we should better go with a generic check to be on the safe side. I have tested your suggestion, but I don't get it to work. I had to include an extra `.*` at the beginning of the line, plus also allowing `extended-regexp` features. Can you give it a try?

```
# cfg node
debian-11-mongodb-cfg-1      | mongodb 11:24:19.86 DEBUG ==> Current Mongosh Log ID:    65c9ffe30679116476222e37
debian-11-mongodb-cfg-1      | Connecting to:           mongodb://127.0.0.1:27017/admin?directConnection=true&serverSelectionTimeoutMS=2000&appName=mongosh+2.1.3
debian-11-mongodb-cfg-1      | Using MongoDB:           7.0.5
debian-11-mongodb-cfg-1      | Using Mongosh:           2.1.3
debian-11-mongodb-cfg-1      |
(...)
debian-11-mongodb-cfg-1      |
debian-11-mongodb-cfg-1      | cfgreplicaset [direct: primary] admin> true
debian-11-mongodb-cfg-1      | cfgreplicaset [direct: primary] admin>

# shard0 node
debian-11-mongodb-shard0-1   | mongodb 11:24:19.97 DEBUG ==> Current Mongosh Log ID:    65c9ffe35d1e614684af32f1
debian-11-mongodb-shard0-1   | Connecting to:           mongodb://127.0.0.1:27017/admin?directConnection=true&serverSelectionTimeoutMS=2000&appName=mongosh+2.1.3
debian-11-mongodb-shard0-1   | Using MongoDB:           7.0.5
debian-11-mongodb-shard0-1   | Using Mongosh:           2.1.3
debian-11-mongodb-shard0-1   |
debian-11-mongodb-shard0-1   | For mongosh info see: https://docs.mongodb.com/mongodb-shell/
debian-11-mongodb-shard0-1   |
(...)
debian-11-mongodb-shard0-1   |
debian-11-mongodb-shard0-1   | shard0 [direct: primary] admin> true
debian-11-mongodb-shard0-1   | shard0 [direct: primary] admin>
```

Checking outputs above, I updated the grep directive to the one below

```
grep -qE ".*\[direct:\s?\S+\]\s+admin>\s+true" <<<"$result"
```

bitnami/mongodb-sharded/7.0/debian-11/rootfs/opt/bitnami/scripts/libmongodb.sh
```diff
@@ -1052,7 +1052,7 @@ mongodb_is_primary_node_up() {
 db.isMaster().ismaster
 EOF
     )
-    grep -q "true" <<<"$result"
+    grep -q "\[direct: \S\+\] admin> true" <<<"$result"
```
gongomgra, line 1055:
```suggestion
    grep -qE ".*\[direct:\s?\S+\]\s+admin>\s+true" <<<"$result"
```

gongomgra:
@esasidharan,

I have contributed the PR with my proposal. Please give it a try and let me know if it works for you. Thanks.

esasidharan:
Hi @gongomgra ,
    Thanks for the information regarding the manual try. Yes, you are correct that it's better to have a general solution as you have suggested.
    
     I tested your solution and it's working for me as well, and hence we can use your solution. (Both my solution and your solution is working for me.)

Difference between basic and extended reg-exp : https://zerodayhacker.com/basic-or-extended-regex/
Basic reg-exp : use \ before special characters (Eg : \?, \+)
Extended reg-exp : don't use \ before special characters (Eg : ?, +)

     Hence, both ways it should.

     I'm not sure of why .* is making a difference. It's working for me both with and without .*

Thanks,
Sasi

gongomgra:
Hi @esasidharan,

Sorry for the delay. We are more used to extended regexp than marking special characters with backslashes. Regarding the `.*` at the beginning, I think it is more descriptive to mention there is something else in the line before the part of interest.

gongomgra:
@esasidharan I checked this PR with other team members and we are fine with the changes. Do you plan to push any other change or we can merge it? Thanks!

esasidharan:
Hi @gongomgra ,
     Thanks so much for helping to review this PR with valuable inputs. I don't have any changes. We can merge this PR.

Thanks,
Sasi

gongomgra (approved):

gongomgra:
@esasidharan, thanks for your contribution. We are merging your changes

PR #56617, opened 2024-02-08
[bitnami/reportserver] Add deprecation notice

Mauraza:
### Description of the change
Add deprecation notice

### DIFF
--- a/bitnami/reportserver/README.md
+++ b/bitnami/reportserver/README.md
@@ -1,5 +1,9 @@
 # Bitnami package for ReportServer Community
 
+## Deprecation Notice
+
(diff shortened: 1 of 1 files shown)


bitnami/reportserver/README.md
```diff
@@ -1,5 +1,9 @@
 # Bitnami package for ReportServer Community
 
+## Deprecation Notice
+
+The ReportServer container is no longer maintained by upstream and is now deprecated. This image will no longer be released in our catalog, but already released container images will still persist in the registries.
```
carrodher, line 5:
```suggestion
The ReportServer container will no longer be released in our catalog, but already released container images will still persist in the registries.
```

In this case, the reason is not the asset deprecated upstream but an internal decision to keep the catalog fresh

carrodher (changes requested):

carrodher (approved):

PR #62906, opened 2024-02-21
[bitnami/*] Remove debian 11 distro

dgomezleon:
### Description of the change

Since debian-12 is now the default distro, this PR removes debian-11 references:
* Remove `*/*/debian-11` folders
* Remove `os-shell/11` branch
* Remove `python/3.7` since it is deprecated
* Remove MongoDB and MongoDB Sharded 5.0 and 6.0 since they do not support debian-12
* Remove Ruby 3.0 since it does not support debian-12

PS: It keeps `openldap/*/debian-11` as we are in the process of adding debian-12 support.


carrodher (approved):
LGTM

acastong:
The removal of support of MongoDB 5 and 6 can be problematic for products that have supported software running on these versions. So products that are still using MongoDB 5 or 6 will no longer have access to security updates using bitnami containers until a window is available to make a migration to MongoDB 7.  Considering we need to support previous product releases for some time (on previous MongoDB version), we typically need security updates for these older MongoDB versions even if the latest software is running on latest MongoDB.

For deployments that were still on MongoDB 5: MongoDB version can only be increased by one major increment while keeping the persistent data. Doing two increments from 5 to 6 and then 6 to 7 is also not possible because then that would prevent a rollback in case of problem, so it also limits how quickly we can move to 7...

The goal of this post is simply to share that removing support of previous MongoDB versions does have an impact for users relying on bitnami MongoDB container images.

PR #55278, opened 2024-01-22
[bitnami/cassandra] Add possibility to update cassandra.yaml from the environment variables

dpavlov-smartling:
### Description of the change

Changes allow updating cassandra.yaml configuration file from environment variables using the same syntax as already used for CASSANDRA_CFG_ENV_, etc.

### Benefits
Currently, you can update only a few, pre-configured parameters of cassandra.yaml. For example, I needed to set parameterinternode_compression to all, but that can't be done via an environment variable. So, instead of adding a single environment variable for this particular parameter, I added the possibility to update multiple parameters via environment variables.

### Possible drawbacks

TBD
### Applicable issues
TBD

### Additional information

Current limitations:
* can't update commented parameters in cassandra.yaml. This is because the internal function `replace_in_file` uses regex that requires a match from the beginning of the string
* can't update nested yaml parameters. The reason is the same as for the previous limitation

PR #63358, opened 2024-02-23
[bitnami/cassandra] Add possibility to update cassandra.yaml from the environment variables

dpavlov-smartling:
@CeliaGMqrz  This is the same PR as https://github.com/bitnami/containers/pull/55278
I didn't realize how to properly solve conflict without dropping original commits, so, created this new pull request.
Sorry for the inconvenience. 

Description of the change
Changes allow updating cassandra.yaml configuration file from environment variables using the same syntax as already used for CASSANDRA_CFG_ENV_, etc.

Benefits
Currently, you can update only a few, pre-configured parameters of cassandra.yaml. For example, I needed to set parameterinternode_compression to all, but that can't be done via an environment variable. So, instead of adding a single environment variable for this particular parameter, I added the possibility to update multiple parameters via environment variables.

Possible drawbacks
TBD

Applicable issues
TBD

Additional information
Current limitations:

can't update commented parameters in cassandra.yaml. This is because the internal function replace_in_file uses regex that requires a match from the beginning of the string
can't update nested yaml parameters. The reason is the same as for the previous limitation


juan131 (commented):
Hi @dpavlov-smartling 

Thanks so much for proposing this solution!

The main disadvantage I identify on this approach is that it's complex and error-prone since it requires the user to do the translation between config key/value properties to the equivalent environment variable name/value pair. Also, it doesn't supported nested config properties. 

We already support providing any kind of customization on the Cassandra configuration file by mounting a custom config file as it's described in the link below:

- https://github.com/bitnami/containers/tree/main/bitnami/cassandra#configuration-file

Isn't this approach simpler?

dpavlov-smartling:
Hello @juan131 ,

This is the same approach that this docker image uses to update other 3 files - https://github.com/bitnami/containers/blob/d0d2c7b39108535b1cc3c75fe2f6c253c7463412/bitnami/cassandra/README.md?plain=1#L239C1-L241C121
So, my pool request uses this as an example and that is why it was implemented in that way.
And no, mounting the whole `cassandra.yaml` file isn't easier. To do this you need to generate `cassandra.yaml` file and cover all the cases, which docker image already covers like updated hostname, seeds, SSL configuration, etc.
In my particular case I needed only to change 1 option and updating the Environment variable is much easier than generating and mounting the whole `cassandra.yaml`
Hope this clarifies things and helps speed up the merging of the PR
Thanks,

juan131:
Hi @dpavlov-smartling 

Thanks for the details. I think this is good-to-have feature but we should find a solution for nested config properties before accepting it. Once we have it, I think we're ready to accept it.

dpavlov-smartling:
Hello @juan131 ,

I am sorry for the confusion. It does update nested parameters. I used your image default function to update yaml parameters `cassandra_yaml_set`. You can find its code here -  https://github.com/bitnami/containers/blob/7766faad45aa1beff1910417b06985d3edc29f54/bitnami/cassandra/4.0/debian-12/rootfs/opt/bitnami/scripts/libcassandra.sh#L29C1-L40C2
It is used all around in the image, including for the updating nested parameters.

Thanks

dpavlov-smartling:
Removed warning about nested parameters from the readme file.
@juan131 I hope this helps.
Thanks,

juan131:
Thanks so much!

@dpavlov-smartling could you add a small example in the README about how to use this new way to adapt the configuration? It'd be nice if you could in the example the use of env. vars to customize nested properties.

dpavlov-smartling:
@juan131 
I have updated readme file with examples.

juan131 (approved):
LGTM

juan131:
Hi @dpavlov-smartling 

Thanks so much for your contribution! Please expect a new image to be released within the next 24 hours including your changes.

juan131:
Hi @dpavlov-smartling 

New images (*4.0.12-debian-12-r8* and *4.1.4-debian-12-r4*) have been released including your changes.

PR #63461, opened 2024-02-29
[bitnami/wordpress] Deflate javascript with mime-type text/javascript

ametad:
### Description of the change

Configuration of the Apache webserver to include the mime-type `text/javascript` for compressing the response.

### Benefits

Javascript will be smaller and therefore loaded more quickly.

### Applicable issues

- fixes #

### Additional information

Servers should use text/javascript for JavaScript resources, in accordance with Updates to ECMAScript Media Types. Servers should not use other JavaScript MIME types for JavaScript resources, and must not use non-JavaScript MIME types. [RFC9239](https://www.rfc-editor.org/rfc/rfc9239)

This is the second Pull Request for the same thing. But this time I used "signed-off" commits.

### DIFF
--- a/bitnami/wordpress/6/debian-12/rootfs/opt/bitnami/apache/conf/deflate.conf
+++ b/bitnami/wordpress/6/debian-12/rootfs/opt/bitnami/apache/conf/deflate.conf
@@ -1,5 +1,5 @@
 <IfModule mod_deflate.c>
-  AddOutputFilterByType DEFLATE text/html text/plain text/xml text/css
+  AddOutputFilterByType DEFLATE text/html text/plain text/xml text/css text/javascript
   AddOutputFilterByType DEFLATE application/x-javascript application/javascript application/ecmascript
   AddOutputFilterByType DEFLATE application/rss+xml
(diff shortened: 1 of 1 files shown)


ametad:
The [first attempt](https://github.com/bitnami/containers/pull/63460) had no signed-off commit. So I made a new PR, this one.

ametad:
The reason why to also include mime-type `text/javascript`, is found here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#textjavascript

In a nutshell, it says:

> Note: Even though any given [user agent](https://developer.mozilla.org/en-US/docs/Glossary/User_agent) may support any or all of these, you should only use text/javascript. It's the only MIME type guaranteed to work now and into the future.

ametad:
Hello @migruiz4 , is there something I can do perhaps to get the review approved? I am happy to help.

migruiz4 (approved):
I'm sorry for the late resposne @ametad.

Changes look good to me! A new release of the bitnami/wordpress container image should be released soon including your changes

PR #63498, opened 2024-03-02
remove duplicates from README

kubealex:
Remove duplicates:
- MINIO_DATA_DIR
- MINIO_API_PORT_NUMBER (confusing as it had two values)
### Benefits

Avoids confusion.


migruiz4 (approved):
Thank you for your contribution @kubealex and sorry for the late response.

The env variables tables are automatically generated and we had to address an internal issue to permanently fix this, otherwise, your change would have been overridden by the next release.

We fixed the issue and duplicated entries should be fixed in MinIO and all the other container READMEs affected by this issue.

PR #63706, opened 2024-03-07
[bitnami/keycloak] Add KEYCLOAK_CREATE_ADMIN_USER to list of configs documented in README

benelot:
### Description of the change

I added KEYCLOAK_CREATE_ADMIN_USER to the README. I did not get an ADMIN USER from the container by default, so I guess the default for that is false. This config is part of the original keycloak container.

### Benefits

User is informed how to get an admin user.

### Possible drawbacks

None.


fevisera:
Hi @benelot,

Thank you for creating the PR. Please, note that `KEYCLOAK_CREATE_ADMIN_USER` was deleted some time ago as mentioned in
https://github.com/bitnami/charts/issues/10715#issuecomment-1161495230.

Could you provide more details on why you cannot get an admin user by running the default container?

I got the following using the [docker-compose.yml](https://github.com/bitnami/containers/blob/main/bitnami/keycloak/docker-compose.yml) which does not seem to have any additional configuration for keycloak container:
```
$ cd bitnami/keycloak 
$ docker-compose up -d
```
image

PR #63790, opened 2024-03-13
[bitnami/elasticsearch] Support configuring environment variables

### LINKED ISSUE #63477: [bitnami/elasticsearch] Support customizing Elasticsearch config file path
### Name and Version

docker.io/bitnami/elasticsearch (all versions)

I'm using [bitnami/elasticsearch](https://artifacthub.io/packages/helm/bitnami/elasticsearch) Helm chart and I want to configure [readOnlyRootFilesystem](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) for security. To do so, any dir that needs write access should be mounted as an empty dir volume.

The issue is that the dir `/opt/bitnami/elasticsearch/config` has files shipped with the image in addition to ES config file `/opt/bitnami/elasticsearch/config/elasticsearch.yml` which auto-generated at the startup.

For that reason, the config file path should be customizable to set it to a writable dir.

Allow to set the value of the var `ELASTICSEARCH_CONF_FILE` here:
https://github.com/bitnami/containers/blob/4a21bf3/bitnami/elasticsearch/8/debian-12/rootfs/opt/bitnami/scripts/elasticsearch-env.sh#L127

From:
```
export ELASTICSEARCH_CONF_FILE="${DB_CONF_DIR}/elasticsearch.yml"
```

To:
```
export ELASTICSEARCH_CONF_FILE="${ELASTICSEARCH_CONF_FILE:-$DB_CONF_DIR/elasticsearch.yml}"
```

_No response_

aabouzaid:
### Description of the change

Allow to configure the Elasticsearch environment variables.

### Benefits

Support setting some paths env vars which allow enabling security best practices like enabling `readOnlyRootFilesystem`.   

### Possible drawbacks

I've changed all `ELASTICSEARCH_*` vars to be customizable but it's not clear to me if there is a technical reason why some vars are read-only (not clear if it's just a matter of support or there is a technical limitation).

### Applicable issues

- fixes https://github.com/bitnami/containers/issues/63477

### Additional information

There is no behavior change introduced in this PR, only the ability to customize the vars without changing in the default values.


migruiz4:
Hi @aabouzaid,

I'm sorry, but there is no point in customizing those env variables because changing them will only cause the container to fail.
Most paths and settings, such as the 'ELASTICSEARCH_USERNAME', can not be customized; therefore, providing a different value will only cause the container to fail.

> Support setting some paths env vars which allow enabling security best practices like enabling readOnlyRootFilesystem.

I don't see the relation between making all the env variables writable and `readOnlyRootFilesystem`. This image is already used in the `bitnami/elasticsearch` chart, which has been updated to run using `readOnlyRootFilesystem: true` (https://github.com/bitnami/charts/pull/24303)

aabouzaid:
@migruiz4 I'm using an older version of the ES so I cannot use the latest chart.
My idea was to make the fix in the Docker container and release it as a patch.

But by looking at the container again, my change would help anyway since all of these changes are for the newer image.

So I will also close the main issue https://github.com/bitnami/containers/issues/63477

PR #64098, opened 2024-03-24
[bitnami/mediawiki] improve configuration for reverse proxy configs

### LINKED ISSUE #64097: [bitnami/mediawiki] 404 errors when overriding MEDIAWIKI_WIKI_PREFIX
### Name and Version

bitnami/mediawiki:1

amd64

set a non-default value for the prefix in the docker compose `environment`:
```
- MEDIAWIKI_WIKI_PREFIX=/something
```
then run `docker compose up`

MediaWiki is accessible at `/something` instead of `/wiki`

### What do you see instead?

Every page under `/something` returns a 404 error:
```
mediawiki-1  | mediawiki 22:25:41.93 INFO  ==> Running MediaWiki install script
mediawiki-1  | mediawiki 22:25:43.06 INFO  ==> Configuring MediaWiki settings
mediawiki-1  | mediawiki 22:25:43.07 INFO  ==> Setting MediaWiki short URLs
mediawiki-1  | mediawiki 22:25:43.08 INFO  ==> Persisting MediaWiki installation
mediawiki-1  | mediawiki 22:25:44.95 INFO  ==> ** MediaWiki setup finished! **
mediawiki-1  |
mediawiki-1  | mediawiki 22:25:44.95 INFO  ==> ** Starting Apache **
mediawiki-1  | [Sat Mar 23 22:25:45.180354 2024] [mpm_prefork:notice] [pid 1] AH00163: Apache/2.4.58 (Unix) OpenSSL/3.0.11 configured -- resuming normal operations
mediawiki-1  | [Sat Mar 23 22:25:45.180392 2024] [core:notice] [pid 1] AH00094: Command line: '/opt/bitnami/apache/bin/httpd -f /opt/bitnami/apache/conf/httpd.conf -D FOREGROUND'
mediawiki-1  | 192.168.160.1 - - [23/Mar/2024:22:26:06 +0000] "GET / HTTP/1.1" 301 -
mediawiki-1  | 192.168.160.1 - - [23/Mar/2024:22:26:07 +0000] "GET /something/Main_Page HTTP/1.1" 404 196
mediawiki-1  | 192.168.160.1 - - [23/Mar/2024:22:26:07 +0000] "GET /favicon.ico HTTP/1.1" 404 196
```

### Additional information

_No response_

dunn:
several changes to improve behavior when running behind a reverse proxy:

- we need the `RewriteRule` in the vhost to match [`$wgArticlePath`](https://www.mediawiki.org/wiki/Special:MyLanguage/Manual:$wgArticlePath) in the application `LocalSettings.php`, which requires that it be modified at startup in order to respect custom values of `MEDIAWIKI_WIKI_PREFIX`.
- setting [`$wgScriptPath`](https://www.mediawiki.org/wiki/Special:MyLanguage/Manual:$wgScriptPath) to `""` is fine for the normal case, but if you're running behind a reverse proxy and want to serve MediaWiki on one path (e.g., `/wiki`) and other services on other paths, being able to set the script path as non-root makes proxy routing easier.
- setting [`$wgServer`](https://www.mediawiki.org/wiki/Special:MyLanguage/Manual:$wgServer) to `MEDIAWIKI_HOST` results in absolute URLs including the full hostname and port, instead of relative URLs, which breaks links when the reverse proxy is running on a different port. for the same reason, set [`$wgAssumeProxiesUseDefaultProtocolPorts`](https://www.mediawiki.org/wiki/Special:MyLanguage/Manual:$wgAssumeProxiesUseDefaultProtocolPorts) to false.

Fixes https://github.com/bitnami/containers/issues/64097


bitnami/mediawiki/1/debian-12/rootfs/opt/bitnami/scripts/mediawiki/setup.sh
```diff
@@ -32,7 +32,10 @@ fi
 mediawiki_validate
 
 # Update web server configuration with runtime environment (needs to happen before the initialization)
-web_server_update_app_configuration "mediawiki"
```
dunn, line 35: if we remove this here, i don't think `web_server_update_app_configuration` is used anywhere

dunn:
added a few commits, i was going to make separate PRs but these changes depend on each other. i'll update the PR description.

dunn:
anything i can do to help move this forward?

bitnami/mediawiki/1/debian-12/rootfs/opt/bitnami/scripts/mediawiki/setup.sh
migruiz4: Both `ensure_web_server_app_configuration_exists` and `web_server_update_app_configuration` serve different purposes.

`ensure_web_server_app_configuration_exists` initializes the Apache configuration at image build time, while `web_server_update_app_configuration` updates it at runtime to configure the server-specific host and port configured via env vars.

It is possible to replace `ensure_web_server_app_configuration_exists` from build-time to run-time, but you would need one of the following:
- Keep `web_server_update_app_configuration` after `ensure_web_server_app_configuration_exists`.
- Add args `--http-port ` and `--https-port` to `ensure_web_server_app_configuration_exists`.

I would personally recommend option 1.

bitnami/mediawiki/1/debian-12/rootfs/opt/bitnami/scripts/mediawiki/setup.sh
```diff
@@ -32,7 +32,11 @@ fi
...
+ensure_web_server_app_configuration_exists "mediawiki" --type php --apache-extra-directory-configuration "
+RewriteEngine On
+RewriteRule ^/?${MEDIAWIKI_WIKI_PREFIX:1}(/.*)?$ %{DOCUMENT_ROOT}/index.php
+RewriteRule ^/?${MEDIAWIKI_SCRIPT_PATH:1}(.*)$ %{DOCUMENT_ROOT}/\$1 [L]
```
migruiz4, line 38: Are `${MEDIAWIKI_WIKI_PREFIX:1}` and `${MEDIAWIKI_SCRIPT_PATH:1}` typos?

What would be the reason to slice the first character?

bitnami/mediawiki/1/debian-12/rootfs/opt/bitnami/scripts/mediawiki-env.sh
```diff
@@ -81,6 +82,7 @@ export MEDIAWIKI_DATA_TO_PERSIST="${MEDIAWIKI_DATA_TO_PERSIST:-images extensions
 export MEDIAWIKI_SKIP_BOOTSTRAP="${MEDIAWIKI_SKIP_BOOTSTRAP:-}" # only used during the first initialization
 export MEDIAWIKI_WIKI_NAME="${MEDIAWIKI_WIKI_NAME:-Bitnami MediaWiki}" # only used during the first initialization
 export MEDIAWIKI_WIKI_PREFIX="${MEDIAWIKI_WIKI_PREFIX:-/wiki}" # only used during the first initialization
+export MEDIAWIKI_SCRIPT_PATH="${MEDIAWIKI_SCRIPT_PATH:-/w}" # only used during the first initialization
```
migruiz4, line 85: Why the default value `MEDIAWIKI_SCRIPT_PATH=/w`?

If possible, I would recommend this feature be an enhancement to available settings instead of modifying the current default feature.

bitnami/mediawiki/1/debian-12/rootfs/opt/bitnami/scripts/libmediawiki.sh
```diff
@@ -317,14 +322,14 @@ mediawiki_configure_host() {
...
+      [[ "$MEDIAWIKI_EXTERNAL_HTTP_PORT_NUMBER" != "80" ]] && url+=":${MEDIAWIKI_EXTERNAL_HTTP_PORT_NUMBER}"
     fi
-    mediawiki_conf_set "\$wgServer" "$url"
 
+    mediawiki_conf_set "\$wgServer" ""
+    # these keys aren't included in the default LocalSettings
+    cat >> "$MEDIAWIKI_CONF_FILE" <<EOF
+\$wgCanonicalServer = "$url";
+\$wgEnableCanonicalServerLink = true;
+EOF
```
migruiz4, line 334: Could you please provide more context about these changes? Why setting `$wgServer` to an empty string and removing the protocol-relative conditional? (`url="//${host}"`)

According to [mediawiki documentation](https://www.mediawiki.org/wiki/Manual:$wgServer):
> Since version 1.18.0 MediaWiki also supports setting $wgServer to a protocol-relative URL (e.g., //www.mediawiki.org). This is used for supporting both HTTP and HTTPS with the same caches by using links that work under both protocols. When doing this, [$wgCanonicalServer](https://www.mediawiki.org/wiki/Special:MyLanguage/Manual:$wgCanonicalServer) can be used to set the full URL including protocol that will be used in locations such as emails that do not support protocol relative URLs. If the URL starts with https://, MediaWiki will assume that your wiki prefers the HTTPS protocol or supports it exclusively. In that case, you should also set [$wgForceHTTPS](https://www.mediawiki.org/wiki/Special:MyLanguage/Manual:$wgForceHTTPS). See [Manual:HTTPS](https://www.mediawiki.org/wiki/Special:MyLanguage/Manual:HTTPS) for more information.

migruiz4 (commented):
Thank you for your contribution @dunn!

I have added several comments, please take a look at them.

Additionally, I would appreciate if you could remove the following section from the tests:
https://github.com/bitnami/containers/blob/b8c665b06776304658c497b495b77d979f5f42f8/.vib/mediawiki/goss/mediawiki.yaml#L40-L45

Because you are moving the Apache config creation from buildtime to runtime, the file no longer exists, causing the tests (and the CI/CD pipeline) to fail.

bitnami/mediawiki/1/debian-12/rootfs/opt/bitnami/scripts/mediawiki/setup.sh
dunn: the first character is the forward slash `/`, which is already present in the regexp

bitnami/mediawiki/1/debian-12/rootfs/opt/bitnami/scripts/mediawiki-env.sh
dunn: that was to match the upstream default: https://www.mediawiki.org/wiki/Manual:$wgScriptPath

bitnami/mediawiki/1/debian-12/rootfs/opt/bitnami/scripts/libmediawiki.sh
dunn: i'm setting `$wgServer` to an empty string since it's used by MediaWiki to generate links, which results in absolute URLs in the application interface instead of relative ones.  this adds needless complexity when configuring a reverse proxy, or just when running locally in development.

bitnami/mediawiki/1/debian-12/rootfs/opt/bitnami/scripts/mediawiki/setup.sh
migruiz4: Ah, that's true, I didn't notice the forward slash, thank you!

bitnami/mediawiki/1/debian-12/rootfs/opt/bitnami/scripts/mediawiki-env.sh
migruiz4: I'm sorry but I think I can not accept this PR unless the default change is reversed to an empty string, as it may cause issues with existing MediaWiki deployments because of the following would happen:

- Setting `$wgScriptPath` is only set during the first initialization, therefore it won't be updated for existing deployments.
- Because we moved the webserver configuration from build-time to run-time, it will be regenerated on each restart, meaning that upgrading will set these in the Apache configuration:
    ```
    RewriteRule ^/?${MEDIAWIKI_WIKI_PREFIX:1}(/.*)?$ %{DOCUMENT_ROOT}/index.php
    RewriteRule ^/?${MEDIAWIKI_SCRIPT_PATH:1}(.*)$ %{DOCUMENT_ROOT}/\$1 [L]
    ```
- Because $MEDIAWIKI_SCRIPT_PATH is set to default `/w`, it will end up in the RewriteRule, even if MediaWiki was configured with `$wgScriptPath=""` when the server was first deployed.

bitnami/mediawiki/1/debian-12/rootfs/opt/bitnami/scripts/libmediawiki.sh
migruiz4: Understood, thanks you for your comment!

migruiz4 (commented):
Hi @dunn,

Thank you for implementing my suggestions, please take a look at my comment regarding `MEDIAWIKI_SCRIPT_PATH`.

dunn:
@migruiz4 thanks for the review! I added a commit setting `MEDIAWIKI_SCRIPT_PATH` to `/` by default, restoring the existing behavior.

migruiz4 (approved):
Thank you very much for your contribution @dunn, looks good to me!

jotamartos:
Hi @dunn,

Thank you for your contribution. However, these changes are breaking the functionality of the container, you get several redirect loops when you try to access the app or work with it. This is preventing our test to pass and update the Chart solution. 

We are planing some changes to make the solution work again and they will be included in the next release:
- Recover the old logic to configure the $wgServer parameter in the conf file
- The $wgCanonicalServer parameter uses that configuration by default according to the official documentation: https://www.mediawiki.org/wiki/Manual:$wgCanonicalServer
- RewriteRules will be updated to use the ones mentioned in the official documentation https://www.mediawiki.org/wiki/Manual:Short_URL/wiki/Page_Title_--_.htaccess

If you continue running into issues when running the app behind a proxy, please do create another PR to update the configuration taking these changes into account.

jotamartos:
Changes were applied here: https://github.com/bitnami/containers/pull/67621

Issue #63545, opened 2024-03-05
[bitnami/keycloak] upgrade to 24.0.0


Jojoooo1:
### Name and Version

bitnami/keycloak

### What is the problem this feature will solve?

Upgrade to last Keycloak version 24.0.0

### What is the feature you are proposing to solve the problem?

Latest features and fixes from keycloak

carrodher:
Thank you for bringing this to our attention. Our automated test and release pipeline detected the new version; however, we have identified some issues during the packaging process that require manual intervention.

Our team is actively addressing these issues at the moment. We are committed to resolving them as quickly as possible. We will provide updates on this thread as soon as the new version is successfully released.

tanadeau:
See also https://github.com/adorsys/keycloak-config-cli/issues/996, as that not-yet-released version of `keycloak-config-cli` will need to be brought in as well.

tanadeau:
Are there any updates on this? We'd like to upgrade to Keycloak 24 for some CVE fixes among other reasons.

carrodher:
Thanks for your patience. Container images for Keycloak 24 (24.0.2) were released some hours ago, see https://hub.docker.com/r/bitnami/keycloak/tags

tanadeau:
Are images pushed to https://gallery.ecr.aws/bitnami/keycloak automatically? It looks like the new tag isn't available there yet.

carrodher:
There was a pending step which already succeeded, the image is now present in AWS Public Gallery and DockerHub

PR #64184, opened 2024-03-27
[bitnami/keycloak] Add KC_RUN_IN_CONTAINER env. var.

tanadeau:
### Description of the change

Adds the `KC_RUN_IN_CONTAINER` environment variable to `keycloak-env.sh`. This matches the change in https://github.com/keycloak/keycloak/pull/26661, where `kc.sh` now uses that to automatically set the heap options based on container resources.

### Benefits

Users no longer need to override `JAVA_OPTS`, etc. to set container heap options.

### Possible drawbacks

None known.

### Applicable issues

See also #63545 as Keycloak 24 is where support for this was added.

### Additional information

None.


rafariossaa:
Hi,
  Thanks for adding this. 
  I think this should be added to the rest of environment variables in [keycloack-env.sh](https://github.com/bitnami/containers/blob/main/bitnami/keycloak/24/debian-12/rootfs/opt/bitnami/scripts/keycloak-env.sh). Could you move it there ?

tanadeau:
@rafariossaa: I certainly can if desired. I saw that script, but it seemed to be focused on setting/defaulting the Bitnami-specific `KEYCLOAK_*` environment variables vs. variables that are used by upstream Keycloak scripts.

rafariossaa (approved):
Thanks for adding it.
LGTM

PR #64639, opened 2024-04-03
[bitnami/seaweedfs] Initial vib configuration

fmulero:
### Description of the change

Initial vib configuration for SeaweedFS container. This container is still in an early stage.

### Benefits

Just add basic configuration to enable linters and security scans.

### Possible drawbacks

None


andresbono (approved):
LGTM

PR #64646, opened 2024-04-03
[bitnami/seaweedfs] Configure tests

fmulero:
### Description of the change

Initial set of tests for SeaweedFS container.

### Benefits

Ensure binary file is running and version is correct.

### Possible drawbacks

None


.vib/seaweedfs/vib-verify.json
```diff
@@ -50,6 +50,21 @@
               "OS"
             ]
           }
+        },
+        {
+          "action_id": "goss",
```
andresbono, line 55: The default entrypoint/cmd of the container keeps it up, right? No need for custom runtime parameters?

andresbono (dismissed):
LGTM

.vib/seaweedfs/vib-verify.json
fmulero: Good catch! I forgot it

andresbono (approved):
LGTM

PR #65198, opened 2024-04-16
[bitnami/elasticsearch] Enclose IPv6 addresses in square brackets in the healthcheck URL.

prelz:
### Description of the change

The change adds a pair of square brackets around IPv6 addresses in the
URL used in the elasticsearch_healthcheck() function. The generated URL
is invalid otherwise, and the heath check will fail on IPv6-only hosts.

### Benefits

Ability to use the chart for an IPv6-only installation of elasticsearch.

### Possible drawbacks

This is a bug fix.

### Applicable issues

Didn't create an issue for this.

### DIFF
--- a/bitnami/elasticsearch/7/debian-12/rootfs/opt/bitnami/scripts/libelasticsearch.sh
+++ b/bitnami/elasticsearch/7/debian-12/rootfs/opt/bitnami/scripts/libelasticsearch.sh
@@ -909,6 +909,9 @@ elasticsearch_healthcheck() {
     local host
 
     host=$(get_elasticsearch_hostname)
+    if validate_ipv6 "$host"; then
+        host="[$host]"
+    fi
 
     if is_boolean_yes "$DB_ENABLE_SECURITY"; then
         command_args+=("-k" "--user" "${DB_USERNAME}:${DB_PASSWORD}")
--- a/bitnami/elasticsearch/7/debian-12/rootfs/opt/bitnami/scripts/libelasticsearch.sh
+++ b/bitnami/elasticsearch/7/debian-12/rootfs/opt/bitnami/scripts/libelasticsearch.sh
@@ -909,8 +909,8 @@ elasticsearch_healthcheck() {
     local host
 
     host=$(get_elasticsearch_hostname)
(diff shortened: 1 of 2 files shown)


bitnami/elasticsearch/7/debian-12/rootfs/opt/bitnami/scripts/libelasticsearch.sh
```diff
@@ -909,6 +909,9 @@ elasticsearch_healthcheck() {
     local host
 
     host=$(get_elasticsearch_hostname)
+    if validate_ipv6 "$host"; then
+        host="[$host]"
```
fmulero, line 913: Just a little suggestion. As good practice, we usually use `${var}` when calling variables inside strings, to make the code clearer and avoid substitutions or expansion errors . 

```suggestion
        host="[${host}]"
```

bitnami/elasticsearch/8/debian-12/rootfs/opt/bitnami/scripts/libelasticsearch.sh
```diff
@@ -909,6 +909,9 @@ elasticsearch_healthcheck() {
     local host
 
     host=$(get_elasticsearch_hostname)
+    if validate_ipv6 "$host"; then
+        host="[$host]"
```
fmulero, line 913: Same here

```suggestion
        host="[${host}]"
```

fmulero (commented):
Great! thanks a lot @prelz, Good catch!

I've just left a little suggestion, Could you take it a glance?

bitnami/elasticsearch/8/debian-12/rootfs/opt/bitnami/scripts/libelasticsearch.sh
prelz: Should make no difference in case of a validated IPv6 address, but I added the braces. I added them in the previous
line for consistency as well.

fmulero: Thanks

fmulero (approved):
LGTM

PR #65835, opened 2024-05-03
feat:  mysql_upgrade is not needed when the version greater than 8.0.16

ayhanaltunkaynak:
https://dev.mysql.com/doc/refman/8.0/en/replication-upgrade.html#:~:text=If%20you%20have%20upgraded%20to%20MySQL%208.0.16%20or%20later%2C%20do%20not%20invoke%20mysql_upgrade.%20From%20that%20release%2C%20MySQL%20Server%20performs%20the%20entire%20MySQL%20upgrade%20procedure%2C%20disabling%20binary%20logging%20during%20the%20upgrade.

### Applicable issues

- fixes #

### DIFF
--- a/bitnami/mysql/8.0/debian-12/rootfs/opt/bitnami/scripts/libmysql.sh
+++ b/bitnami/mysql/8.0/debian-12/rootfs/opt/bitnami/scripts/libmysql.sh
@@ -825,7 +825,7 @@ mysql_upgrade() {
     if [[ "$DB_FLAVOR" = *"mysql"* ]] && [[
         "$major_version" -gt "8"
         || ( "$major_version" -eq "8" && "$minor_version" -gt "0" )
-        || ( "$major_version" -eq "8" && "$minor_version" -eq "0" && "$patch_version" -ge "16" )
+        || ( "$major_version" -eq "8" && "$minor_version" -eq "0" && "$patch_version" -le "16" )
     ]]; then
         mysql_stop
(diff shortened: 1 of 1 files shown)


aairey (approved):

jotamartos:
Hi @ayhanaltunkaynak,

Thanks for your contribution. I just reviewed the current logic and it looks good to me, it doesn't require changes. It works this way:

```
if version > 8.x.x OR version > 8.0.x OR version >= 8.0.16
then stop and start MySQL
else start MySQL and run the upgrade command
```

The condition is true with all versions equal or higher than 8.0.16. The else block is run when using a lower version. 

Is there something I'm missing? Did you run into any issue when using the container? 

Thanks

jotamartos:
With your changes, the script will try to use mysql_upgrade in versions higher than 8.0.16 and that's not supported anymore according to the documentation you shared. 

Apart from that, the upgrade flag is a valid one when performing an upgrade as you can see [here](https://dev.mysql.com/doc/refman/8.0/en/upgrading-what-is-upgraded.html). 

I'm going to close this PR as this change will break the upgrade process but please do not hesitate to create a new one updating the upgrade process if required.

ayhanaltunkaynak:
A message has been received stating that a server upgrade from '8.0.36' to '8.0.36' has begun. Consequently, MySQL is restarting. While we thought that this code block was the issue, we have discovered that the problem may be something else. We are currently investigating the matter in more detail.

PR #66232, opened 2024-05-07
[bitnami/ghost] Update Ghost image with SMTP ENV fixes

### LINKED ISSUE #20871: GHOST_EMAIL is used for "mail.from" instead of checking GHOST_SMTP_FROM_ADDRESS
### Name and Version

bitnami/ghost:5.30.1

1. Define `GHOST_SMTP_FROM_ADDRESS` as "noreply@example.com" but `GHOST_EMAIL` as "real_user@example.com"
2. Check config.production.json

`GHOST_SMTP_FROM_ADDRESS` should be used instead of `GHOST_EMAIL` for `mail.from` setting if exists. Some mailers are critically require an equality of `mail.from` and `mail.options.auth.user` fields.

### What do you see instead?

`GHOST_SMTP_FROM_ADDRESS` is not used

### Additional information

See [here](https://github.com/bitnami/containers/blob/main/bitnami/ghost/5/debian-11/rootfs/opt/bitnami/scripts/libghost.sh#L310) the according code

### LINKED ISSUE #20872: Ghost "mail.options.secure" property instead of "mail.options.secureConnection"
### Name and Version

bitnami/ghost:5.30.1

1. Check documentation of Ghost [here](https://ghost.org/docs/config/#mail)
2. It uses `secure` property to define secured connection for SMTP setup

Configuration is consistent with tool documentation

### What do you see instead?

Bitnami generates property `secureConnection` which is not by Ghost documentation

### Additional information

I tried to dive deeper into [smtp-connection](https://github.com/nodemailer/nodemailer/blob/master/lib/smtp-connection/index.js#L55) which is used by Ghost. It uses `secure` property from the constructor arguments but stores it in the instance property called `secureConnection`. Not sure honestly, maybe Ghost can do reassignments later to save compatibility with previous `secureConnection` naming. But documentation contains `secure`.

Just-Insane:
Fixes some SMTP ENV configuration

Fixes:
#20872 
#20871 

Signed-off-by: Jax Gauthier <jax@gauthier.id>

### Description of the change

Fixes the following environment variables:

Adds `GHOST_SMTP_FROM_ADDRESS` to set the Ghost FROM Address, rather than using `GHOST_EMAIL`

Fixes SMTP configuration for secure connection - Ghost docs now use `mail.options.secure` rather than `mail.options.secureConnection`

### Benefits

Resolves two outstanding issues with SMTP configuration in Ghost, allowing to properly set the SMTP From Address and using secure connections.

### Possible drawbacks

It might be good to mention in the documentation that `mail.options.secure` only applies when the SMTP port is set to 456, per the Ghost Docs.

### Applicable issues

- fixes #20872
- fixes #20871 

### Additional information

Tested by building the image and deploying - confirmed the changes are working as expected.

Note that these changes only work on new installs (initial setup) - recommend exporting settings and posts via Admin panel and importing after re-deploy.


bitnami/ghost/5/debian-12/rootfs/opt/bitnami/scripts/ghost-env.sh
```diff
@@ -108,6 +109,8 @@ export GHOST_EMAIL="${GHOST_EMAIL:-user@example.com}" # only used during the fir
 # Ghost SMTP credentials
 GHOST_SMTP_HOST="${GHOST_SMTP_HOST:-"${SMTP_HOST:-}"}"
 export GHOST_SMTP_HOST="${GHOST_SMTP_HOST:-}" # only used during the first initialization
+GHOST_SMTP_FROM_ADDRESS="${GHOST_SMTP_FROM_ADDRESS:-"${SMTP_FROM:-}"}"
+GHOST_SMTP_FROM_ADDRESS="${GHOST_SMTP_FROM_ADDRESS:-"${GHOST_SMTP_FROM_ADDRESS:-}"}"
```
jotamartos, line 113: I think we can keep using "GHOST_EMAIL" as default value and modify it if the user provides a different one. Sounds good? 

```suggestion
GHOST_SMTP_FROM_ADDRESS="${GHOST_SMTP_FROM_ADDRESS:-"${GHOST_EMAIL:-}"}"
```

jotamartos (commented):
Thanks for your contribution. Could you please check my comment?

bitnami/ghost/5/debian-12/rootfs/opt/bitnami/scripts/ghost-env.sh
Just-Insane: Yeah, that makes sense. Removes a breaking change and ensures a sane default.

jotamartos (approved):
The next revision of the container will include these changes. Thanks

PR #67011, opened 2024-05-14
[bitnami/flink] update scripts how configuration is handled

### LINKED ISSUE #67010: flink issue with configuration handling
### Name and Version

bitnami/flink:1.19.0-debian-12-r5

amd64

This approach for configuration was valid earlier:
https://github.com/bitnami/containers/blob/fe7ac75eb9e200ed06efeeb6a11bdef77f922cd7/bitnami/flink/1/debian-12/rootfs/opt/bitnami/scripts/flink/entrypoint.sh#L21C1-L26C1

but when Flink 1.19 moved to YAML config there is issue with this approach.

[flink 1.19](https://nightlies.apache.org/flink/flink-docs-master/release-notes/flink-1.19/#support-standard-yaml-for-flink-configuration) which has some major changes how configuration is handled. Under [link](https://nightlies.apache.org/flink/flink-docs-release-1.19/docs/deployment/config/#flink-configuration-file) we have major behaviour change:

```
Duplicated keys:

flink-conf.yaml: Allows duplicated keys and takes the last key-value pair for the corresponding key that appears in the file.
config.yaml: Does not allow duplicated keys, and an error will be reported when loading the configuration.
```

so now preserving config.yaml file and reruning entrypoint can cause configuration corruption caused by duplicate keys.

Error:
error:
```sh
flink 07:20:57.96 INFO  ==> ** Starting Apache Flink Task Manager

[ERROR] The execution result is empty.
[ERROR] Could not get JVM parameters and dynamic configurations properly.
[ERROR] Raw output from BashJavaUtils:
WARNING: sun.reflect.Reflection.getCallerClass is not supported. This will impact performance.
INFO  [] - Using standard YAML parser to load flink configuration file from /opt/bitnami/flink/conf/config.yaml.
ERROR [] - Failed to parse YAML configuration
org.snakeyaml.engine.v2.exceptions.YamlEngineException: while constructing a mapping
 in reader, line 21, column 1
found duplicate key taskmanager.memory.process.size
 in reader, line 313, column 1

        at org.snakeyaml.engine.v2.constructor.StandardConstructor.processDuplicateKeys(StandardConstructor.java:90) ~[flink-dist-1.19.0.jar:1.19.0]
        at org.snakeyaml.engine.v2.constructor.StandardConstructor.flattenMapping(StandardConstructor.java:70) ~[flink-dist-1.19.0.jar:1.19.0]
        at org.snakeyaml.engine.v2.constructor.StandardConstructor.constructMapping2ndStep(StandardConstructor.java:119) ~[flink-dist-1.19.0.jar:1.19.0]
        at org.snakeyaml.engine.v2.constructor.BaseConstructor.constructMapping(BaseConstructor.java:278) ~[flink-dist-1.19.0.jar:1.19.0]
        at org.snakeyaml.engine.v2.constructor.StandardConstructor$ConstructYamlMap.construct(StandardConstructor.java:203) ~[flink-dist-1.19.0.jar:1.19.0]
```

Reason of such problem is flag `-n` in  https://github.com/bitnami/containers/blob/fe7ac75eb9e200ed06efeeb6a11bdef77f922cd7/bitnami/flink/1/debian-12/rootfs/opt/bitnami/scripts/flink/entrypoint.sh#L25

```
       -n, --no-clobber
              do not overwrite an existing file (overrides a -u or
              previous -i option). See also --update
```

Configuration should be valid

### What do you see instead?

Simplest solution for now will be probably remove `-n` from https://github.com/bitnami/containers/blob/fe7ac75eb9e200ed06efeeb6a11bdef77f922cd7/bitnami/flink/1/debian-12/rootfs/opt/bitnami/scripts/flink/entrypoint.sh#L25.

Bigger scope will be rewrite entrypoint scripts to produce valid YAML config which will be merging duplicates and ensuring it is valid in terms of YAML specification.

### Additional information

It is also related to https://github.com/bitnami/charts/issues/25457

daroga0002:
Closes https://github.com/bitnami/containers/issues/67010 https://github.com/bitnami/charts/issues/25457

### Description of the change

This will solve issue with configuration consistency and allow to avoid situation where YAML keys are duplicated after second run of entrypoint.

### Benefits

It will start working when entrypoint will be second time.

### Possible drawbacks

It will change bahaviour for users which are mounting /opt/bitnami/flink/conf as config.yaml will be always overwritten.

### Additional information

Configuration scripts are causing that main config should be threated as ephemeral. If they want to have persistent configuration then they should replace default base configs FLINK_DEFAULT_CONF_DIR:
https://github.com/bitnami/containers/blob/fe7ac75eb9e200ed06efeeb6a11bdef77f922cd7/bitnami/flink/1/debian-12/rootfs/opt/bitnami/scripts/flink-env.sh#L51

and then setup script will merge default base and environmental configs correctly.

### DIFF
--- a/bitnami/flink/1/debian-12/rootfs/opt/bitnami/scripts/flink/entrypoint.sh
+++ b/bitnami/flink/1/debian-12/rootfs/opt/bitnami/scripts/flink/entrypoint.sh
@@ -18,11 +18,8 @@ set -o pipefail
 
 print_welcome_page
 
-# We add the copy from default config in the entrypoint to not break users 
-# bypassing the setup.sh logic. If the file already exists do not overwrite (in
-# case someone mounts a configuration file in /opt/bitnami/flink/conf)
 debug "Copying files from $FLINK_DEFAULT_CONF_DIR to $FLINK_CONF_DIR"
-cp -nfr "$FLINK_DEFAULT_CONF_DIR"/. "$FLINK_CONF_DIR"
+cp -fr "$FLINK_DEFAULT_CONF_DIR"/. "$FLINK_CONF_DIR"
 
 if [[ "$1" = "/opt/bitnami/scripts/flink/run.sh" ]]; then
     info "** Starting Apache Flink ${FLINK_MODE} setup **"


daroga0002:
@jotamartos and @rafariossaa can you take a look into

jotamartos:
Hi @daroga0002,

As I just mentioned in the issue, we can't accept this change because it'll break the use-case when users mount the configuration file. The -n option prevents the configuration file to be overwritten. 

Regarding the issue, I posted all the information I got. Could you please take a look at it?

PR #67042, opened 2024-05-14
[bitnami/pgbouncer] Added server_round_robin settings support to pgbouncer

mehdithreem:
### Description of the change

Allow passing PGBOUNCER_SERVER_ROUND_ROBIN environment variable to the container.

### Benefits

From pg bouncer docs:

> By default, PgBouncer reuses server connections in LIFO (last-in, first-out) manner, so that few connections get the most load. This gives best performance if you have a single server serving a database. But if there is a round-robin system behind a database address (TCP, DNS, or host list), then it is better if PgBouncer also uses connections in that manner, thus achieving uniform load.


gongomgra (approved):

gongomgra:
@mehdithreem thanks for your contribution!

PR #67144, opened 2024-05-20
Minimal modifications to get bedrock and sage working

oaxfreeman:
### Description of the change

root.io/bedrock framework and roots.io/sage theme integration with bitnami/wordpress image

### Benefits

Providing users with Bedrock framework for modern wordpress development using Bitnami goodies

### Possible drawbacks

Depending on bedrock product instead of wordpress

### Applicable issues

- Wordpress instalation folder is still present ( need for cleanup but there is a strong coupling between bitnami scripts and old wordpress instalation )
- Wordpress initialization script had to be disabled ( as they are not applicable any longer )
- wp-cli is not configured to modify bedrock framework ( but i can be )

### Additional information

!!! THIS IS NOT MENT TO BE USED IN PRODUCTION !!!
!!! THIS PR IS JUST TO DEMONSTRATE MODIFICATIONS REQUIRED TO RUN BEDROCK FRAMEWORK IN BITNAMI WORDPRESS CONTAINER !!!


carrodher:
The `install_packages` script is placed at https://github.com/bitnami/containers/blob/main/bitnami/wordpress/6/debian-12/prebuildfs/usr/sbin/install_packages, you need to ensure that script is present to install packages

PR #67360, opened 2024-05-31
[bitnami/openldap] Allow to optionally set olcSuffix via LDAP_SUFFIX env var

xadhoom:
### Description of the change

Right now the olcSuffix config is fixed to the `LDAP_ROOT` value. While this is most of the times reasonable, those values are not strictly related, being olcSuffix the prefix of the queries handled by the internal mdb backend.
This change allows to optionally set the LDAP olcSuffix to a different value from `LDAP_ROOT`. The change does not break existing behaviour.

### Benefits

In multi tenant scenarios you end to have multiple trees, right now you either have to use a container for each tree (which is not easy if you create those trees dynamically), or provide a script on container init that does this change. Or, worse, prefix all those trees with a common one.
By allowing to set the olcSuffix on the container, simplifies a lot usage in those scenarios.

### Possible drawbacks

None

### DIFF
--- a/bitnami/openldap/2.5/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh
+++ b/bitnami/openldap/2.5/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh
@@ -53,6 +53,7 @@ export LDAP_DAEMON_GROUP="slapd"
 export LDAP_PORT_NUMBER="${LDAP_PORT_NUMBER:-1389}"
 export LDAP_LDAPS_PORT_NUMBER="${LDAP_LDAPS_PORT_NUMBER:-1636}"
 export LDAP_ROOT="${LDAP_ROOT:-dc=example,dc=org}"
+export LDAP_SUFFIX="$(if [ -z "${LDAP_SUFFIX+x}" ]; then echo "${LDAP_ROOT}"; else echo "${LDAP_SUFFIX}"; fi)"
 export LDAP_ADMIN_USERNAME="${LDAP_ADMIN_USERNAME:-admin}"
 export LDAP_ADMIN_DN="${LDAP_ADMIN_USERNAME/#/cn=},${LDAP_ROOT}"
 export LDAP_ADMIN_PASSWORD="${LDAP_ADMIN_PASSWORD:-adminpassword}"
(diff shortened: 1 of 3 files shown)


xadhoom:
Additional comment: the assignment expression of the new var is different from the others, to allow to set the empty value, that's setting olcSuffix to the empty string, which is legit and allows to store any tree in the underlying database.

bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh
```diff
@@ -53,6 +53,7 @@ export LDAP_DAEMON_GROUP="slapd"
 export LDAP_PORT_NUMBER="${LDAP_PORT_NUMBER:-1389}"
 export LDAP_LDAPS_PORT_NUMBER="${LDAP_LDAPS_PORT_NUMBER:-1636}"
 export LDAP_ROOT="${LDAP_ROOT:-dc=example,dc=org}"
+export LDAP_SUFFIX="$(if [ -z ${LDAP_SUFFIX+x} ]; then echo ${LDAP_ROOT}; else echo ${LDAP_SUFFIX}; fi)"
```
migruiz4, line 56: Could you please quote the variables?

```suggestion
export LDAP_SUFFIX="$(if [ -z "${LDAP_SUFFIX+x}" ]; then echo "${LDAP_ROOT}"; else echo "${LDAP_SUFFIX}"; fi)"
```

bitnami/openldap/2.5/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh
```diff
@@ -53,6 +53,7 @@ export LDAP_DAEMON_GROUP="slapd"
 export LDAP_PORT_NUMBER="${LDAP_PORT_NUMBER:-1389}"
 export LDAP_LDAPS_PORT_NUMBER="${LDAP_LDAPS_PORT_NUMBER:-1636}"
 export LDAP_ROOT="${LDAP_ROOT:-dc=example,dc=org}"
+export LDAP_SUFFIX="$(if [ -z ${LDAP_SUFFIX+x} ]; then echo ${LDAP_ROOT}; else echo ${LDAP_SUFFIX}; fi)"
```
migruiz4, line 56: Same suggestion here:

```suggestion
export LDAP_SUFFIX="$(if [ -z "${LDAP_SUFFIX+x}" ]; then echo "${LDAP_ROOT}"; else echo "${LDAP_SUFFIX}"; fi)"
```

migruiz4 (commented):
Thank you very much for your contribution @xadhoom!

Could you please take a look at my comment?

xadhoom:
> Could you please take a look at my comment?

Fixed and pushed, thanks.

migruiz4 (approved):
Looks good to me, thank you for your contribution!

PR #67399, opened 2024-06-03
[bitnami/wordpress] make from_email and from_name adjustable

mrclrchtr:
### Description of the change

This change makes it possible to adjust the sender email and the sender name in the smtp config. It is backwards compatible because it uses the previous environment variables if none are specified.

### Applicable issues

- discussion: https://github.com/bitnami/charts/issues/25174


gongomgra:
Hi @mrclrchtr,

Thanks for your contribution. We are looking into it

gongomgra (approved):
Thanks for your contribution!

mrclrchtr:
Thank you.

I have added DCO.

mrclrchtr:
I don't know why, but I didn't have a DCO yesterday. Now it's there... sorry about the circumstances.

gongomgra:
The DCO signature seems valid according to the "Commits" tab. We have had some issues with the action on our side though. We have temporarily disabled it, and I have also enabled auto-merge here. It should be merged automatically once the other validations succeeds

PR #67957, opened 2024-06-18
[bitnami/keycloak] Add support for proxy-headers

### LINKED ISSUE #65190: [bitnami/keycloak] Adapt scripts to proxy related changes in Keycloak 24
### Name and Version

bitnami/keycloak:24

Keycloak 24 introduces changes when it comes to proxy configuration.
See in particular:
- https://www.keycloak.org/server/reverseproxy#_proxy_modes
    > The support for setting proxy modes is deprecated and will be removed in a future Keycloak release. Consider configuring accepted reverse proxy headers instead as described in the chapter above. For migration instructions consult the [Upgrading Guide](https://www.keycloak.org/docs/latest/upgrading/index.html#deprecated-proxy-option).
- https://github.com/keycloak/keycloak/pull/25178
   It looks like `--proxy-headers` is the only option supported now, isn't it?

How to reproduce:
1. Run Docker container with `KEYCLOAK_PROXY=edge` for example.
2. See the following warning in the container logs:
    ```
    WARN  [org.keycloak.quarkus.runtime.cli.Picocli] (main) The following used options or option values are DEPRECATED and will be removed in a future release:
	- proxy: Use proxy-headers.
    ```

This Docker image should probably swap the `KEYCLOAK_PROXY` environment variable for a new one called `KEYCLOAK_PROXY_HEADERS`, mapping to the `proxy-headers` setting in Keycloak.

Note: this change will affect the Keycloak chart at https://github.com/bitnami/charts/tree/main/bitnami/keycloak.

Not really an "alternative", but in the meantime, I used `KEYCLOAK_EXTRA_ARGS: --proxy-headers xforwarded` to run Keycloak 24 (on Kubernetes).

Kajot-dev:
### Description of the change

Added support for `proxy-headers` via `KEYCLOAK_PROXY_HEADERS`. This also removes support for `KEYCLOAK_PROXY`
Implementing both this and KEYCLOAK_PROXY when the current default for KEYCLOAK_PROXY  is `passthorugh` (and not empty) would be a mess since we won't really know if user meant to set "passthrough" or just did not specify this at all (unless we would change the default, but this would be a breaking change anyway) and creating to many possible combinations. 

### Benefits

Not using deprecated options

### Possible drawbacks

Lack of `KEYCLOAK_PROXY` must be accounted for in helm charts etc.

### Applicable issues

- fixes #65190


juan131:
Hi @Kajot-dev 

This isn't simply adding support for `proxy-headers` but dropping support for customizing the proxy mode (which is going to be deprecated). Based on the Upgrading Guide below:

- https://www.keycloak.org/docs/latest/upgrading/index.html#deprecated-proxy-option

I'm assuming you're not setting any proxy-header by default given it's the equivalent to the "passthrough" proxy mode, am I right?

Kajot-dev:
@juan131 Yes, that's correct!

If you wish, I may try to implement backwards compatibility by introducing KEYCLOAK_LEGACY_PROXY which defaulta to the value of KEYCLOAK_PROXY (and is only a helper variable) before it's assigned the default "passthrough" value. That way the intentions of the user would be clear. Then if KEYCLOAK_LEGACY_PROXY is empty, we can use proxy, headers, else use proxy setting like before. Let me know what you think?

Kajot-dev:
I will make counterpart PR for helm chart and link it here

singhbaljit:
According to the [docs](https://www.keycloak.org/docs/latest/upgrading/index.html#deprecated-proxy-option), the `edge` mode (`kc.sh --proxy edge`) is equivalent to `kc.sh --proxy-headers forwarded|xforwarded --http-enabled true`. Do we need to add another variable to also support the `http-enabled` option?

Kajot-dev:
> According to the [docs](https://www.keycloak.org/docs/latest/upgrading/index.html#deprecated-proxy-option), the `edge` mode (`kc.sh --proxy edge`) is equivalent to `kc.sh --proxy-headers forwarded|xforwarded --http-enabled true`. Don't we need to add another variable to also support the `http-enabled` option (`KC_HTTP_ENABLED`)?.. HTTP is disabled by default... or just rely on the Keycloak provided variable. Perhaps we keep the proxy mode still an option in the chart, but conditionally update the `KEYCLOAK_PROXY_HEADERS` and `KC_HTTP_ENABLED` based on the mode value, per that table.

I don't think so, because of:
https://github.com/bitnami/containers/blob/main/bitnami/keycloak/24/debian-12/rootfs/opt/bitnami/scripts/libkeycloak.sh#L222

In bitnami containers `http-enabled` is always set to true unconditionally

Kajot-dev:
@juan131 Please see: https://github.com/bitnami/charts/pull/27890

juan131 (approved):
LGTM

juan131:
Hi @Kajot-dev 

We released a new container image version (see `24.0.5-debian-12-r3`) including your changes, see https://github.com/bitnami/containers/commit/536990b719e19f1b9e2967985c242db2ae01180a

Thanks for your contribution

PR #68066, opened 2024-06-25
[bitnami/moodle] Use default moodle db env vars in post-init.d script #68054

### LINKED ISSUE #68054: [bitnami/moodle] sql-script in post-init.d should use exising database environment variables
### Name and Version

bitnami/moodle:4.4.1

None

1. Use the docker-compose.yaml within the git repo
2. Create a simple sql script and place it in a directory that docker can reach
3. Change so that the directory used before is mounted in /docker-entrypoint.d/
4. Start the containers using the docker-compose up

The SQL script is applied to the moodle database

### What do you see instead?

An error stating that mysql command could not succeeed.

### Additional information

If the sql-mysql.sh script in /post-init.d/ used the MOODLE_DATABASE variables defined in the docker-compose.yaml file, then it was able to connect to the database using the same credentials as Moodle would connect to the database.
I would assume that the changes one would make, is limitied to the database that was initialized for moodle.
I use it for applying configuration during init instead of having to click-ops them for the installation of moodle.

ThomasRasmussen:
### Description of the change
When mounting post-init sql scripts, use MOODLE_DATABASE environment variables by default unless the MARIADB environment variables have been specified. 

### Benefits
Allow easier configuration of the moodle installation as we use the same database user that moodle will use to connect.

### Possible drawbacks
If for some reason you need to run something as root user, then you need to specify these credentials.

### Applicable issues

- fixes #68054

### DIFF
--- a/bitnami/airflow-worker/2/debian-12/Dockerfile
+++ b/bitnami/airflow-worker/2/debian-12/Dockerfile
@@ -7,15 +7,15 @@ ARG TARGETARCH
 
 LABEL com.vmware.cp.artifact.flavor="sha256:c50c90cfd9d12b445b011e6ad529f1ad3daea45c26d20b00732fae3cd71f6a83" \
       org.opencontainers.image.base.name="docker.io/bitnami/minideb:bookworm" \
-      org.opencontainers.image.created="2024-08-09T15:37:20Z" \
+      org.opencontainers.image.created="2024-08-16T05:17:50Z" \
       org.opencontainers.image.description="Application packaged by Broadcom, Inc." \
       org.opencontainers.image.documentation="https://github.com/bitnami/containers/tree/main/bitnami/airflow-worker/README.md" \
       org.opencontainers.image.licenses="Apache-2.0" \
-      org.opencontainers.image.ref.name="2.9.3-debian-12-r5" \
+      org.opencontainers.image.ref.name="2.10.0-debian-12-r0" \
       org.opencontainers.image.source="https://github.com/bitnami/containers/tree/main/bitnami/airflow-worker" \
       org.opencontainers.image.title="airflow-worker" \
       org.opencontainers.image.vendor="Broadcom, Inc." \
-      org.opencontainers.image.version="2.9.3"
+      org.opencontainers.image.version="2.10.0"
 
 ENV HOME="/" \
     OS_ARCH="${TARGETARCH:-amd64}" \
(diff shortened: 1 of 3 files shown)


bitnami/moodle/4.4/debian-12/rootfs/post-init.d/sql-mysql.sh
```diff
@@ -21,7 +21,7 @@ fi
 mysql_execute() {
     local -r sql_file="${1:?missing file}"
     local failure=0
-    mysql_cmd=("mysql" "-h" "$MARIADB_HOST" "-P" "$MARIADB_PORT_NUMBER" "-u" "$MARIADB_ROOT_USER")
+    mysql_cmd=("mysql" "-h" "$MARIADB_HOST" "-P" "$MARIADB_PORT_NUMBER" "-u" "$MARIADB_ROOT_USER" "-D" "${MARIADB_DATABASE}")
```
rafariossaa, line 24: This change need to be added to the other moodle's versions too.

bitnami/moodle/4.4/debian-12/rootfs/post-init.d/sql-mysql.sh
```diff
@@ -33,6 +33,15 @@ mysql_execute() {
     return "$failure"
 }
 
+MARIADB_HOST=${MARIADB_HOST:-$MOODLE_DATABASE_HOST}
+MARIADB_PORT_NUMBER=${MARIADB_PORT_NUMBER:-$MOODLE_DATABASE_PORT_NUMBER}
+MARIADB_ROOT_USER=${MARIADB_ROOT_USER:-$MOODLE_DATABASE_USER}
+MARIADB_DATABASE=${MARIADB_DATABASE:-$MOODLE_DATABASE_NAME}
+if [[ "${ALLOW_EMPTY_PASSWORD:-no}" != "yes" ]]; then
+  MARIADB_ROOT_PASSWORD=${MARIADB_ROOT_PASSWORD:-$MOODLE_DATABASE_PASSWORD}
+fi
```
rafariossaa, line 42: I would move this block to line 20, above the function `mysql_execute`.

rafariossaa (changes requested):

bitnami/moodle/4.4/debian-12/rootfs/post-init.d/sql-mysql.sh
ThomasRasmussen: I have now alligned the versions to use the same script

bitnami/moodle/4.4/debian-12/rootfs/post-init.d/sql-mysql.sh
ThomasRasmussen: Yeah I see your point. This has now been changed

rafariossaa:
Hi,
  Sorry for the long delay.
  Thanks for your PR. I was looking on how could I integrate this into our CI, but unfortunately this kind of modifications are not supported right now, so I can not merged this change.
  As you tried to simplify, the only way right now is to indicate the user credentials in `MOODLE_xxx` and `MARIADB_xxx` environment variables.
  I opened an internal task to look for some improvement in this area, but I can not provide an ETA for it.

PR #68236, opened 2024-06-29
[bitnami/postgresql] Updated libpostgresql.sh to specify a host to Postgres Commands

### LINKED ISSUE #68027: [bitnami/postgresql] Custom unix_socket_directories causes startup configuration to fail
### Name and Version

bitnami/postgresql:16.3.0

amd64

1. In any environment.
2. Use the Bitnami Postgresql Helm chart
3. Update Postgresql configuration to change where unix sockets are created:

```yaml
primary:
  configuration: |-
    unix_socket_directories = "/tmp/foo"
    listen_addresses = "0.0.0.0"
```
4.  Deploy with the custom configuration.

Postgresql starts up and as expected and uses the custom socket directory of "/tmp/foo" to store Unix sockets.

### What do you see instead?

Postgresql starts up and creates a socket in the custom directory, but the startup Bitnami scripts never detect that Postgresql is ready. This causes the container to eventually restart without finishing the startup configuration (such as setting up accounts and passwords). After restarting, it never re-runs the startup configuration, so the accounts and passwords for Postgresql are never properly initialized.

### Additional information

This appears to happen in `postgresql_start_bg` at https://github.com/bitnami/containers/blob/main/bitnami/postgresql/16/debian-12/rootfs/opt/bitnami/scripts/libpostgresql.sh#L778. At line 797, It runs `pg_isready`, but does not pass in any arguments to specify the hostname which now differs from the default hostname/unix socket. This is an asymmetry with how the probes in the [Bitnami Postgresql Helm Chart works](https://github.com/bitnami/charts/blob/main/bitnami/postgresql/templates/primary/statefulset.yaml#L434), which pass in `-h 127.0.0.1` to specify the hostname. This creates a situation where the pod reports it is healthy and ready, but ultimately restarts because the `postgresql_start_bg` never detects postgresql is running and forces the script to exit.

When exiting, the PID file for postgresql is never cleaned up, so the next time the entrypoint runs and `postgresql_start_bg` runs again, it detects that Postgres is already running and never attempts to re-run the startup configuration. This puts Postgres in a state where it is running happily, but has never actually finalized its startup configuration so accounts and databases have not been initialized.

As for why a user may want to change the `unix_socket_directories`, Iron Bank provides a [hardened Bitnami Postgresql image](https://repo1.dso.mil/dsop/bitnami/postgres/postgresql16/-/blob/development/Dockerfile?ref_type=heads) built on top of the RHEL UBI. The RHEL Postgresql port defaults to creating sockets under "/run", to avoid the possibility of a Postgresql instance being mimicked when running on bare metal with unix sockets stored under "/tmp". This poses a problem for secured containers as read-only root filesystems are a typical security control for secure containers.

This can be worked around by adjusting the container's security context to make the root filesystem read-write, at the expense of security precautions. Instead, it would be useful to be able to update the configuration of Postgresql to store unix sockets under /tmp when running in containers. 

It seems that a straightforward solution here might be for the `postgresql_start_bg` function to pass the host option with the loopback address to the `pg_isready` function similar to how the Helm chart's probes work. However, there may be some edge cases here that I haven't considered.

james-mchugh:
### Description of the change

Updates the `postgresql_start_bg` and `postgresql_execute_print_output` functions in libpostgresql.sh to specify the loopback address as the host for `psql` and `pg_isready` commands.

### Benefits

If a custom unix socket directory is specified in the Primary configuration, startup will fail as the `psql` and `pg_isready` commands do not look at the correct location for the Unix socket directory. This updates them to use the loopback host which should work regardless of where the unix socket file is stored, and it is also consistent with the behavior of the probes used by the Helm chart.

Being able to specify a custom Unix socket directory is useful, as the [Bitnami Postgresql image provided by Iron Bank](https://repo1.dso.mil/dsop/bitnami/postgres/postgresql16/-/blob/development/Dockerfile?ref_type=heads) is built on top of the RHEL Universal Base Image and uses the RHEL Postgresql port. The RHEL Postgresql port defaults to storing Postgresql's unix socket under /run, which poses an issue as it is a common security measure to deploy containers with read-only root filesystems.

### Possible drawbacks

None that I'm aware of. I've tested this in standalone and replication mode, and both appeared to work as expected.

### Applicable issues

- fixes #68027 

### Additional information

There appears to be a separate bug here mentioned in #68027 where if the startup setup fails the first time, it will never be completed. When this happens, the Postgresql container will act as if it is healthy even though the authentication and databases have not been properly setup. This PR does not fix that issue.

### DIFF
--- a/bitnami/postgresql/16/debian-12/rootfs/opt/bitnami/scripts/libpostgresql.sh
+++ b/bitnami/postgresql/16/debian-12/rootfs/opt/bitnami/scripts/libpostgresql.sh
@@ -792,7 +792,7 @@ postgresql_start_bg() {
     else
         "${pg_ctl_cmd[@]}" "start" "${pg_ctl_flags[@]}" >/dev/null 2>&1
     fi
-    local pg_isready_args=("-U" "postgres" "-p" "$POSTGRESQL_PORT_NUMBER")
+    local pg_isready_args=("-U" "postgres" "-p" "$POSTGRESQL_PORT_NUMBER" "-h" "127.0.0.1")
     local counter=$POSTGRESQL_INIT_MAX_TIMEOUT
     while ! "$POSTGRESQL_BIN_DIR"/pg_isready "${pg_isready_args[@]}" >/dev/null 2>&1; do
         sleep 1
@@ -1078,7 +1078,7 @@ postgresql_execute_print_output() {
     local opts
     read -r -a opts <<<"${@:4}"
 
-    local args=("-U" "$user" "-p" "${POSTGRESQL_PORT_NUMBER:-5432}")
+    local args=("-U" "$user" "-p" "${POSTGRESQL_PORT_NUMBER:-5432}" "-h" "127.0.0.1")
     [[ -n "$db" ]] && args+=("-d" "$db")
     [[ "${#opts[@]}" -gt 0 ]] && args+=("${opts[@]}")
 


jotamartos:
Thanks for taking the time to create this PR. Please let us test the changes and ensure this doesn't break the configuration of any solution in the catalog. Thanks

jotamartos (approved):
Internal tests succeeded. Thanks for your contribution

PR #68513, opened 2024-07-01
[bitnami/pgbouncer] Fixed order of `PGBOUNCER_INIT_SLEEP_TIME` and `PGBOUNCER_INIT_MAX_RETRIES` arguments inside `pgbouncer_initialize`

kchaber:
### Description of the change

The `retry_while` function expects the `retries` parameter to be passed before the `sleep_time`. The PR fixes the order of these arguments, which are currently swapped.


andresbono (approved):
LGTM. Good catch, thank you for your contribution!

PR #69426, opened 2024-07-22
[bitnami/postgresql-repmgr] Fixing POSTGRESQL_INIT_MAX_TIMEOUT didn't take effect

ccsheller:
### Description of the change

This change adds the -W option to the pg_ctl command used to start PostgreSQL in slow machines, such as virtual machines. The -W option explicitly tells pg_ctl not to wait for the startup to complete.

### Additional information

If the -w option is used, pg_ctl will still be waiting for the startup to complete, which can take more than a minute on slow machines. If pg_ctl times out, the entire container will be terminated, and POSTGRESQL_INIT_MAX_TIMEOUT will not take effect.


jotamartos (approved):
Thanks for your contribution. I'll update our internal code and release a new version of the solution with the change

PR #69439, opened 2024-07-22
[bitnami/rabbitmq] RabbitMQ UI support IPv6

### LINKED ISSUE #69256: RabbitUI does not work in IPv6
### Name and Version

bitnami/rabbitmq:3.13.4

None

Run rabbitmq in k8s with IPv6 only network and use ingress.

UI is available with DNS mentioned in ingress for rabbitmq

### What do you see instead?

503 Service Temporarily Unavailable

### Additional information

I can see in the repo that UI works only on IPv4
https://github.com/bitnami/containers/blob/main/bitnami/rabbitmq/3.13/debian-12/rootfs/opt/bitnami/scripts/librabbitmq.sh#L129C1-L129C64

Is there any reason for this? And is it possible to make an improvement so that UI will be available in IPv6?

SpeedyGonzaless:
### Description of the change

Updated IP validation function in RabbitMQ configuration script to support IPv6

### Benefits

RabbitMq UI will work in IPv6 only networks

### Applicable issues

- fixes #69256


dgomezleon (approved):

PR #69505, opened 2024-07-23
[bitnami/volsync] ci: :construction_worker: Add VIB integration

javsalgar:
### Description of the change

This PR adds basic VIB tests for Volsync. It passes all the Bitnami standard tests and it checks that the binaries and configuration folders are present. As it is a Kubernetes-native application we cannot perform more tests.


migruiz4 (approved):
LGTM!

PR #70398, opened 2024-07-29
[bitnami/openldap] Add support for setting the password crypt format

mistial-dev:
### Description of the change

This patch adds the configuration option "LDAP_PASSWORD_CRYPT_SALT_FORMAT", allowing the crypt salt format to be set in the container environment variables.

Additionally, this patch changes the default for password encoding from SHA (which is now retired by NIST) to salted SHA-256 using crypt() (which defaults to 5,000 rounds).  This change is also applied to administrative passwords (admin, config admin, accesslog admin).

### Benefits

NIST recommends moving away from SHA-1 due to insecurity.

> The SHA-1 algorithm, one of the first widely used methods of protecting electronic information, has reached the end of its useful life, according to security experts at the National Institute of Standards and Technology (NIST). The agency is now recommending that IT professionals replace SHA-1, in the limited situations where it is still used, with newer algorithms that are more secure.  

While it is possible to use a third party module for SHA-256 and SHA-512 from an atlassian employee, it is simpler and more compatible to use {CRYPT} support.  This allows the number of rounds to be user-configured for increased attack resistance, allows the algorithm to be specified, and provides the hash in a way that other applications using compatible crypt() implementations can interoperate with.

This improves the default security posture while supporting existing configurations.

### Possible drawbacks

Applications designed to work directly with hash values (for example, the SHA or MD5 hash) will not work unless they also have support for crypt passwords.  The previous behaviour can be restored by setting LDAP_PASSWORD_HASH to "SSHA".

Note that applications that bind as a user with a simple password should be unaffected by this change.  Only applications using the password hash directly may be affected.

### Additional information

To take advantage of the new security defaults, LDAP_CONFIGURE_PPOLICY and LDAP_PPOLICY_HASH_CLEARTEXT should be set to yes.  This will result in user passwords automatically being hashed, even if they are set in plaintext.  This applies to users created through the LDAP_USERS and LDAP_PASSWORDS environment variables.

20351936-3250-4940-912B-B61BF8E994D3


mistial-dev:
I kept the automatic plaintext conversion disabled in the interest of compatibility.  I could change that as well to make this more secure out of the box, or add a LDAP_HARDENED environment variable that sets some sensible security defaults, including adding ACLs.

Unless you need something like MSCHAPv2 support (VPN for windows, mainly), it's a very good idea to have ACLs set which prohibit reading the contents of the userPassword attribute.  If I were doing a hardened config, I'd also be tempted to limit anonymous binding and disable enumeration by users (permitting search by email or username), and adding an ou for service accounts that had full access.

andresbono:
Hi, thank you very much for this PR, it looks like a very nice addition! I will run some manual tests on my side and I'll get back to you.

Thanks again!

bitnami/openldap/2.5/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh
```diff
@@ -633,9 +634,7 @@ ldap_initialize() {
             ldap_add_custom_schemas
         fi
         # additional configuration
-        if [[ ! "$LDAP_PASSWORD_HASH" == "{SSHA}" ]]; then
-            ldap_configure_password_hash
-        fi
+        ldap_configure_password_hash
```
andresbono, line 637: I guess it does not hurt to always add those configs no matter the algorithm, right?

bitnami/openldap/2.5/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh
```diff
@@ -112,9 +113,9 @@ done
 unset ldap_env_vars
 
 # Setting encrypted admin passwords
-export LDAP_ENCRYPTED_ADMIN_PASSWORD="$(echo -n $LDAP_ADMIN_PASSWORD | slappasswd -n -T /dev/stdin)"
-export LDAP_ENCRYPTED_CONFIG_ADMIN_PASSWORD="$(echo -n $LDAP_CONFIG_ADMIN_PASSWORD | slappasswd -n -T /dev/stdin)"
-export LDAP_ENCRYPTED_ACCESSLOG_ADMIN_PASSWORD="$(echo -n $LDAP_ACCESSLOG_ADMIN_PASSWORD | slappasswd -n -T /dev/stdin)"
+export LDAP_ENCRYPTED_ADMIN_PASSWORD="$(echo -n $LDAP_ADMIN_PASSWORD | slappasswd -c '$5$%.16s' -n -T /dev/stdin)"
+export LDAP_ENCRYPTED_CONFIG_ADMIN_PASSWORD="$(echo -n $LDAP_CONFIG_ADMIN_PASSWORD | slappasswd -c '$5$%.16s' -n -T /dev/stdin)"
+export LDAP_ENCRYPTED_ACCESSLOG_ADMIN_PASSWORD="$(echo -n $LDAP_ACCESSLOG_ADMIN_PASSWORD | slappasswd -c '$5$%.16s' -n -T /dev/stdin)"
```
andresbono, line 118: If `LDAP_PASSWORD_CRYPT_SALT_FORMAT` is configurable, should it be used here instead of hardcoding the value?

andresbono (commented):
Could you please check my comments? Thanks!

mistial-dev:
> bitnami/openldap/2.5/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh

Looking at them now.

bitnami/openldap/2.5/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh
mistial-dev: This approach forced the use of the higher security default for the admin password.  I could make it configurable, but it wouldn't force the higher security option for the admin password.

bitnami/openldap/2.5/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh
mistial-dev: Making it explicit should not hurt.

bitnami/openldap/2.5/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh
andresbono: I see. I think it should be configurable. If users may need different formats for admin and the rest, maybe we can define a new env-var: `LDAP_ADMIN_PASSWORD_CRYPT_SALT_FORMAT`. What do you think?

andresbono (commented):
Thank you, could you please check my comment?

bitnami/openldap/2.5/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh
mistial-dev: That makes sense.

mistial-dev: > I see. I think it should be configurable. If users may need different formats for admin and the rest, maybe we can define a new env-var: `LDAP_ADMIN_PASSWORD_CRYPT_SALT_FORMAT`. What do you think?

Added.  Additionally, I changed the defaults to enable ppolicy and hash plaintext passwords by default.

It can be turned off easily enough, either through environment variables or ldapmodify after the fact.  Sane defaults are better, and with Microsoft Active Directory using hashed passwords, there's precedent for not doing plaintext passwords.  It also works fine for shadow support, and the use of CRYPT instead of something weird increases compatibility when used elsewhere.

bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh
```diff
@@ -76,10 +76,12 @@ export LDAP_REQUIRE_TLS="${LDAP_REQUIRE_TLS:-no}"
...
 export LDAP_LOGLEVEL="${LDAP_LOGLEVEL:-256}"
-export LDAP_PASSWORD_HASH="${LDAP_PASSWORD_HASH:-{SSHA\}}"
-export LDAP_CONFIGURE_PPOLICY="${LDAP_CONFIGURE_PPOLICY:-no}"
+export LDAP_PASSWORD_HASH="${LDAP_PASSWORD_HASH:-{CRYPT\}}"
+export LDAP_PASSWORD_CRYPT_SALT_FORMAT="${LDAP_PASSWORD_CRYPT_SALT_FORMAT:-\$5\$%.16s}"
+export LDAP_ADMIN_PASSWORD_CRYPT_SALT_FORMAT="${LDAP_ADMIN_PASSWORD_CRYPT_SALT_FORMAT:-\$5\$%.16s}"
+export LDAP_CONFIGURE_PPOLICY="${LDAP_CONFIGURE_PPOLICY:-yes}"
 export LDAP_PPOLICY_USE_LOCKOUT="${LDAP_PPOLICY_USE_LOCKOUT:-no}"
-export LDAP_PPOLICY_HASH_CLEARTEXT="${LDAP_PPOLICY_HASH_CLEARTEXT:-no}"
+export LDAP_PPOLICY_HASH_CLEARTEXT="${LDAP_PPOLICY_HASH_CLEARTEXT:-yes}"
```
mistial-dev, line 84: Applying this through the policy means that it's applied when necessary.  It doesn't force a particular hashed format if non-plaintext passwords are applied.

bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh
```diff
@@ -76,10 +76,12 @@ export LDAP_REQUIRE_TLS="${LDAP_REQUIRE_TLS:-no}"
 export LDAP_ULIMIT_NOFILES="${LDAP_ULIMIT_NOFILES:-1024}"
 export LDAP_ALLOW_ANON_BINDING="${LDAP_ALLOW_ANON_BINDING:-yes}"
 export LDAP_LOGLEVEL="${LDAP_LOGLEVEL:-256}"
-export LDAP_PASSWORD_HASH="${LDAP_PASSWORD_HASH:-{SSHA\}}"
-export LDAP_CONFIGURE_PPOLICY="${LDAP_CONFIGURE_PPOLICY:-no}"
+export LDAP_PASSWORD_HASH="${LDAP_PASSWORD_HASH:-{CRYPT\}}"
+export LDAP_PASSWORD_CRYPT_SALT_FORMAT="${LDAP_PASSWORD_CRYPT_SALT_FORMAT:-\$5\$%.16s}"
+export LDAP_ADMIN_PASSWORD_CRYPT_SALT_FORMAT="${LDAP_ADMIN_PASSWORD_CRYPT_SALT_FORMAT:-\$5\$%.16s}"
+export LDAP_CONFIGURE_PPOLICY="${LDAP_CONFIGURE_PPOLICY:-yes}"
```
mistial-dev, line 82: Setting this as the default ensures that password hashing is used even when the environment variables are used to pass credentials and no policy is set.  Leaving plaintext passwords around is risky.

andresbono:
@mistial-dev, just one minor thing, The DCO check failed. Please read https://github.com/bitnami/containers/pull/70398/checks?check_run_id=30415074724 to see how you can fix that. Thanks!

mistial-dev:
> @mistial-dev, just one minor thing, The DCO check failed. Please read https://github.com/bitnami/containers/pull/70398/checks?check_run_id=30415074724 to see how you can fix that. Thanks!

I tried what it suggested earlier, and it ended up making a hash of things.  I'll try again.

mistial-dev:
> @mistial-dev, just one minor thing, The DCO check failed. Please read https://github.com/bitnami/containers/pull/70398/checks?check_run_id=30415074724 to see how you can fix that. Thanks!

@andresbono I have followed the instructions in the document.  It is now attempting to modify 2,409 files.

gongomgra:
Hi @mistial-dev,

Thanks for fixing the DCO issue. Can you rebase the main branch on your fork and fix the conflicts?

PR #70431, opened 2024-07-30
[.vib] Add chainloop containers VIB files

migruiz4:
### Description of the change

Add basic vib-verify pipelines for scratch containers.


fmulero (approved):
LGTM.

Just to double check. Can these containers start and keep running without any configuration or dependency? If not, please add the runtime_parameters field.

migruiz4:
@fmulero They can't, but they are `scratch` containers so adding runtime_parameters is pointless. We add basic tests as we do in other scratch containers such as `charts-syncer`, `oras`, etc.

PR #70466, opened 2024-07-31
[bitnami/keycloak] add variables to support hostname v2

### LINKED ISSUE #70461: [bitnami/keycloak] Use Hostname v2 in Keycloak v25 images
### Name and Version

bitnami/keycloak:25

Hostname v1 options were [deprecated (disabled?) in Keycloak v25](https://www.keycloak.org/docs/latest/upgrading/index.html#new-hostname-options). Per docs:

> Hostname v2 options are supported by default, as the old hostname options are deprecated and will be removed in the following releases. New options are activated by default, so Keycloak will not recognize the old options.

We need to update container entrypoint scripts. Currently, there's only one variable `KEYCLOAK_HOSTNAME`. We should probably add `KEYCLOAK_HOSTNAME_ADMIN` as well. Currently, the chart is specifying `KC_HOSTNAME_URL` and `KC_HOSTNAME_ADMIN_URL` directly; the chart should use these variables instead (see https://github.com/bitnami/charts/issues/28586). `hostname-strict`/`hostname-backchannel-dynamic` options need to updated accordingly as well.

Migrate to use V2 hostname so that we can simplify definition of these variables in the Keycloak chart.

No alternative as the v1 hostname variables are disabled by default, and will be removed.

singhbaljit:
### Description of the change

Use [Hostname v2](https://www.keycloak.org/docs/latest/upgrading/index.html#new-hostname-options) options as v1 options are not used by default.

### Benefits

Use updated configurations in alignment with upstream Keycloak project.

### Possible drawbacks

`KEYCLOAK_HOSTNAME` variable is the same between v1 and v2 option; however, its usage is slightly different in v2. User  may not realize the change and expect the same behavior as v1.

### Applicable issues

- fixes #70461

### Additional information

PR to use the options in the chart: https://github.com/bitnami/charts/pull/28611.


bitnami/keycloak/25/debian-12/rootfs/opt/bitnami/scripts/libkeycloak.sh
```diff
@@ -198,7 +198,7 @@ keycloak_configure_health_endpoints() {
 keycloak_configure_hostname() {
     info "Configuring hostname settings"
     ! is_empty_value "$KEYCLOAK_HOSTNAME" && keycloak_conf_set "hostname" "${KEYCLOAK_HOSTNAME}"
-    keycloak_conf_set "hostname-strict" "false"
```
singhbaljit, line 201: Per [docs](https://www.keycloak.org/server/hostname):
>Should always be set to true in production, unless your reverse proxy overwrites the Host header. If enabled, the hostname option needs to be specified.

This option should be `true` by default. The user can override this using `KC_HOSTNAME_STRICT` if necessary.

singhbaljit: I realized that chart supports empty hostname; I added another variable to define this value via chart.

singhbaljit:
I added support for two additional variables: `KEYCLOAK_HOSTNAME_ADMIN` and `KEYCLOAK_HOSTNAME_STRICT` with the same defaults as before. The chart will have to be updated to override the defaults based on ingress configs.

dgomezleon (commented):
Hi @singhbaljit ,

Please note that the logic is shared for both branches 24 and 25, so you should apply it for version 24 too. I don't see any risk in doing that since the default values did not change.
Could you please check it?

singhbaljit:
@dgomezleon I updated v24 scripts as well.

dgomezleon (dismissed):

dgomezleon:
Thanks @singhbaljit 

LGTM. Could you please check the conflicts?

singhbaljit:
The conflict is due to the fact that v24 is now removed. 😄 . So, I'll just undo the v24 changes.

dgomezleon:
> The conflict is due to the fact that v24 is now removed 😄, see #70607 . So, I'll just undo the v24 changes.

Just today :). Sorry for the noise.

dgomezleon (approved):

PR #70543, opened 2024-08-02
feat: install some popular tools for nodejs developers

jarqvi:
### Description of the change
Install some popular tools for nodejs developers


carrodher:
Hi @jarqvi, thank you for your contribution!

To maintain a balance between usability and security in Bitnami images, we aim to include only the packages and tools that are strictly necessary. This approach helps to minimize the attack surface, reduce potential vulnerabilities, and keep the image size manageable.

If you need to add additional packages or tools, we recommend extending the _bitnami/node_ container for your specific use case. This way, you can customize the container while keeping the core image streamlined for others.

jarqvi:
Ohh, ok.
Thanks @carrodher.

PR #70597, opened 2024-08-06
[bitnami/etcd] fix: healthcheck will failed when startup etcd with on…

### LINKED ISSUE #70554: [bitnami/etcd] bug: healthcheck will failed when startup etcd with one-way tls authentication
### Name and Version

bitnami/etcd: 3.5.15

None

1. container.env
```
ALLOW_NONE_AUTHENTICATION = "yes"
ETCD_LISTEN_CLIENT_URLS = "https://0.0.0.0:2379"
ETCD_ADVERTISE_CLIENT_URLS = "https://etcd:2379"
ETCD_CERT_FILE = "/certs/tls.crt"
ETCD_KEY_FILE = "/certs/tls.key"
```
2. compose.yaml
```yaml
services:
  etcd:
    image: docker.io/bitnami/etcd:3.5.15`
    restart: unless-stopped
    env_file: ./container.env
    #ports: [2379,2380]
    deploy:
      resources:
        limits:
          cpus: '1'
          memory: 2G
    volumes:
      - etcd_data:/bitnami/etcd/data
      - ./certs:/certs
```
3. startup etcd and go into the container
4. run /opt/bitnami/scripts/etcd/healthcheck.sh will fail
image

healthcheck.sh should still succeed when startup etcd with one-way tls authentication

### What do you see instead?

/opt/bitnami/scripts/etcd/healthcheck.sh failed

### Additional information

1. when startup etcd with two-way tls authentication, healthcheck.sh no problem
container.env
```
ALLOW_NONE_AUTHENTICATION = "yes"
ETCD_LISTEN_CLIENT_URLS = "https://0.0.0.0:2379"
ETCD_ADVERTISE_CLIENT_URLS = "https://etcd:2379"
ETCD_CERT_FILE = "/certs/tls.crt"
ETCD_KEY_FILE = "/certs/tls.key"
ETCD_TRUSTED_CA_FILE = "/certs/ca.crt"
```
2. but when startup ectd with one-way tls authentication, healthcheck.sh will fail
as not configured ETCD_TRUSTED_CA_FILE, so when running healthcheck.sh, this script runs as a client to connect the etcd server, if not skip the tls verification, and it will try to find a CA file to verify the server certs, but now there is no such CA file, so will fail
3. just need to allow healthcheck.sh always skip tls verify, modify libetcd.sh like 
image
as when running script healthcheck.sh, it just runs as a client, so we should allow it to skip tls verification

4. related bitnami/etcd helm chart issue: https://github.com/bitnami/charts/issues/27109

chenraoCR:
### Description of the change

modified the script healthcheck.sh to skip tls verification to allow startup etcd with one-way tls authentication

### Applicable issues

- fixes #70554

### DIFF
--- a/bitnami/etcd/3.4/debian-12/rootfs/opt/bitnami/scripts/etcd/healthcheck.sh
+++ b/bitnami/etcd/3.4/debian-12/rootfs/opt/bitnami/scripts/etcd/healthcheck.sh
@@ -21,9 +21,12 @@ host="$(parse_uri "${advertised_array[0]}" "host")"
 port="$(parse_uri "${advertised_array[0]}" "port")"
 read -r -a extra_flags <<< "$(etcdctl_auth_flags)"
 extra_flags+=("--endpoints=${host}:${port}")
-if [[ $ETCD_AUTO_TLS = true ]]; then
+
+ # if ETCD_AUTO_TLS true or CA file not exists, just skip server cert verification
+if [[ $ETCD_AUTO_TLS = true ]] || [[ ! -f "$ETCD_TRUSTED_CA_FILE" ]]; then
      extra_flags+=("--insecure-skip-tls-verify")
 fi
+
 if etcdctl endpoint health "${extra_flags[@]}"; then
     exit 0
 else
--- a/bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/healthcheck.sh
+++ b/bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/healthcheck.sh
@@ -21,9 +21,12 @@ host="$(parse_uri "${advertised_array[0]}" "host")"
 port="$(parse_uri "${advertised_array[0]}" "port")"
 read -r -a extra_flags <<< "$(etcdctl_auth_flags)"
 extra_flags+=("--endpoints=${host}:${port}")
-if [[ $ETCD_AUTO_TLS = true ]]; then
+
+ # if ETCD_AUTO_TLS true or CA file not exists, just skip server cert verification
+if [[ $ETCD_AUTO_TLS = true ]] || [[ ! -f "$ETCD_TRUSTED_CA_FILE" ]]; then
      extra_flags+=("--insecure-skip-tls-verify")
 fi
+
 if etcdctl endpoint health "${extra_flags[@]}"; then
     exit 0
 else


chenraoCR:
@andresbono @javsalgar so what's the progress of this PR?

andresbono (commented):
Thank you for submitting this PR!

The changes you are proposing will affect several functions. Isn't adding `--insecure-transport=false` and `--insecure-skip-tls-verify=true` unconditionally too aggressive?

Maybe it could only be necessary when there is no `ETCD_TRUSTED_CA_FILE`? Or maybe it should be scoped to the `healthcheck.sh` script only.

Let me know your thoughts!

chenraoCR:
> Thank you for submitting this PR!
> 
> The changes you are proposing will affect several functions. Isn't adding `--insecure-transport=false` and `--insecure-skip-tls-verify=true` unconditionally too aggressive?
> 
> Maybe it could only be necessary when there is no `ETCD_TRUSTED_CA_FILE`? Or maybe it should be scoped to the `healthcheck.sh` script only.
> 
> Let me know your thoughts!

ya, could  only be necessary when there is no ETCD_TRUSTED_CA_FILE

bitnami/etcd/3.4/debian-12/rootfs/opt/bitnami/scripts/libetcd.sh
```diff
@@ -307,7 +307,8 @@ etcdctl_auth_norbac_flags() {
         authFlags+=("--cert" "${ETCD_DATA_DIR}/fixtures/client/cert.pem" "--key" "${ETCD_DATA_DIR}/fixtures/client/key.pem")
     else
         [[ -f "$ETCD_CERT_FILE" ]] && [[ -f "$ETCD_KEY_FILE" ]] && authFlags+=("--cert" "$ETCD_CERT_FILE" "--key" "$ETCD_KEY_FILE")
-        [[ -f "$ETCD_TRUSTED_CA_FILE" ]] && authFlags+=("--cacert" "$ETCD_TRUSTED_CA_FILE")
+        # if CA file exists, then use CA to verify server certs; otherwise, just skip server certs verification
+        [[ -f "$ETCD_TRUSTED_CA_FILE" ]] && authFlags+=("--cacert" "$ETCD_TRUSTED_CA_FILE") || authFlags+=("--insecure-transport=false --insecure-skip-tls-verify=true")
```
andresbono, line 311: Is `--insecure-transport=false` necessary? We didn't have it set at https://github.com/bitnami/containers/blob/c4fcee416cc319efbde9b6e603f510e06ed602fc/bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/healthcheck.sh#L25

chenraoCR: ya, can remove it, and already updated it

chenraoCR: > Is `--insecure-transport=false` necessary? We didn't have it set at
> 
> https://github.com/bitnami/containers/blob/c4fcee416cc319efbde9b6e603f510e06ed602fc/bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/healthcheck.sh#L25

already updated, please kindly approve this PR

chenraoCR:
@andresbono already updated, please kindly approve this PR

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/libetcd.sh
```diff
@@ -307,7 +307,8 @@ etcdctl_auth_norbac_flags() {
         authFlags+=("--cert" "${ETCD_DATA_DIR}/fixtures/client/cert.pem" "--key" "${ETCD_DATA_DIR}/fixtures/client/key.pem")
     else
         [[ -f "$ETCD_CERT_FILE" ]] && [[ -f "$ETCD_KEY_FILE" ]] && authFlags+=("--cert" "$ETCD_CERT_FILE" "--key" "$ETCD_KEY_FILE")
-        [[ -f "$ETCD_TRUSTED_CA_FILE" ]] && authFlags+=("--cacert" "$ETCD_TRUSTED_CA_FILE")
+        # if CA file exists, then use CA to verify server certs; otherwise, just skip server certs verification
+        [[ -f "$ETCD_TRUSTED_CA_FILE" ]] && authFlags+=("--cacert" "$ETCD_TRUSTED_CA_FILE") || authFlags+=("--insecure-skip-tls-verify")
     fi
```
andresbono, line 312: I feel like there is some redundancy between setting `--insecure-skip-tls-verify` here and in the `healthcheck.sh` script. Do you think it can be simplified somehow?

For example, removing it from `healthcheck.sh` and apply the following change:

```suggestion
    fi
    # If CA file exists, then use CA to verify server certs; otherwise, just skip server certs verification
    [[ -f "$ETCD_TRUSTED_CA_FILE" ]] && authFlags+=("--cacert" "$ETCD_TRUSTED_CA_FILE") || authFlags+=("--insecure-skip-tls-verify")
    fi
```

But that could be insecure. Users not setting the recommended CA file would not notice it is missing, although it is recommended.

The other option is that you only set the flag in the `healthcheck.sh` script, changing the condition in which it is set so it fits you use case.

andresbono (commented):
I added a new comment, please let me know your thoughts.

chenraoCR:
> I added a new comment, please let me know your thoughts.

just updated as u suggested

andresbono (approved):
LGTM, thank you for your contribution!!

PR #70605, opened 2024-08-06
[bitnami/opencart] chore: :fire: :wastebasket: Deprecate container

javsalgar:
### Description of the change

This PR removes Opencart from the catalog


migruiz4 (approved):

scientes:
Quick question why was opencart removed from the catalog? i could not find the reason anywhere. Also can you recommend another image for opencart.

carrodher:
I understand your concern regarding this deprecation. Opencart has been deprecated from the Bitnami catalog because our focus has shifted towards infrastructure solutions. As a result, we are phasing out certain web application assets to better align with our strategic direction.

PR #70737, opened 2024-08-07
[bitnami/postgresql-repmgr] Support running scripts on primary node when container starts

### LINKED ISSUE #70606: [bitnami/postgresql-repmgr] Running custom script on primary database when container starting
### Name and Version

bitnami/postgresql-repmgr:16.3.0

Whenever the database container is started, I want to have some sanity check preformed on the database and possibly fix something in the database if need be.
As far as I can see there is only the docker-entrypoint.initdb.d directory but this is only sourced the first time the database is initialized and then a file will prevent scripts here ever being run again.

One usecase is infact related to upgrading the container from 16.1.0 to 16.3.0 where a collation version needs to be refreshed. I wanted to dump a script that will detect if a collation refresh is needed and then perform the necessary actions. But I cannot see any way of implementing this in the current image.

Having a change in the startup scripts that will always run the scripts in fx. /docker-entrypoint.d/*.{sh,sql} - something like the way it is done with /docker-entrypoint-initdb.d/ but without the limit on only running on initialization of database.

* manually hacking the startup scripts to implement this
* Having an external app running this (requires the app to have access to db as postgres user)

ThomasRasmussen:
### Description of the change

This PR will add support for putting scripts in /docker-entrypoint-startdb.d/ and have these executed every time the container starts (limited to the PRIMARY node). Works the same way that /docker-entrypoint-initdb.d  however this is limited to only running when database is initialized. 

### Benefits

Run validationscripts or other generic scripts every time the database container starts

### Possible drawbacks

If scripts are not idempotent then this may impact the database. 

### Applicable issues

- fixes #70606


carrodher:

ThomasRasmussen:
@carrodher  hmm... don't know why the other commits got to be a part of this PR :-( Apparently I mistankenly merged in another branch into this one... is it possible for me to remove these commits?

javsalgar:
Hi,

It seems the PR is including changes for Moodle, which I believe they are not expected. Could you check it?

PR #70999, opened 2024-08-16
[bitnami/openldap] feat: pldap and pldaps support

Arano-kai:
### Description of the change
Ability to enable pldap(s) protocols.

### Benefits
The pldap(s) is required when LDAP servers is run behind LB proxy. The proxy protocol v2 allow reveal the real client IP on LDAP server side in such setup.

### Possible drawbacks

Previous deployments not affected, this feature is disabled by default.
In new deployments that want use pldap(s) and replication, user must provide additional config to move proxy protocol enabled ports to non-default values, since replication can't run on pldap(s).

### Applicable issues

None.

### Additional information

The /opt/bitnami/scripts/openldap/run.sh are refactored to allow independent configuration and advanced logic for each startup flag. This changes are BASH specific and require additional patch if container moves to other shell.
Additional check for LDAP_PORT_NUMBER and LDAP_LDAPS_PORT_NUMBER is set, which should not affect valid deployments.

### DIFF
--- a/bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/openldap/run.sh
+++ b/bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/openldap/run.sh
@@ -37,10 +37,35 @@ flags_map["-d"]="${LDAP_LOGLEVEL}"
 flags_map["-h"]+="${flags_map["-h"]:+" "}ldapi:///"
 
 # Add LDAP URI
-flags_map["-h"]+="${flags_map["-h"]:+" "}ldap://:${LDAP_PORT_NUMBER}/"
+# Since 'proxied LDAP' default port number is same as 'LDAP',
+# enable LDAP URI when one of the following conditions are met:
+# * proxy protocol capability is disabled
+# * proxy protocol capability is enabled and proxy protocol port differ
+if ! is_boolean_yes "${LDAP_ENABLE_PROXYPROTO}" \
+    || [[ "${LDAP_PORT_NUMBER}" -ne "${LDAP_PROXYPROTO_PORT_NUMBER}" ]]
+then
+  flags_map["-h"]+="${flags_map["-h"]:+" "}ldap://:${LDAP_PORT_NUMBER}/"
+fi
 
 # Add LDAPS URI when TLS is enabled
-is_boolean_yes "${LDAP_ENABLE_TLS}" && flags_map["-h"]+="${flags_map["-h"]:+" "}ldaps://:${LDAP_LDAPS_PORT_NUMBER}/"
+# Since 'proxied LDAP over SSL' default port number is same as 'LDAP over SSL',
+# enable LDAPS URI when one of the following conditions are met:
+# * proxy protocol capability is disabled
+# * proxy protocol capability is enabled and proxy protocol tls port differ
+if is_boolean_yes "${LDAP_ENABLE_TLS}" \
+  && { ! is_boolean_yes "${LDAP_ENABLE_PROXYPROTO}" \
+    || [[ "${LDAP_LDAPS_PORT_NUMBER}" -ne "${LDAP_PROXYPROTO_LDAPS_PORT_NUMBER}" ]]; }
+then
+  flags_map["-h"]+="${flags_map["-h"]:+" "}ldaps://:${LDAP_LDAPS_PORT_NUMBER}/"
+fi
+
+# Add PLDAP URI when proxy protocol capability is enabled
+if is_boolean_yes "${LDAP_ENABLE_PROXYPROTO}"; then
+  flags_map["-h"]+="${flags_map["-h"]:+" "}pldap://:${LDAP_PROXYPROTO_PORT_NUMBER}/"
+  # Also add PLDAPS URI when TLS is enabled
+  is_boolean_yes "${LDAP_ENABLE_TLS}" \
+    && flags_map["-h"]+="${flags_map["-h"]:+" "}pldaps://:${LDAP_PROXYPROTO_LDAPS_PORT_NUMBER}/"
+fi
 
 # Build flags list
 for flag in "${!flags_map[@]}"; do
--- a/bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/openldap/run.sh
+++ b/bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/openldap/run.sh
@@ -76,4 +76,5 @@ done
 flags+=("$@")
 
 info "** Starting slapd **"
+debug "Startup cmd: ${command}" "${flags[*]}"
 exec "${command}" "${flags[@]}"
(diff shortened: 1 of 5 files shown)


andresbono:
Hi @Arano-kai, sorry for the late reply. I will review the PR and I will get back to you soon. Thank you for your contribution and patience!

bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/openldap/run.sh
```diff
@@ -21,14 +21,60 @@ command="$(command -v slapd)"
...
+am_i_root && flags_map["-u"]="${LDAP_DAEMON_USER}"
+
+# Set config dir
+flags_map["-F"]="${LDAP_CONF_DIR}/slapd.d"
+
+# Enable debug with desired level
+flags_map["-d"]="${LDAP_LOGLEVEL}"
+
+# The LDAP IPC is always on
+flags_map["-h"]+="${flags_map["-h"]:+" "}ldapi:///"
```
andresbono, line 37: Why not just simply this?

```suggestion
flags_map["-h"]="ldapi:///"
```

`flags_map["-h"]` is being declared here for the first time, so the value is going to be always `ldapi:///` right? And I think it's easier to read and understand.

bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/openldap/run.sh
```diff
@@ -21,14 +21,60 @@ command="$(command -v slapd)"
...
+
+# Add LDAP URI
+# Since 'proxied LDAP' default port number is same as 'LDAP',
+# enable LDAP URI when one of the following conditions are met:
+# * proxy protocol capability is disabled
+# * proxy protocol capability is enabled and proxy protocol port differ
+if ! is_boolean_yes "${LDAP_ENABLE_PROXYPROTO}" \
+    || [[ "${LDAP_PORT_NUMBER}" -ne "${LDAP_PROXYPROTO_PORT_NUMBER}" ]]
+then
+  flags_map["-h"]+="${flags_map["-h"]:+" "}ldap://:${LDAP_PORT_NUMBER}/"
```
andresbono, line 47: As per above comment, we are sure `flags_map["-h"]` will have a value. What if we just do this?

```suggestion
  flags_map["-h"]+=" ldap://:${LDAP_PORT_NUMBER}/"
```

Same suggestion in the concatenations below.

bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/openldap/run.sh
```diff
@@ -21,14 +21,60 @@ command="$(command -v slapd)"
...
-is_boolean_yes "$LDAP_ENABLE_TLS" && flags=("-h" "ldap://:${LDAP_PORT_NUMBER}/ ldaps://:${LDAP_LDAPS_PORT_NUMBER}/ ldapi:///")
+# Since 'proxied LDAP over SSL' default port number is same as 'LDAP over SSL',
+# enable LDAPS URI when one of the following conditions are met:
+# * proxy protocol capability is disabled
+# * proxy protocol capability is enabled and proxy protocol tls port differ
+if is_boolean_yes "${LDAP_ENABLE_TLS}" \
+  && { ! is_boolean_yes "${LDAP_ENABLE_PROXYPROTO}" \
+    || [[ "${LDAP_LDAPS_PORT_NUMBER}" -ne "${LDAP_PROXYPROTO_LDAPS_PORT_NUMBER}" ]]; }
+then
+  flags_map["-h"]+="${flags_map["-h"]:+" "}ldaps://:${LDAP_LDAPS_PORT_NUMBER}/"
```
andresbono, line 59:
```suggestion
  flags_map["-h"]+=" ldaps://:${LDAP_LDAPS_PORT_NUMBER}/"
```

bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/openldap/run.sh
```diff
@@ -21,14 +21,60 @@ command="$(command -v slapd)"
...
+fi
+
+# Add PLDAP URI when proxy protocol capability is enabled
+if is_boolean_yes "${LDAP_ENABLE_PROXYPROTO}"; then
+  flags_map["-h"]+="${flags_map["-h"]:+" "}pldap://:${LDAP_PROXYPROTO_PORT_NUMBER}/"
```
andresbono, line 64:
```suggestion
  flags_map["-h"]+=" pldap://:${LDAP_PROXYPROTO_PORT_NUMBER}/"
```

bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/openldap/run.sh
```diff
@@ -21,14 +21,60 @@ command="$(command -v slapd)"
...
+  # Also add PLDAPS URI when TLS is enabled
+  is_boolean_yes "${LDAP_ENABLE_TLS}" \
+    && flags_map["-h"]+="${flags_map["-h"]:+" "}pldaps://:${LDAP_PROXYPROTO_LDAPS_PORT_NUMBER}/"
```
andresbono, line 67:
```suggestion
    && flags_map["-h"]+=" pldaps://:${LDAP_PROXYPROTO_LDAPS_PORT_NUMBER}/"
```

andresbono (commented):
I left some suggestions, could you please take a look at them? Thanks!

bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/openldap/run.sh
Arano-kai: This is just stable idiom to insert arg in multi-valued flag, which in turn give possibility to move/add/remove additional "blocks" between declare and final assembling without affecting overall flow.
Idea that is:
- single-valued flags sets as `flags_map["<flag>"]="<arg>"`, so the last setter wins.
- multi-valued flags sets as `flags_map["<flag>"]+="${flags_map["<flag>"]:+"<delim>"}<arg>"`, so args line compiled natively and "block" position only affect arg sorting for particular flag.

bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/openldap/run.sh
Arano-kai: Surely for now `flags_map["-h"]` have value, but in case:
- incorporate additional logic (e.g: enabling ldapi:// by condition)
- rearrange "blocks"
- new protocol emerge (e.g: ldap 9.1 now support uldap:// -- LDAP over UDP :upside_down_face: )
all that required -- refactor/move/add "logic block" without deep analyze neighborhood "blocks".

If you think this is overkill, i'm no worries to accept proposed suggestions.

bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/openldap/run.sh
andresbono: I see your point 👍

andresbono (approved):
LGTM! Thank you for your contribution!

PR #71168, opened 2024-08-23
[bitnami/airflow] fix LD_LIBRARY_PATH

alekseyolg:
In the environment variables LD_LIBRARY_PATH of images, a variable is specified whose path does not exist inside the file system of containers. This pr solves this problem

UPD!: Update to python3.12
image

### Applicable issues

- fixes #


alekseyolg:
@jotamartos I would be grateful for any feedback.

jotamartos (approved):
Thanks for taking the time to create this PR. I'm going to update our internal systems and build a new image with this change. 

Thanks

PR #72279, opened 2024-09-09
[bitnami/redis] User Environmental Variables

### LINKED ISSUE #70958: [bitnami/redis-cluster] Cannot create Redis cluster with other than default user
### Name and Version

bitnami/redis-cluster:7.2.5

It's a common security best practice that the default administrator username is disabled. The ability to do so is explicitly stated in [Redis security documentation](https://redis.io/docs/latest/operate/oss_and_stack/management/security/acl/). This can be achieved in Redis cluster by creating new users into ACL and setting `masteruser` and `masterauth` configurations for replication. The ACL file can be created before-hand with another instance of redis-cluster.

However, there doesn't seem to be any way to set custom admin user name into bitnami redis-cluster creation. You have got the REDIS_PASSWORD environment variable, but that assumes default user for all operations.

If I disable the default user with ACL, cluster creation fails and docker log is filled with `AUTH failed: WRONGPASS invalid username-password pair or user is disabled.`.

I'm using bitnami/redis-cluster:7.2.5 on Docker. I switched to the 7.4.0 that was release a few days ago, but since it's source is not yet available in Github, I cannot tell if there are any new features regarding this.

New environmental variables that can be configured through docker like the rest of them:

- "REDIS_USER". Default value "default".
- "REDIS_MASTER_USER". Default value inherits from "REDIS_USER".
- "REDIS_REQUIREPASS". Default value inherits from "REDIS_PASSWORD". This would be for backwards compatibility only, so if there is a cleaner approach, go for it.

And logic changes to scripts:

- All usages of this new variable should be paired with "--user" parameter in redis-cli.
- All existing "REDIS_PASSWORD" parameters should be paired with "--pass" parameter, instead of "-a" parameter, in redis-cli to indicate a change.
- If "REDIS_REQUIREPASS" is set, use it for current "--requirepass" parameters. If it's empty, don't set "--requirepass" and trust that the user has set redis.conf and ACL list accordingly.
- Use "REDIS_MASTER_USER" and "REDIS_MASTER_PASSWORD" (existing(?)) appropriately with "--masteruser" and "--masterauth" parameters. If the parameters are empty, they should have no effect and trust that user has set redis.conf and ACL list on all nodes accordingly. Right now there seems to be erronous usage in https://github.com/bitnami/containers/blob/main/bitnami/redis-cluster/7.2/debian-12/rootfs/opt/bitnami/scripts/redis-cluster/run.sh lines 28-31

There should be a possibility to pass arbitrary parameters into the start up commands. There currently is nothing to accomplish that with.

seppo498573908457:
### Description of the change

Added new environmental variables: "REDIS_REQUIREPASS", "REDIS_USER" and "REDIS_MASTER_USER". Modified default value for "REDIS_MASTER_PASSWORD".
Modified scripts to make use of the new variables and retaining backwards compatibility.
Modified the README to reflect changes.

Fixed a bug, "--masterauth" now correctly receives value from "REDIS_MASTER_PASSWORD".

### Benefits

Allows use of ACL users in the environmental variables.
Retains backwards compatibility with old configurations by using default values for new variables and this is considered in script changes.

### Applicable issues

- fixes #70958

### Additional information

Similar changes were made to "redis-cluster" image on a different PR.


andresbono:
Hi @seppo498573908457, thank you for submitting this PR. In the meantime we review your contribution and give you some feedback, we would need the changes to be applied in all the branches, that means `6.2` as well. Are your proposed changes compatible with `6.2`?

seppo498573908457:
It looks like your script files have not changed on relevant parts, ie. they should be compatible. I will still need to test it out and check some of the Redis settings...

PR #72365, opened 2024-09-12
Replication Manager flag to enable consult each other to build a quorum

### LINKED ISSUE #63819: Add a primary_visibility_consensus flag to postgresql-repmgr
### Name and Version

postgresql-repmgr:16

Using the configuration with two pg-0 (master), pg-1(standby) and one witness pgw-0 in the middle doesn't solve the split brain problem.

If pg-0 and pg-1 are located at different DC(1,2), while pgw-0 located at DC3. In the case of network outage between dc1 and dc2, and whilst pgw-0 can still observe both pg-0 and pg-1, it won't help. pg-1 stands to Master role, and it doesn't consider any information from the witness. 

In order to change it, we can config repmgr on all node with -  `primary_visibility_consensus=true` flag [(repmgrd manual),](https://www.repmgr.org/docs/4.4/repmgrd-primary-visibility-consensus.html). And add a few `reconnect_attempts='3'` on all nodes, so pg-1 will get the info from pgw-0 before it decide to promote itself to master. In this case pg-1 will show logs: 
```
[2024-03-14 11:38:05] [NOTICE] cancelling failover as some nodes can still see the primary
[2024-03-14 11:38:05] [NOTICE] election cancelled
```
And will wait for primary.

Shagrat2:
### Description of the change

bitnami/postgresql-repmgr

Append env REPMGR_PRIMARY_VISIBILITY_CONSENSUS to repmgr_generate_repmgr_config()

### Benefits

repmgr.conf - primary_visibility_consensus
When the primary becomes unavailable in a multi-node cluster, the standbys can consult each other to build a quorum about a failover. This is done by asking each standby about the time it last saw the primary. If a node’s last communication was very recent and later than the time the local node saw the primary, the local node assumes the primary is still available, and does not go ahead with a failover decision. 

### Possible drawbacks

no

### Applicable issues

- fixes [Add a primary_visibility_consensus flag to postgresql-repmgr #63819](https://github.com/bitnami/containers/issues/63819)


jotamartos (approved):
Changes LGTM! Thank you for your contribution

PR #73141, opened 2024-10-07
[bitnami/keycloak] Added aliases for environment variables from Keycloak documentation

Fydon:
### Description of the change

Currently some of the environment variable mentioned in the [Keycloak documentation](https://www.keycloak.org/server/all-config) have aliases in the scripts, but others do not. I followed the changes from #72481 to build this pull request.

### Benefits

Transferring an existing setup from the Keycloak containers to the Bitnami setup takes less effort, although there are differences.

### Possible drawbacks

It may make the setup differences less obvious, but then all aliases should be removed.

### Applicable issues

Currently none.

### Additional information

Keycloak 26 has been released. When the new release is created there may be new environment variables that can have aliases.


migruiz4:
Hi @Fydon,

Thank you very much for your contribution!

Nevertheless, I don't think we can merge this feature.

First, I would like to clarify the difference between `KEYCLOAK_` and `KC_` environment variables.

The `KEYCLOAK_` env variables correspond to variables used by the Bitnami initialization logic (libkeycloak.sh), while the `KC_` env variables correspond to variables directly used by the Keycloak binary.

For context, many of the `KEYCLOAK_` were implemented before the `KC_` variables existed, so that is why there are duplicities between some of them.

Our goal now is to start removing the logic behind for some of the `KEYCLOAK_` variables and use the `KC_` instead while keeping the `KEYCLOAK_` env var as an alias for backward-compatibility purposes.

The reason why I can not accept this PR is because it performs the opposite we are looking for. It aliases `KC_` env variables to `KEYCLOAK_` variables, which may cause issues due to the keycloak binary receiving configuration duplicated (from settings configured by the init logic and the KC_ env variable).

I'm sorry for the inconvenience.

PR #73187, opened 2024-10-09
[bitnami/mongodb] add allow remote connection env

sanbei101:
**added a environment variable to allow remote connections**

### DIFF
--- a/bitnami/jwt-cli/6/debian-12/Dockerfile
+++ b/bitnami/jwt-cli/6/debian-12/Dockerfile
@@ -7,11 +7,11 @@ ARG TARGETARCH
 
 LABEL com.vmware.cp.artifact.flavor="sha256:c50c90cfd9d12b445b011e6ad529f1ad3daea45c26d20b00732fae3cd71f6a83" \
       org.opencontainers.image.base.name="docker.io/bitnami/minideb:bookworm" \
-      org.opencontainers.image.created="2024-09-11T00:30:09Z" \
+      org.opencontainers.image.created="2024-10-11T00:52:07Z" \
       org.opencontainers.image.description="Application packaged by Broadcom, Inc." \
       org.opencontainers.image.documentation="https://github.com/bitnami/containers/tree/main/bitnami/jwt-cli/README.md" \
       org.opencontainers.image.licenses="Apache-2.0" \
-      org.opencontainers.image.ref.name="6.1.1-debian-12-r2" \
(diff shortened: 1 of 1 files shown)


bitnami/mongodb/8.0/debian-12/rootfs/opt/bitnami/mongodb/templates/mongodb.conf.tpl
```diff
@@ -25,7 +25,7 @@ net:
     enabled: true
     pathPrefix: {{MONGODB_TMP_DIR}}
   ipv6: {{MONGODB_DEFAULT_ENABLE_IPV6}}
-  bindIpAll: false
+  bindIpAll: {{ALLOW_REMOTE_CONNECTION}}
```
juan131, line 28: This is a default setting, hence any env. var modifying this behaviour should be prefixed with `MONGODB_DEFAULT_`.

That said, please note this won't allow you to edit the value for `bindIpAll` dynamically on runtime. This template is only used when building the container image, see:

- https://github.com/bitnami/containers/blob/4fe538c531073079670e646b7ef9ddabded23d15/bitnami/mongodb/8.0/debian-12/rootfs/opt/bitnami/scripts/mongodb/postunpack.sh#L25

In order to customize the configuration file at runtime, we recommend you to follow the instruction below instead:

- https://github.com/bitnami/containers/tree/main/bitnami/mongodb#configuration-file

juan131 (changes requested):

bitnami/mongodb/8.0/debian-12/rootfs/opt/bitnami/mongodb/templates/mongodb.conf.tpl
sanbei101: I understand what you mean, but in some cases, we really need a quick way to create a MongoDB instance that can be accessed remotely, even though it won't be changed during runtime. It seems that bitnami/postgres follows the same logic.

juan131: I don't think that's a reason to configure MongoDB to bind **by default** all IPv4 and IPv6 addresses. I understand it's sth useful for your use case but it's not a secure setting that should be used except for certain development purposes.

That's why, I see to options:

- Adapt this PR so this setting is adapted during the container initialization (instead of doing so at build time).
- Use a custom configuration following the instructions I shared before.

PR #73489, opened 2024-10-22
[bitnami/redis-cluster] Allow setting IP_VERSION for DNS lookup

mtenrero:
### Description of the change

Allows to set IP Lookup mode 

### Benefits

It's required for DualStack environments in the case it's some stack preference. Specially on OpenShift
### Applicable issues

DualStack networks 
Helm Chart

### DIFF
--- a/bitnami/redis-cluster/7.4/debian-12/prebuildfs/opt/bitnami/scripts/libnet.sh
+++ b/bitnami/redis-cluster/7.4/debian-12/prebuildfs/opt/bitnami/scripts/libnet.sh
@@ -39,8 +39,9 @@ wait_for_dns_lookup() {
     local hostname="${1:?hostname is missing}"
     local retries="${2:-5}"
     local seconds="${3:-1}"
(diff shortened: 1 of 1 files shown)


jotamartos:
Hi,

Thank you for taking the time to contribute. The main problem I see with this change is that there will not be any documentation about that env var and nobody will know that it's possible to use it. Let us review the changes and come up with a better alternative to fix this problem.

jotamartos (changes requested):
Hi again, 

As I mentioned above, using the IP_VERSION env var will difficult its usage as it won't be documented anywhere. However, I'd make the following changes

- add the ip_version parameter as argument in each function. I'd include as first/second parameter depending on if host/hostname is also an argument of the function.
- The default value if the argument is not provided will be empty
- Add a new env var in the redis-cluster-env.sh script that allows you set the IP version (REDIS_DNS_LOOKUP_IP_VERSION or similar)
- Update the librediscluster.sh script to use that env var as argument of the different methods.

Let me know if you have any questions.

Thanks

PR #73541, opened 2024-10-23
[bitnami/postgresql-pgpool] use correct node port on healthcheck

### LINKED ISSUE #72979: Incorrect PGOOL_PORT_NUMBER used for PostgreSQL node connection in PgPool
### Name and Version

bitnami/pgpool:4.5.4

None

There is an issue with the `PGPOOL_PORT_NUMBER` variable in the bitnami/pgpool:4.5.4 container image, where the wrong port is used for connecting to a PostgreSQL node. Instead of connecting to the correct port of a node, the PgPool script incorrectly attempts to connect to the port of PgPool itself.

https://github.com/bitnami/containers/blob/7f4dbf453c2a132f3b14e4edaf9402efbb47a3e8/bitnami/pgpool/4/debian-12/rootfs/opt/bitnami/scripts/libpgpool.sh#L319

healthcheck.sh is connecting to the wrong port, leading to incorrect behaviour when interacting with PostgreSQL nodes.

### What do you see instead?

Connect to wrong port.

### Additional information

Suggested Fix
In the file libpgpool.sh, line 319, the following change is recommended:
```bash
local node_port="${node_info[2]}"

            if [[ $(PGCONNECT_TIMEOUT=3 PGPASSWORD="${PGPOOL_POSTGRES_PASSWORD}" psql -U "${PGPOOL_POSTGRES_USERNAME}" \
                -d postgres -h "${node_host}" -p "${node_port}" -tA -c "SELECT 1" || true) == 1 ]]; then
```

yukha-dw:
### Description of the change

Based on @AlexShander's finding on #72979, pgpool's healthcheck uses its port rather than postgresql's port when checking connection to them

### Benefits

Let user to set different port between pgpool and postgresql

### Applicable issues

- fixes #72979

### DIFF
--- a/bitnami/pgpool/4/debian-12/rootfs/opt/bitnami/scripts/libpgpool.sh
+++ b/bitnami/pgpool/4/debian-12/rootfs/opt/bitnami/scripts/libpgpool.sh
@@ -315,8 +315,9 @@ pgpool_healthcheck() {
             IFS="|" read -ra node_info <<< "$node"
             local node_id="${node_info[0]}"
             local node_host="${node_info[1]}"
+            local node_port="${node_info[2]}"
             if [[ $(PGCONNECT_TIMEOUT=3 PGPASSWORD="${PGPOOL_POSTGRES_PASSWORD}" psql -U "${PGPOOL_POSTGRES_USERNAME}" \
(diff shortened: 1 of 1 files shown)


jotamartos (approved):
Changes LGTM! Thank you for your contribution.

Let us sync the changes in our system and launch a build before merging this change.

PR #73542, opened 2024-10-23
[bitnami/postgresql-repmgr] use passfile on primary_conninfo

yukha-dw:
### Description of the change

Currently, `postgresql.conf` stores literal password on `primary_conninfo`, this could be a security issue

### Benefits

Let user to hide their password on `primary_conninfo` on `postgresql.conf` (often visible to anyone, 644) using `passfile`

### Possible drawbacks

I haven't checked much about `POSTGRESQL_REPLICATION_` envs, `POSTGRESQL_REPLICATION_USE_PASSFILE` isn't used at all and I don't know its background


fmulero (commented):
Thanks a lot @yukha-dw for your contribution!

I completely agree with your approach, this is a very interesting feature. I have some concerns that I'd like to discuss or suggest:
* File associated to `POSTGRESQL_REPLICATION_PASSFILE_PATH` seems not created by default, Should users provide it?
* I see an overlapping with `REPMGR_USE_PASSFILE` and `REPMGR_PASSFILE_PATH` env variables in lib. Not sure how they work together, have you seen [this](https://github.com/bitnami/containers/blob/69b7934deb4535814feb541fcb18f707343263f8/bitnami/postgresql-repmgr/12/debian-12/rootfs/opt/bitnami/scripts/postgresql-env.sh#L447-L448)? 
* Could you add any documentation in the README file about the behaviour/use of the env variables?
* Could you provide the steps you  followed to test the solution?

yukha-dw:
From what I understand, `POSTGRESQL_REPLICATION_*` variables should be identic with `REPMGR_*`, except for `POSTGRESQL_REPLICATION_PASSFILE_PATH` which somehow we can overwrite even though we are using same username stated in `REPMGR_USERNAME`. This could be a remnant of short circuiting the configuration (also see `POSTGRESQL_REPLICATION_USE_PASSFILE`, it doesn't being used anywhere)
https://github.com/bitnami/containers/blob/69b7934deb4535814feb541fcb18f707343263f8/bitnami/postgresql-repmgr/12/debian-12/rootfs/opt/bitnami/scripts/postgresql-env.sh#L450-L457

> File associated to POSTGRESQL_REPLICATION_PASSFILE_PATH seems not created by default, Should users provide it?

The file associated to `POSTGRESQL_REPLICATION_PASSFILE_PATH` must contain same content with `REPMGR_PASSFILE_PATH`. It's better to leave it alone. By default, when `REPMGR_USE_PASSFILE` is set to `true`, this container will create the required passfile when user hasn't provide one.
https://github.com/bitnami/containers/blob/c41d3ebc68f4542ec23d20721adce816e304140a/bitnami/postgresql-repmgr/12/debian-12/rootfs/opt/bitnami/scripts/librepmgr.sh#L493-L496

> I see an overlapping with REPMGR_USE_PASSFILE and REPMGR_PASSFILE_PATH env variables in lib. Not sure how they work together, have you seen [this](https://github.com/bitnami/containers/blob/69b7934deb4535814feb541fcb18f707343263f8/bitnami/postgresql-repmgr/12/debian-12/rootfs/opt/bitnami/scripts/postgresql-env.sh#L447-L448)?

`REPMGR_USE_PASSFILE` is a boolean to determine if user want to use passfile or plain text password, while `REPMGR_PASSFILE_PATH` is the path of the passfile. User can provide its own passfile or let the container create one from them.
https://github.com/bitnami/containers/blob/c41d3ebc68f4542ec23d20721adce816e304140a/bitnami/postgresql-repmgr/12/debian-12/rootfs/opt/bitnami/scripts/librepmgr.sh#L493-L496

> Could you add any documentation in the README file about the behaviour/use of the env variables?

I am not sure how to do this without introducing any breaking change. Current README is quite coupled with bitnami/charts. My apology.

> Could you provide the steps you followed to test the solution?

I simply test this by altering `REPMGR_USE_PASSFILE` value.

Before change,
`REPMGR_USE_PASSFILE=false` -> repmgr use PGPASSWORD, postgresql_configure_recovery use PGPASSWORD
`REPMGR_USE_PASSFILE=true` -> repmgr use PGPASSFILE, postgresql_configure_recovery use PGPASSWORD

After change,
`REPMGR_USE_PASSFILE=false` -> repmgr use PGPASSWORD, postgresql_configure_recovery use PGPASSWORD
`REPMGR_USE_PASSFILE=true` -> repmgr use PGPASSFILE, postgresql_configure_recovery use **PGPASSFILE**

yukha-dw:
I've added few changes here https://github.com/bitnami/containers/pull/73542/commits/b35bd301a9d3588c3be9a850db764438ab464a00:
1. Update old postgresql versions (12, 13, 14, & 15)
2. Misconfig prevention - Prevent user from overwriting `POSTGRESQL_REPLICATION_PASSFILE_PATH`. `POSTGRESQL_REPLICATION_PASSFILE_PATH` always will be alias of `REPMGR_PASSFILE_PATH`
3. Update README - Move related variables to read-only section (as it is intended to)
4. Update README - Add this notice:
> NOTE: `repmgr` user is also used as PostgreSQL replication user.

fmulero:
Thanks a lot @yukha-dw for such detailed explanation, I really appreciate it!

Your code is correct for `postgresql-repmgr` but I've been testing the solution in our environments and I've seen some problems in the `postgresql` container (some parts of the code are common between both containers) because in case of replication the slave postgresql container will try to use the passfile (`POSTGRESQL_REPLICATION_PASSFILE_PATH`) and it doesn't exist. To don't bother you with our dependencies I've opened an internal task to find a solution. I'll keep you posted on.

fmulero:
Similar changes has been applied and released. Could you give them a glance? For more details please see:
* https://github.com/bitnami/containers/pull/74782
* https://github.com/bitnami/containers/pull/74777

yukha-dw:
@fmulero Thank you! both seems fine to me. I'll close this PR

Issue #70769, opened 2024-08-08
[bitnami/schema-registry] Support IAM auth way


hongbo-miao:
### Name and Version

bitnami/schema-registry:7.7.0

### What is the problem this feature will solve?

I am hoping to use AWS IAM auth way according to

- https://github.com/confluentinc/schema-registry/issues/1898#issuecomment-1289526114
- https://stackoverflow.com/questions/70500586/using-confluent-schema-registry-with-msk/70522089#70522089

But then found

- `SCHEMA_REGISTRY_KAFKASTORE_SECURITY_PROTOCOL`
- `SCHEMA_REGISTRY_KAFKASTORE_SASL_MECHANISM`
- `SCHEMA_REGISTRY_KAFKASTORE_SASL_JAAS_CONFIG`
- `SCHEMA_REGISTRY_KAFKASTORE_SASL_CLIENT_CALLBACK_HANDLER_CLASS`

do not exist in https://github.com/bitnami/containers/tree/main/bitnami/schema-registry#configuration

It only exist in [confluentinc/cp-schema-registry](https://hub.docker.com/r/confluentinc/cp-schema-registry).

### What is the feature you are proposing to solve the problem?

It would be great to support similar envs so that we can use IAM auth way, thanks!

### What alternatives have you considered?

Use [confluentinc/cp-schema-registry](https://hub.docker.com/r/confluentinc/cp-schema-registry) Docker image.

carrodher:
The environment variables used by the bitnami/schema-registry container are listed at https://github.com/bitnami/containers/tree/main/bitnami/schema-registry#environment-variables. Although the exact name does not have to match other schema-registry containers, there is likely some variable for the same purpose.

hongbo-miao:
Hi @carrodher sorry, I forgot to put that link in the description. I actually tried to find matched env in the list. But still cannot find these four corresponding ones 🥲

- `SCHEMA_REGISTRY_KAFKASTORE_SECURITY_PROTOCOL`
- `SCHEMA_REGISTRY_KAFKASTORE_SASL_MECHANISM`
- `SCHEMA_REGISTRY_KAFKASTORE_SASL_JAAS_CONFIG`
- `SCHEMA_REGISTRY_KAFKASTORE_SASL_CLIENT_CALLBACK_HANDLER_CLASS`

javsalgar:
Hi!

[...]

PR #74014, opened 2024-10-28
[binatmi/schema-registry] Add support for Amazon MSK IAM auth

bpesics:
### Description of the change

Add Amazon MSK IAM authentication support.

### Benefits

Allows Schema Registry to use AWS IAM for authentication and authorization against Amazon MSK clusters that have AWS IAM enabled as an authentication mechanism. This is obviously more desired than no authentication or plain text with static credentials.

### Possible drawbacks

N/A

### Applicable issues

- implements: https://github.com/bitnami/containers/issues/70769
- related to: https://github.com/bitnami/charts/issues/28766

### Additional information

As a side issue, this image insisted on mounting a keystore file for any `SECURITY_PROTOCOL` [which includes `SSL` or `SASL`](https://github.com/bitnami/containers/blob/045686e0991f5ea239cdd4ea988cdf72999afb67/bitnami/schema-registry/7.7/debian-12/rootfs/opt/bitnami/scripts/libschemaregistry.sh#L145-L160), presumably  for TLS client authentication. 
 In order to avoid breaking the existing behaviour I added `SCHEMA_REGISTRY_KAFKASTORE_CLIENT_AUTH_DISABLED` to provide a way to disable validation when mounting keystore files isn't really necessary.


migruiz4:
Hi @bpesics,

Thank you very much for your contribution! Before I can accept this feature, I would need to ask you about the following information:
- Could you please provide Schema Registry documentation where support for  AWS MSK is mentioned?
- What would be compatibility restrictions between the AWS MSK jar and Schema registry versions? 

I'm sorry but without that information, this would need to remain as a customization on users' side.

migruiz4:
We could accept adding support for the `SCHEMA_REGISTRY_KAFKASTORE_*` environment variables, but we need the information requested in order to add the AWS_MSK  jar file to the source image.

michalmisiewicz:
I encountered a similar issue with Google Cloud Managed Kafka and proposed a provider-agnostic solution in #74972. I believe the best way to support custom `sasl.mechanism` is by providing a configuration file specified in SCHEMA_REGISTRY_CONF_FILE.

bpesics:
> I encountered a similar issue with Google Cloud Managed Kafka and proposed a provider-agnostic solution in #74972. I believe the best way to support additional `sasl.mechanism` is by providing a configuration file specified in SCHEMA_REGISTRY_CONF_FILE.

@michalmisiewicz Nice, thanks for the heads-up! Your PR addressed what I refer to here as "client authentication" issue.

> I encountered a similar issue with Google Cloud Managed Kafka and proposed a provider-agnostic solution in #74972. I believe the best way to support additional `sasl.mechanism` is by providing a configuration file specified in SCHEMA_REGISTRY_CONF_FILE.

Thanks! Your PR addressed what I referred to here as "client authentication" issue, that should simplify things.

On the other hand, `SCHEMA_REGISTRY_CONF_FILE` is a read-only variable and the only way to pass in configuration from outside is mounting a file with the same relative path as `SCHEMA_REGISTRY_CONF_FILE` with `SCHEMA_REGISTRY_MOUNTED_CONF_DIR`. Using this injected file pretty much none of the functionality of this image works anymore. And the Amazon MSK Library still has to be baked in somehow.

PR #74018, opened 2024-10-28
allow to increase the db_maxconn of the connection pool using an environment variable

pedominguezbr:
Update the line with; `db_maxconn = 64`, to something like: `db_maxconn = {{ODOO_DATABASE_MAXCONN}}` to allow the `db_maxconn` to be assigned via environment variable, since the parameter does not exist in the odoo cmdline


dgomezleon (changes requested):
Hi @pedominguezbr,

Thanks for contributing.

You would need to apply this for all the odoo versions. Also you will need to add this to `README.md` and `odoo-env.sh` as other environment variables do.

Issue #74017, opened 2024-10-28
valkey sentinel is broken


amirreza8002:
### Name and Version

bitnami/valkey-sentinel:7.2.7

### What architecture are you using?

None

### What steps will reproduce the bug?

running valkey:7.2.7 fails: `Unrecognized sentinel configuration statement.`

### What do you see instead?

seems like the sentinel conf is not supported by valkey:7.2.7
worth noting that this is a new error(3 days ago), the same tests had no problems some time ago

### Additional information

[7.2.7](https://github.com/amirreza8002/django-valkey/actions/runs/11561620322/job/32181035918) fails
[latest](https://github.com/amirreza8002/django-valkey/actions/runs/11561620322/job/32181035551) passes

jotamartos:
We are working on building an updated version of the container and adding new test to our container

jotamartos:
[7.2.7-debian-12-r3](https://github.com/bitnami/containers/pull/74092) was released with the changes to fix this issue. Did you have the chance to take a look at it?

PR #74090, opened 2024-10-31
[bitnami/valkey-sentinel] Update tests to check the master to primary change in the configuration

jotamartos:
### Description of the change
Add test to validate we are not adding both parameters to the configuration

### Benefits
Will find errors like this one earlier 

https://github.com/bitnami/containers/issues/74017

### Possible drawbacks
None

### Additional information
This test simply checks that we do not accidentally add both conf parameters to the conf file

### DIFF
--- a/.vib/valkey-sentinel/goss/valkey-sentinel.yaml
+++ b/.vib/valkey-sentinel/goss/valkey-sentinel.yaml
@@ -13,7 +13,7 @@ file:
 command:
   # Remove if sentinel gets updated: https://github.com/valkey-io/valkey/blob/7.2/sentinel.conf
   check-primary-master-change:
-    exec: if [ $(grep "SENTINEL \(master-reboot-down-after-period\|primary-reboot-down-after-period\)" /tmp/sentinel.conf | wc -l) -gt 1 ]; then exit 1; fi
+    exec: if [ $(grep "SENTINEL \(master-reboot-down-after-period\|primary-reboot-down-after-period\)" /opt/bitnami/valkey-sentinel/etc/sentinel.conf | wc -l) -gt 1 ]; then exit 1; fi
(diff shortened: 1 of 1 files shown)


carrodher (dismissed):

.vib/valkey-sentinel/goss/valkey-sentinel.yaml
```diff
@@ -11,6 +11,10 @@ file:
       - "daemonize no"
       - 'logfile ""'
 command:
+  # Remove if sentinel gets updated: https://github.com/valkey-io/valkey/blob/7.2/sentinel.conf
+  check-primary-master-change:
+    exec: if [ $(grep "SENTINEL \(master-reboot-down-after-period\|primary-reboot-down-after-period\)" /tmp/sentinel.conf | wc -l) -gt 1 ]; then exit 1; fi
+    exit-status: 0
   check-valkey-server:
```
jotamartos, line 18: This test shouldn't be working. Checking why it succeeded 

```
valkey-sentinel-1  | valkey-sentinel 15:58:19.88 INFO  ==> ** Starting Valkey Sentinel **
valkey-sentinel-1  | 
valkey-sentinel-1  | *** FATAL CONFIG FILE ERROR (Version 7.2.7) ***
valkey-sentinel-1  | Reading the configuration file, at line 358
valkey-sentinel-1  | >>> 'SENTINEL master-reboot-down-after-period mymaster 0'
valkey-sentinel-1  | No such master with specified name.
```

jotamartos:
The initialization is not executed in this test so that's why we didn't see the error. Will think about a different way to test this change

jotamartos:
Reopening PR after updating the tests.

dgomezleon (approved):

PR #74112, opened 2024-11-02
[bitnami/kong] Remove validation of non-existent variable

### LINKED ISSUE #74031: [bitnami/kong] Invalid `KONG_POSTGRESQL_PASSWORD_FILE`
### Name and Version

bitnami/kong:3.8.0-debian-12-r8

None

Shouldn't `KONG_POSTGRESQL_PASSWORD_FILE` be `KONG_PG_PASSWORD_FILE`? `KONG_POSTGRESQL_PASSWORD` is not an alias of `KONG_PG_PASSWORD`.

https://github.com/bitnami/containers/blob/95dd8afa0efa48cbcce76a9b8f78b44a2251cab6/bitnami/kong/3/debian-12/rootfs/opt/bitnami/scripts/libkong.sh#L83

One way to fix it perhaps by adding this line on `kong-env.sh`
```
kong_env_vars=(
    ...
    KONG_PG_PASSWORD
    KONG_POSTGRESQL_PASSWORD
)
...
KONG_PG_PASSWORD="${KONG_PG_PASSWORD:-"${KONG_POSTGRESQL_PASSWORD:-}"}"
export KONG_PG_PASSWORD="${KONG_PG_PASSWORD:-}"
```

`KONG_POSTGRESQL_PASSWORD_FILE` becomes alias of `KONG_PG_PASSWORD_FILE`

### What do you see instead?

Container validates `KONG_POSTGRESQL_PASSWORD_FILE`, but does not use it

### Additional information

`KONG_POSTGRESQL_PASSWORD_FILE` is also mentioned in bitnami/charts

yukha-dw:
### Description of the change

Remove validation of non-existent variable `KONG_POSTGRESQL_PASSWORD_FILE`. The validation only introduces confusion and doesn't work as intended

### Benefits

Reduce confusion when user reads the source code

### Possible drawbacks

No drawback

### Applicable issues

- fixes #74031

### Additional information

If anyone want to store the PASSWORD on file instead of env, use `KONG_PG_PASSWORD_FILE` instead. In fact, you can do this to almost any variable by using `_FILE` suffix
https://github.com/bitnami/containers/blob/e26969165953164cec8a8fb5293375a3af83f29e/bitnami/kong/3/debian-12/rootfs/opt/bitnami/scripts/kong-env.sh#L26-L54

### DIFF
--- a/bitnami/kong/3/debian-12/rootfs/opt/bitnami/scripts/libkong.sh
+++ b/bitnami/kong/3/debian-12/rootfs/opt/bitnami/scripts/libkong.sh
@@ -42,12 +42,6 @@ kong_validate() {
         fi
     }
 
-    check_password_file() {
-        if [[ -n "${!1:-}" ]] && ! [[ -f "${!1:-}" ]]; then
-            print_validation_error "The variable ${1} is defined but the file ${!1} is not accessible or does not exist"
-        fi
-    }
-
     check_resolved_hostname() {
         if ! is_hostname_resolved "$1"; then
             warn "Hostname ${1} could not be resolved, this could lead to connection issues"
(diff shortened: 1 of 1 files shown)


jotamartos (changes requested):
Thanks for taking the time to contribute. Unfortunately, we can't merge this change because it'll break the deployment of the Chart solution 

https://github.com/search?q=repo%3Abitnami%2Fcharts%20KONG_POSTGRESQL_PASSWORD_FILE&type=code

The Chart will need to be updated before releasing the container with this change.

yukha-dw:
@jotamartos got it! in the mean time, I'll close this PR

PR #74166, opened 2024-11-06
[bitnami/pgpool] Remove deprecated users from pg_hba.conf

### LINKED ISSUE #73526: [bitnami/pgpool] `pool_hba.conf` contains unknown users?
### Name and Version

bitnami/pgpool:4.5.4-debian-12-r2

amd64

It's not really a bug but I'm not sure where to ask. On default `pool_hba.conf`, current script let `pop_user` & `wide` users to be accessed without auth (`trust`). Is it intentional or remnant of previous version? I don't see any reference here nor pgpool's source code. Thank you in advance.

https://github.com/bitnami/containers/blob/2de4377adaaa5ba0ba682bc1aa62747cea63bc22/bitnami/pgpool/4/debian-12/rootfs/opt/bitnami/scripts/libpgpool.sh#L365-L372

`pool_hba.conf` doesn't contain give free pass to unknown users

### What do you see instead?

`pool_hba.conf` gives free pass to unknown users

### Additional information

_No response_

yukha-dw:
### Description of the change

Remove deprecated users `wide` & `pop_users` from `pg_hba.conf`. Those users should be safe to be removed from `pg_hba.conf` as both are not used anymore.

### Benefits

Increase security by removing unwanted access

### Possible drawbacks

Nothing that I can think of

### Applicable issues

- fixes #73526

### DIFF
--- a/bitnami/pgpool/4/debian-12/rootfs/opt/bitnami/scripts/libpgpool.sh
+++ b/bitnami/pgpool/4/debian-12/rootfs/opt/bitnami/scripts/libpgpool.sh
@@ -366,8 +366,6 @@ EOF
     cat >>"$PGPOOL_PGHBA_FILE" <<EOF
 ${sr_check_auth_line}
 ${postgres_auth_line}
-host     all             wide               all         trust
-host     all             pop_user           all         trust
 host     all             all                all         ${all_authentication}
 EOF
 }


jotamartos (approved):
Thank you for your contribution

PR #74924, opened 2024-11-15
[bitnami/redis] Install libssl1 package required for redis-stack (loadmodule /opt/bitnami/redis/lib/redistimeseries.so)

rajdudhare1:
Install libssl1 package required for redis-stack (redis time-series  module)

### Description of the change

The suggestion is to either add libssl package to the image or allow options which load modules much like redis-stack

### Benefits

Whenever the new redis-stack is spun up from bitnami helm chart it will work, right now it's failing when the time-series module is loaded in the redis cluster.

### Possible drawbacks

Image size will be increased by couple of MBs.

### Applicable issues

When using bitnami helm charts to create a Redis cluster or instance with the time-series module, the cluster cannot load modules due to missing libssl.

### Additional information

https://github.com/bitnami/charts/issues/27854


jotamartos (changes requested):
I'm sorry @rajdudhare1 but I can't accept this change. [OpenSSL 1.1.1 is no longer supported](https://endoflife.date/openssl) so that means that it contains security vulnerabilities. Apart from that, you are using a .deb file for Debian 11 when building the Debian 12 image. If you want to include the package, you will need to build your own image using the Bitnami's one as base image and include the package. 

Thanks

rajdudhare1:
I would like to request you again to reconsider the same as we have to rebuild the images on our side if this isn't part of that.

Issue #74719, opened 2024-11-10
[bitnami/moodle] Moodle won't start "grep: /opt/bitnami/moodle/config.php: No such file or directory"


EDIflyer:
### Name and Version

bitnami/moodle4.5

### What architecture are you using?

amd64

### What steps will reproduce the bug?

1. Setup new Linode server (Debian 12)
2. Install Docker
3. Install nginx proxy manager (for reverse proxy)
4. Try to setup Moodle

### What is the expected behavior?

Containers start and run OK

### What do you see instead?

Initially I couldn't get mariadb to work but then read elsewhere that it needs PGID/PUID of 1001 so created a user with that and did a `chown` for the mariadb folders.  It now starts fine.

The main Moodle container boots but then has an issue - see debug logs below.  It just keeps restarting.  I also note that nothing is written to the mounted volumes (compared to the mariadb ones).  I've down the same `chown` for the moodle folders too.

### Additional information

Portainer stack/Docker compose:

``` yaml
# Copyright Broadcom, Inc. All Rights Reserved.
# SPDX-License-Identifier: APACHE-2.0

services:
  mariadb:
    image: docker.io/bitnami/mariadb:11.4
    expose:
      - 3306    
    environment:
      # ALLOW_EMPTY_PASSWORD is recommended only for development.
      - ALLOW_EMPTY_PASSWORD=yes
      - MARIADB_USER=bn_moodle
      - MARIADB_DATABASE=bitnami_moodle
      - MARIADB_CHARACTER_SET=utf8mb4
      - MARIADB_COLLATE=utf8mb4_unicode_ci
    restart: unless-stopped
    networks:
      - nginx-proxy-manager_default
    volumes:
      - /home/ediflyer/containers/moodle/mariadb:/bitnami/mariadb

  moodle:
    image: docker.io/bitnami/moodle:4.5
    expose:
      - 8080
    environment:
      - BITNAMI_DEBUG=true
      - MOODLE_DATABASE_TYPE=mariadb
      - MOODLE_DATABASE_HOST=mariadb
      - MOODLE_DATABASE_PORT_NUMBER=3306
      - MOODLE_DATABASE_USER=bn_moodle
      - MOODLE_DATABASE_NAME=bitnami_moodle
      # ALLOW_EMPTY_PASSWORD is recommended only for development.
      - ALLOW_EMPTY_PASSWORD=yes
      - MOODLE_HOST=sub.domain.tld
      - MOODLE_REVERSEPROXY=true
[...]

PR #74968, opened 2024-11-17
Update docker-compose.yml

EDIflyer:
Correct internal container paths from `/bitnami/moodle` and `/bitnami/moodledata` to just `/moodle` and `/moodledata`

See discussion at #74719

Other issues raised there:
1. The need to set a `MOODLE_SKIP_BOOTSTRAP=yes` environmental variable after first run (I think this needs to go into the guide at https://hub.docker.com/r/bitnami/moodle)
2. The error in the docker compose changes that mix up `moodle` and `moodledata` (also referenced in the Issue above)

### Description of the change
Correct internal container paths from `/bitnami/moodle` and `/bitnami/moodledata` to just `/moodle` and `/moodledata`

### Benefits
Container now runs OK

### Possible drawbacks
I'm unclear why this change was needed if others have used it OK before

### Applicable issues
Still need to set environmental variable to skip bootstrap after first-run (see above)

### Additional information
Covered in the GitHub issue above.

### DIFF
--- a/bitnami/moodle/docker-compose.yml
+++ b/bitnami/moodle/docker-compose.yml
@@ -26,8 +26,8 @@ services:
       # ALLOW_EMPTY_PASSWORD is recommended only for development.
       - ALLOW_EMPTY_PASSWORD=yes
     volumes:
-      - 'moodle_data:/bitnami/moodle'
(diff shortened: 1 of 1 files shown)


EDIflyer:
See further discussion in issue - no longer required.

PR #75114, opened 2024-11-22
[bitnami/postgresql] Update README.md

fidalgo:
### Description of the change

Fix the POSTGRESQL_PGBACKREST_CONF_FILE description as it defines the path to the pgbackrest.conf file that contains configuration directives for pgBackRest.


juan131 (approved):
Thanks for the fix in the docs! LGTM

PR #75245, opened 2024-11-26
[bitnami/openldap] Renaming env variables as pointed in #19716

### LINKED ISSUE #19716: LDAP groups should be in a different OU than users

barsikus007:
### Description of the change

Fix of https://github.com/bitnami/containers/issues/19716 issue.
I updated my previous PR #51192 (sorry, for long delay, I used other ldap container in my project).

### Benefits

Correct and predictable location of groups organization unit (E.g. [hub.docker.com/r/wheelybird/ldap-user-manager](https://hub.docker.com/r/wheelybird/ldap-user-manager) image expects groups to be in a different OU than users).

### Possible drawbacks

The modifications should not present any drawbacks since they are now compatible with the existing configurations. Additionally, I have included a deprecation notice as per @migruiz4's suggestion.

### Applicable issues

- fixes #19716.


dgomezleon (approved):

PR #75832, opened 2024-12-19
Update vib-action-test.yml

Jackblanket847:
### Applicable issues

- fixes #


carrodher:
Please note the runners for those actions are internally defined and we don't plan to change them

PR #75906, opened 2024-12-25
[bitnami/etcd] Stop relying on files for state

### LINKED ISSUE #16069: [bitnami/etcd] etcd pods are unable to join existing cluster on node drain
### Name and Version

bitnami/etcd-3.5.8

None

I'm using 3 node Kubernetes cluster and 3 instances of etcd.

When I'm deleting a pod, pod is able to restart. 
When I'm only draining a node, the pod is not able to re-join the cluster, and unable to start.

Observations:
`ETCD_INITIAL_CLUSTER_STATE` is 'new' when it's starting from zero (first time).
CASE-1: When deleting a pod `ETCD_INITIAL_CLUSTER_STATE` is changing from 'new' to 'existing', and pod is able to start.
CASE-2: When draining a node `ETCD_INITIAL_CLUSTER_STATE` is staying 'new', and newly created pod is unable to join the cluster and unable to restart.

I tried with and without persistence.

pod should start on node drain. And as per my understanding, the 'ETCD_INITIAL_CLUSTER_STATE' should change to 'existing' on node drain as well.

### What do you see instead?

etcd pod not starting on node drain.

### Additional information

Please let me know, if this behavior is expected or not, and how can I prevent pod restart failure on node drain.

I'm not sure if it will help: 
 1) I have tried this in different clusters (rhel & ubuntu based)
 2) Other services are working fine(kafka, zookeeper), network is fine.

pckhoi:
### Description of the change

The current etcd container and chart have a few major problems:

1. It relies on files outside of the data directory which could contain conflicting information compared to the data dir and the cluster
2. Removing the member during pre-stop hook is problematic. I guess this was added to support scaling down the cluster. If so, this logic is leaky. There are 2 cases where this breaks down:
	1. If the pod was killed for reasons other than `replicas` update then the next time the pod starts, it will not be able to start from the existing data dir which means it must throw away the data dir and start from scratch.
	2. If the cluster is scaled down and the PVC is retained, the next time the cluster is scaled up, the new member will encounter a non-empty data dir which it must discard
3. If the member was removed, the container chokes up on non-empty data dir in most cases except when recovering from a snapshot
4. It might attempt to add a new member even when an old member with the same name already exists. This is caused by relying on files for state
5. It runs `etcdctl member update` for unclear reasons when the data dir is not empty and there is a member ID
6. It relies on `ETCD_INITIAL_CLUSTER_STATE` to know whether the cluster is new which could be inaccurate

This PR add the following changes:

- Add `preupgrade.sh` which should be run in a Helm pre-upgrade hook. When the cluster is scaled down, it detects and removes obsolete members with `etcdctl member remove`.
- Remove `prestop.sh`
- Stop storing/checking member ID from the `member_id` file. Instead, the remote member ID is read from the cluster with `etcdctl member list`, and the local member ID is checked for conflict during startup.
- Stop storing/checking member removal state from `member_removal.log`. Check with `etcdctl member list` instead.
- If the data dir is not empty, check if the member still belongs to the cluster (remote ID and local ID are the same). If there is a conflict, remove the data dir, remove the old member, add a new member, and start the member from scratch.
- Remove environment variable `ETCD_DISABLE_STORE_MEMBER_ID`
- Remove environment variable `ETCD_DISABLE_PRESTOP`
- Environment variable `ETCD_INITIAL_CLUSTER_STATE` becomes read-only

### Benefits

- Not relying on files outside of the data directory means there is only a single source of truths (or only as many as there are live members in the cluster plus the data dir), which makes most operations more reliable
- Removing obsolete members in Helm pre-upgrade hook means the `etcdctl member remove` command tends to be executed against a healthy cluster
- If the pod was killed for reasons other than replica changes, it can rejoin the cluster on its own while keeping all its data intact
- The container no longer chokes up on a non-empty data dir, even when the old member is removed

### Possible drawbacks

- If during initialization there is a network outage and the current member can't connect to other members, it will think that it must start a new cluster. That said, I don't think there is any good solution in this case except manual recovery.
- I have not tested this set of changes outside of Helm/K8s

### Applicable issues

- fixes [#16069](https://github.com/bitnami/charts/issues/16069)

### Additional information

Related changes in the Helm chart: https://github.com/bitnami/charts/pull/31161 and https://github.com/bitnami/charts/pull/31164


pckhoi:
I'm planning to open a complementary PR in the charts repo. I will try to add more tests there.

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/preupgrade.sh
```diff
@@ -0,0 +1,89 @@
...
+# Globals:
+#   ETCD_*
+# Arguments:
+#   None
+# Returns:
+#   None
+#########################
+remove_members() {
+    local -r current=$(mktemp)
+    local -r expected=$(mktemp)
```
juan131, line 66: Is there any reason to save on "current" and "expected" results on temporary files instead of using simple variables and save it in memory?

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/preupgrade.sh
```diff
@@ -0,0 +1,89 @@
...
+    port="$(parse_uri "${advertised_array[0]}" "port")"
+
+    for i in $(seq 0 $(($count - 1))); do
+        pod_name="${MY_STS_NAME}-${i}"
+        endpoints+=("${pod_name}.${ETCD_CLUSTER_DOMAIN}:${port:-2380}")
+    done
+
+    debug "etcdctl endpoints are ${endpoints[*]}"
+    echo "${endpoints[*]}" | tr ' ' ','
+}
```
juan131, line 53: Isn't this function very similar to `get_initial_cluster` available at `libetcd.sh`?

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/libetcd.sh
```diff
@@ -572,6 +520,94 @@ recalculate_initial_cluster() {
...
+        info "The remote member ID is different from the local member ID"
+        ret=1
+    elif grep -q "\"error\":\"cluster ID mismatch\"" "$tmp_file"; then
+        info "The remote cluster ID is different from the local cluster ID"
+        ret=1
+    else
+        info "The member is still part of the cluster"
+    fi
+    rm -f "$tmp_file"
+    return $ret
```
juan131, line 608: You can simplify this:

```suggestion
    trap rm -f "$tmp_file"
    if grep -q "the member has been permanently removed from the cluster\|ignored streaming request; ID mismatch" "$tmp_file"; then
        info "The remote member ID is different from the local member ID"
        return 1
    elif grep -q "\"error\":\"cluster ID mismatch\"" "$tmp_file"; then
        info "The remote cluster ID is different from the local cluster ID"
        return 1
    fi
    
    info "The member is still part of the cluster"
    return 0
```

Also, wouldn't it be better to rely on some etcdctl command to check if the node is still a member instead of starting etcd in background?

bitnami/etcd/README.md
```diff
@@ -219,19 +216,20 @@ Apart from providing your custom configuration file, you can also modify the ser
...
+| `ETCD_DEFAULT_CONF_DIR`      | etcd default configuration directory.                                | `${ETCD_BASE_DIR}/conf.default`    |
+| `ETCD_TMP_DIR`               | Directory where ETCD temporary files are stored.                     | `${ETCD_BASE_DIR}/tmp`             |
+| `ETCD_CONF_FILE`             | ETCD configuration file.                                             | `${ETCD_CONF_DIR}/etcd.yaml`       |
+| `ETCD_NEW_MEMBERS_ENV_FILE`  | File containining the etcd environment to use after adding a member. | `${ETCD_DATA_DIR}/new_member_envs` |
+| `ETCD_DAEMON_USER`           | etcd system user name.                                               | `etcd`                             |
+| `ETCD_DAEMON_GROUP`          | etcd system user group.                                              | `etcd`                             |
+| `ETCD_INITIAL_CLUSTER_STATE` | Initial cluster state. Either "new" or "existing".                   | `nil`                              |
 
 Additionally, you can configure etcd using the upstream env variables [here](https://etcd.io/docs/v3.4/op-guide/configuration/)
```
juan131, line 234: "Notable Changes" section should be updated describing the introduced changes.

juan131 (changes requested):
Hi @pckhoi 

Thanks so much for this amazing contribution! It'd definitely help on making the Bitnami etcd chart more stable.

I think the main concern/challenge with your changes would be providing a solution for users who may scale down the cluster via `kubectl scale sts/etcd --replicas X` (or via some HorizontalPodAutoscaler that may also scale down the cluster without Helm's control via hooks). Correct me if I'm wrong but this use case won't be covered, right?

pckhoi:
@juan131 you're correct that the autoscaling use case isn't covered. People use Etcd for its consistency rather than for handling large, fluctuating traffic so I think autoscaling to handle large traffic is a niche use case.

As for manual scaling, running `helm upgrade` makes more sense if the cluster is installed via Helm. If people are scaling with `kubectl scale` then they're probably not using Helm which makes it difficult to operate. So no, deploying/upgrading without Helm is also not supported.

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/preupgrade.sh
pckhoi: No, it's not the same because this function doesn't use env var `ETCD_INITIAL_ADVERTISE_PEER_URLS` which a pre-upgrade hook shouldn't have defined.

pckhoi: Perhaps I can modify `get_initial_cluster` to stop using `ETCD_INITIAL_ADVERTISE_PEER_URLS` and then we can reuse it here.

juan131:
Thanks for confirming so @pckhoi ! In that case, I'd add a warning at the "Upgrading" section alerting about what these changes imply (I mean, warning users to use exclusively Helm to scale the cluster):

- https://github.com/bitnami/charts/tree/main/bitnami/etcd#upgrading

We could even add it in the chart NOTES:

- https://github.com/bitnami/charts/blob/main/bitnami/etcd/templates/NOTES.txt

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/libetcd.sh
pckhoi: If there is an etcdctl command that checks the data dir directly then I'm not aware of it. Most commands only work on running etcd endpoints.

pckhoi:
Sure, I will do that.

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/preupgrade.sh
pckhoi: Sure, let me change that.

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/preupgrade.sh
pckhoi: So I ended up removing `get_initial_cluster` because `ETCD_INITIAL_CLUSTER` always has URL scheme defined for each member since we only support installing via Helm now.

I also replaced `etcdctl_job_endpoints` function with `endpoints_as_host_port` function which is far simpler.

bitnami/etcd/README.md
pckhoi: Updated

pckhoi:
@juan131 I have updated https://github.com/bitnami/charts/tree/main/bitnami/etcd#upgrading. As for https://github.com/bitnami/charts/blob/main/bitnami/etcd/templates/NOTES.txt, I don't see anything that needs to be updated.

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/libetcd.sh
pckhoi: I have refactored as you suggested. Please check again.

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/preupgrade.sh
pckhoi: Refactored as suggested.

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/libetcd.sh
```diff
@@ -411,7 +340,14 @@ was_etcd_member_removed() {
...
 is_new_etcd_cluster() {
-    [[ "$ETCD_INITIAL_CLUSTER_STATE" = "new" ]] && [[ "$ETCD_INITIAL_CLUSTER" = *"$ETCD_INITIAL_ADVERTISE_PEER_URLS"* ]]
+    local -a extra_flags
+    read -r -a extra_flags <<<"$(etcdctl_auth_flags)"
+    is_boolean_yes "$ETCD_ON_K8S" && extra_flags+=("--endpoints=$(etcdctl_get_endpoints)")
+    if ! etcdctl endpoint status --cluster ${extra_flags[@]}; then
+        return 0
+    else
+        return 1
+    fi
```
juan131, line 350: Shorter:
```suggestion
    ! debug_execute etcdctl endpoint status --cluster ${extra_flags[@]}
```

bitnami/etcd/README.md
```diff
@@ -219,24 +216,33 @@ Apart from providing your custom configuration file, you can also modify the ser
...
 
 ## Notable Changes
 
+### 3.5.17-debian-12-r3
+
+* Drop support for non-Helm deployment. Upgrading of any kind including increasing replica count must also be done with `helm upgrade` exclusively. CD automation tools that respect Helm hooks such as ArgoCD can also be used.
```
juan131, line 240:
```suggestion
* Drop support for non-Helm cluster deployment. Upgrading of any kind including increasing replica count must also be done with `helm upgrade` exclusively. CD automation tools that respect Helm hooks such as ArgoCD can also be used.
```

bitnami/etcd/README.md
```diff
@@ -219,24 +216,33 @@ Apart from providing your custom configuration file, you can also modify the ser
...
+| `ETCD_CONF_DIR`              | etcd configuration directory.                                        | `${ETCD_BASE_DIR}/conf`            |
+| `ETCD_DEFAULT_CONF_DIR`      | etcd default configuration directory.                                | `${ETCD_BASE_DIR}/conf.default`    |
+| `ETCD_TMP_DIR`               | Directory where ETCD temporary files are stored.                     | `${ETCD_BASE_DIR}/tmp`             |
+| `ETCD_CONF_FILE`             | ETCD configuration file.                                             | `${ETCD_CONF_DIR}/etcd.yaml`       |
+| `ETCD_NEW_MEMBERS_ENV_FILE`  | File containining the etcd environment to use after adding a member. | `${ETCD_DATA_DIR}/new_member_envs` |
+| `ETCD_DAEMON_USER`           | etcd system user name.                                               | `etcd`                             |
+| `ETCD_DAEMON_GROUP`          | etcd system user group.                                              | `etcd`                             |
+| `ETCD_INITIAL_CLUSTER_STATE` | Initial cluster state. Either "new" or "existing".                   | `nil`                              |
```
juan131, line 232: Is this still used?

juan131 (changes requested):

bitnami/etcd/README.md
pckhoi: The user does not use it. Is it fine to leave it in the read-only section then?

pckhoi:
Thanks! I have addressed all the suggestsions.

juan131:
@pckhoi I think this PR looks great now! Could you please check my comments in the associated chart PR? Thanks in advance.

bitnami/etcd/README.md
pckhoi: Removed

juan131 (changes requested):

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/preupgrade.sh
```diff
@@ -0,0 +1,51 @@
...
+read -r -a extra_flags <<<"$(etcdctl_auth_flags)"
+is_boolean_yes "$ETCD_ON_K8S" && extra_flags+=("--endpoints=$(endpoints_as_host_port)")
+debug "Listing members"
+if ! current="$(etcdctl member list ${extra_flags[@]} --write-out simple | awk -F ", " '{print $3 ":" $1}')"; then
+    debug "Error listing members, is this a new cluster?"
+    exit 0
+fi
+info "Current cluster members are: $(echo "$current" | awk -F: '{print $1}' | tr -s '\n' ',' | sed 's/,$//g')"
+expected="$(echo $ETCD_INITIAL_CLUSTER | tr -s ',' '\n' | awk -F= '{print $1}')"
+info "Expected cluster members are: $(echo "$expected" | tr -s '\n' ',' | sed 's/,$//g')"
```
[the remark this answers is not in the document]
pckhoi: I have made those changes.

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/preupgrade.sh
```diff
@@ -0,0 +1,51 @@
...
+# from the cluster before running Helm upgrades that potentially scale
+# down the etcd cluster
+
+read -r -a extra_flags <<<"$(etcdctl_auth_flags)"
+is_boolean_yes "$ETCD_ON_K8S" && extra_flags+=("--endpoints=$(endpoints_as_host_port)")
+debug "Listing members"
+if ! current="$(etcdctl member list ${extra_flags[@]} --write-out simple | awk -F ", " '{print $3 ":" $1}')"; then
+    debug "Error listing members, is this a new cluster?"
+    exit 0
+fi
```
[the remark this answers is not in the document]
pckhoi: So I changed the exit code to 1 instead because proceeding in this case is unsafe. This way, I can delegate retrying to k8s/helm - the job will re-run until it succeeds. Plus, the pre-upgrade hook is never run during installation so we shouldn't have a new cluster situation ever.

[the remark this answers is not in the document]
juan131: thanks!

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/libetcd.sh
```diff
@@ -572,6 +474,94 @@ recalculate_initial_cluster() {
...
+    
+    while read -r line; do
+        echo "$line" # Stream the output
+        if [[ "$line" =~ (established TCP streaming connection with remote peer|the member has been permanently removed from the cluster|ignored streaming request; ID mismatch|\"error\":\"cluster ID mismatch\") ]]; then
+            kill "$pid"
+            wait "$pid" 2>/dev/null
+            debug "Stopped etcd"
+            break
+        fi
+    done < <(tail -f "$tmp_file")
```
juan131, line 551: I've been doing more tests today and we'll have to change this block. This always throws etcd start logs to stdout regardless `BITNAMI_DEBUG` is set or not. Alternative:

```suggestion
    am_i_root && start_command=("run_as_user" "$ETCD_DAEMON_USER" "${start_command[@]}")
    [[ -f "$ETCD_CONF_FILE" ]] && start_command+=("--config-file" "$ETCD_CONF_FILE")
    "${start_command[@]}" > "$tmp_file" 2>&1 &
    
    while read -r line; do
        debug_execute echo "$line"
        if [[ "$line" =~ (established TCP streaming connection with remote peer|the member has been permanently removed from the cluster|ignored streaming request; ID mismatch|\"error\":\"cluster ID mismatch\") ]]; then
            etcd_stop
            debug "Stopped etcd"
            break
        fi
    done < <(tail -f "$tmp_file")
```

By the way, it's not necessary to save the PID, you can use `etcd_stop`

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/preupgrade.sh
```diff
@@ -0,0 +1,58 @@
...
+debug "Listing members"
+if ! current="$(etcdctl member list ${extra_flags[@]} --write-out simple | awk -F ", " '{print $3 ":" $1}')"; then
+    error "Unable to list members, are all members healthy?"
+    exit 1
+fi
+info "Current cluster members are: $(echo "$current" | awk -F: '{print $1}' | tr -s '\n' ',' | sed 's/,$//g')"
+
+expected="$(echo $ETCD_INITIAL_CLUSTER | tr -s ',' '\n' | awk -F= '{print $1}')"
+info "Expected cluster members are: $(echo "$expected" | tr -s '\n' ',' | sed 's/,$//g')"
+read -r -a obsolete_members <<<"$(comm -23 <(echo "$current" | awk -F: '{print $1}' | sort) <(echo "$expected" | sort))"
```
juan131, line 49: small fix:
```suggestion
read -r -a obsolete_members <<<"$(comm -23 <(echo "$current" | awk -F: '{print $1}' | sort) <(echo "$expected" | sort) | tr -s '\n' ' ')"
```

pckhoi: This change seems to break the script for me.

pckhoi: Actually, something else is breaking. Let me investigate more.

juan131: What problems are you experiencing? I works for me

Without this change only one member is removed given the `obsolete_members` wasn't an array but a string with the 1st obsolete member to remove, so when I did a test scaling down from 5 to 3 replicas it only removed one.

pckhoi: Sorry, your suggestion works. I just had to fix the loop as well. Everything should work now.

bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/libetcd.sh
pckhoi: Updated.

juan131 (approved):
LGTM

saurabhnetskope:
@juan131 @pckhoi Please check this related issue https://github.com/bitnami/containers/issues/78398

PR #76236, opened 2025-01-14
[bitnami/clickhouse] add support for etc/config.d

pszynk:
### Description of the change

`etc/conf.d` is obselete, `etc/config.d` should be used. 

### Benefits

Consistend with official Clickhouse documentation

### Possible drawbacks

Redundant `conf.d` directory


jotamartos (approved):
Thank you for your taking the time to create this PR. Would you like to [contribute](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md) and update the Bitnami ClickHouse Chart so it starts using the config.d folder?

PR #76609, opened 2025-01-24
[bitnami/nats] Recover Goss test for NATS

juan131:
### Description of the change

We removed Goss tests for NATS at #76501 given we've transformed NATS into a scratch image and the limitations available on VIB to test this kind of images. This limitation is no longer an issue so we should enable them back. 

### Benefits

Better test suite

### Possible drawbacks

None

### Applicable issues

None

### Additional information

N/A


gongomgra (dismissed):
LGTM

.vib/nats/goss/goss.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright Broadcom, Inc. All Rights Reserved.
+# SPDX-License-Identifier: APACHE-2.0
+
+gossfile:
+  # Goss tests exclusive to the current container
+  ../../nats/goss/nats.yaml: {}
+  # Load scripts from .vib/common/goss/templates
+  ../../common/goss/templates/check-app-version-stdout.yaml: {}
```
gongomgra, line 8: I think you are missing the `-no-shell-` substring here.

`.vib/common/goss/templates/check-app-version-no-shell-stdout.yaml`

gongomgra (approved):

PR #76762, opened 2025-01-29
[bitnami/php-fpm-min] Add Goss test

juan131:
### Description of the change

This PR adds basic Goss tests for `php-fpm-min` that ensures expected files and directories exist, the version is correct, the SPDX file is available and expected port and process are in used.


dgomezleon (approved):
LGTM

PR #76831, opened 2025-02-01
[bitnami/zookeeper] fix for stale openssl commands

Drumar:
This fix will avoid performing openssl commands that will become a zombie if openssl is configured but no auth (mutual ssl) is set in the health check.

Related to https://github.com/bitnami/charts/issues/29603

### DIFF
--- a/bitnami/zookeeper/3.9/debian-12/rootfs/opt/bitnami/scripts/libzookeeper.sh
+++ b/bitnami/zookeeper/3.9/debian-12/rootfs/opt/bitnami/scripts/libzookeeper.sh
@@ -597,7 +597,7 @@ zookeeper_healthcheck() {
     local args=()
     local port="$ZOO_PORT_NUMBER"
 
-    if [[ "$ZOO_TLS_CLIENT_ENABLE" = true ]]; then
+    if [[ "$ZOO_TLS_CLIENT_ENABLE" = true ]] && is_boolean_yes "$ZOOKEEPER_ENABLE_AUTH"; then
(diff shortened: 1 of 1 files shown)


Drumar:
Trying to remove the stale condition

bitnami/zookeeper/3.9/debian-12/rootfs/opt/bitnami/scripts/libzookeeper.sh
```diff
@@ -597,7 +597,7 @@ zookeeper_healthcheck() {
     local args=()
     local port="$ZOO_PORT_NUMBER"
 
-    if [[ "$ZOO_TLS_CLIENT_ENABLE" = true ]]; then
+    if [[ "$ZOO_TLS_CLIENT_ENABLE" = true ]] && is_boolean_yes "$ZOOKEEPER_ENABLE_AUTH"; then
```
migruiz4, line 600: The environment variable `ZOOKEEPER_ENABLE_AUTH` does not exist.

I think the variable you are looking for is `ZOO_TLS_CLIENT_AUTH`, which can be set to `none`, `want`, `need`.

So your conditional should be like this: 

```suggestion
    if [[ "$ZOO_TLS_CLIENT_ENABLE" = true && "$ZOOKEEPER_ENABLE_AUTH" != "none" ]]; then
```

migruiz4 (commented):
Hi @Drumar,

I'm sorry for the late response. Could you please take a look at my comment?

PR #77005, opened 2025-02-05
[bitnami/mariadb] fix healthcheck during initialization using skip-networking

### LINKED ISSUE #76241: No way to detect full initialization of the database, healthcheck.sh always returns code 0
### Name and Version

bitnami/mariadb:10.11

None

1. Run mariadb with a lot of sql files which slow down initialization with a compose file like this : 
```yaml
  db:
    image: bitnami/mariadb:10.11
    environment:
      BITNAMI_DEBUG: true
      ALLOW_EMPTY_PASSWORD: "true"
      MARIADB_SKIP_TEST_DB: "yes"
      MARIADB_USER: root
      MARIADB_PASSWORD: password
      MARIADB_DATABASE: bitnami
      MARIADB_ROOT_PASSWORD: password
    ports:
      - 3306:3306
    healthcheck:
      test: [ "CMD", "/opt/bitnami/scripts/mariadb/healthcheck.sh", "--connect", "--innodb_initialized" ]
      start_period: 10s
      interval: 10s
      timeout: 5s
      retries: 10
    volumes:
      - ./sql/:/docker-entrypoint-initdb.d/
```

2. we expect the container to be ready only when initialization is completed. In the official mariadb image, initialization of the db is executed with the server running with `--skip-networking` flag. In this case, `healthcheck.sh --connect` fails and the container is not in a ready state.
With the bitnami image, the `--skip-networking` flag is not used, which prevents to detect the full initialization of the database

`healthcheck.sh --connect` should allow to detect the full initialization of the database

### What do you see instead?

There is no way to detect the real ready state of the database

### Additional information

If this bug is valid and the proposal to use `--skip-networking` during setup is accepted, I am willing to contribute to this, please tell me.

olivierboudet:
### Description of the change

With this change, mariadb starts with `--skip-networking` option during initialization. 

### Benefits

Without networing, `healtcheck.sh` script can not connect to mysql and fails, which is necessary when having a lot of init scripts (which can take one minute to run)

### Applicable issues

- fixes  #76241

### DIFF
--- a/bitnami/mariadb/10.11/debian-12/rootfs/opt/bitnami/scripts/libmariadb.sh
+++ b/bitnami/mariadb/10.11/debian-12/rootfs/opt/bitnami/scripts/libmariadb.sh
@@ -445,7 +445,7 @@ mysql_start_bg() {
     local -a flags=("--defaults-file=${DB_CONF_FILE}" "--basedir=${DB_BASE_DIR}" "--datadir=${DB_DATA_DIR}" "--socket=${DB_SOCKET_FILE}")
 
     # Only allow local connections until MySQL is fully initialized, to avoid apps trying to connect to MySQL before it is fully initialized
-    flags+=("--bind-address=127.0.0.1")
+    flags+=("--skip-networking")
 
     # Add flags specified via the 'DB_EXTRA_FLAGS' environment variable
     read -r -a db_extra_flags <<< "$(mysql_extra_flags)"


bitnami/mariadb/10.11/debian-12/rootfs/opt/bitnami/scripts/libmariadb.sh
```diff
@@ -445,7 +445,7 @@ mysql_start_bg() {
     local -a flags=("--defaults-file=${DB_CONF_FILE}" "--basedir=${DB_BASE_DIR}" "--datadir=${DB_DATA_DIR}" "--socket=${DB_SOCKET_FILE}")
 
     # Only allow local connections until MySQL is fully initialized, to avoid apps trying to connect to MySQL before it is fully initialized
-    flags+=("--bind-address=127.0.0.1")
+    flags+=("--skip-networking")
```
juan131, line 448: Please correct me if I'm wrong but according to [MariaDB docs](https://mariadb.com/docs/server/ref/mdb/system-variables/skip_networking) this option will disable any connection via TCP/IP, hence when we start MariaDB in background we won't be able to run commands for operations such as configuring authentication. For instance, operations like the ones below:

- https://github.com/bitnami/containers/blob/main/bitnami/mariadb/10.11/debian-12/rootfs/opt/bitnami/scripts/libmariadb.sh#L344-L351

Am I right?

juan131 (commented):
Thanks for your contribution! Please check my comments

bitnami/mariadb/10.11/debian-12/rootfs/opt/bitnami/scripts/libmariadb.sh
olivierboudet: I am not an expert at all, but this option disables tcp/ip networking, I suppose the connection through sockets is working.

Just for reference, official mariadb is using this option during initialization : https://github.com/MariaDB/mariadb-docker/blob/master/docker-entrypoint.sh#L121

And when testing the container with my change, I can connect after initialization with the user created at startup, and my init scripts are executed as well.

juan131 (approved):
LGTM

juan131:
@olivierboudet your changes are available on `10.11.11-debian-12-r1` and `11.4.5-debian-12-r1` releases

olivierboudet:
> @olivierboudet your changes are available on `10.11.11-debian-12-r1` and `11.4.5-debian-12-r1` releases

Thanks !

PR #77036, opened 2025-02-06
[bitnami/matomo] Exclude matomo.js from being rsynced (issue #76460)

### LINKED ISSUE #76460: [bitnami/matomo] Exclude matomo.js from being rsynced, same as config.ini.php
### Name and Version

bitnami/matomo:5.1.0

Some officiel plugins like [Media Anaytics](https://developer.matomo.org/guides/media-analytics/setup) modify the tracker file matomo.js 

With the actual bitnami image, all files except `config.ini.php` are made persistent due to [33b63ac](https://github.com/bitnami/containers/commit/33b63ac86846d63964fce5555fedd0d2d9b0e76c)

```
 4.15.0-debian-11-r20
From this version on, all Matomo files are persisted (MATOMO_DATA_TO_PERSIST env var). During the upgrade process, they will be replaced (except the config.ini.php file) as suggested in [the official documentation](https://matomo.org/faq/on-premise/update-matomo/#the-manual-three-step-update)
```

Exclude matomo.js from the rsync command, same as config.ini.php

in `/rootfs/opt/bitnami/scripts/matomo-env.sh`

`export MATOMO_TRACKER_FILE="${MATOMO_BASE_DIR}/matomo.js"`

in `/rootfs/opt/bitnami/scripts/libmatomo.sh`

`rsync -a "$MATOMO_BASE_DIR"/* "$MATOMO_VOLUME_DIR" --exclude "$(realpath --relative-to="$MATOMO_BASE_DIR" "$MATOMO_CONF_FILE")" --exclude "$(realpath --relative-to="$MATOMO_BASE_DIR" "$MATOMO_TRACKER_FILE")"`

As possible alternative:
- specify the plugin tracker `plugins/MediaAnalytics/tracker.min.js`  in the the header as specified in the [Media Anaytics](https://developer.matomo.org/guides/media-analytics/setup) documentation but it implies to modify the code of each website
- use a custom post init script to restore the matomo.js file from a previous save

benemart:
Signed-off-by: Bénédicte Martinez <benedicte.martinez@cea.fr>

### Description of the change

This change prevents the javascript file tracker `matomo.js` from being replaced by the original Matomo code (rsync command when the container starts up).
The `matomo.js` file can be modified by plugins same as the configuration file `config.ini.php`.

### Benefits

It permits to plugins to add custom code directly into the matomo.js file, without then any further modifiation for the web sites.

### Possible drawbacks

This modification implies possible custom action during Matomo upgrade as the original `matomo.js` will not be copied.
For example in Kuberntes, using a job or init container to inject the new matomo.js before the upgrade.

### Applicable issues

- fixes #76460


dgomezleon:
Hi @benemart ,

Thanks for your contribution.

As you can see [here](https://matomo.org/faq/on-premise/update-matomo/#the-manual-three-step-update), the only file that is included in the documentation to be persisted during an upgrade is `config.ini.php`. Since the scenario you describe makes sense, have you considered using something like this?

```
rsync_args=("-a" "$MATOMO_BASE_DIR"/* "$MATOMO_VOLUME_DIR" "--exclude" "$(realpath --relative-to="$MATOMO_BASE_DIR" "$MATOMO_CONF_FILE")")
if [[ -n "$MATOMO_EXTRA_DATA_TO_PERSIST" ]]; then
    for file in $MATOMO_EXTRA_DATA_TO_PERSIST; do
        rsync_args+=("--exclude")
        rsync_args+=("$file")
    done
fi
rsync "${rsync_args[@]}"
```

That way, the default behavior follows the upstream documentation but we can also support your scenario.

This would imply adding MATOMO_EXTRA_DATA_TO_PERSIST env var.

Let me know your thoughts.

benemart:
Hello @dgomezleon 
You are right, your code is smarter than mine.
I will test this and make a new PR if successful.
Thanks for your help

benemart:
I think the variable name is not appropriate : i propose MATOMO_EXCLUDE_DATA_TO_PERSIST as we want it to be excluded from the rsync.
Are you ok with this ?

bitnami/matomo/5/debian-12/rootfs/opt/bitnami/scripts/libmatomo.sh
```diff
@@ -218,7 +218,14 @@ EOF
     else
         info "Persisted Matomo installation detected"
         info "Updating Matomo files in persisted data"
-        rsync -a "$MATOMO_BASE_DIR"/* "$MATOMO_VOLUME_DIR" --exclude "$(realpath --relative-to="$MATOMO_BASE_DIR" "$MATOMO_CONF_FILE")"
+        rsync_args=("-a" "$MATOMO_BASE_DIR"/* "$MATOMO_VOLUME_DIR" "--exclude" "$(realpath --relative-to="$MATOMO_BASE_DIR" "$MATOMO_CONF_FILE")")
+        if [[ -n "$MATOMO_EXCLUDE_DATA_FROM_PERSIST" ]]; then
+            for file in $MATOMO_EXCLUDE_DATA_FROM_PERSIST; do
+                rsync_args+=("--exclude")
+                rsync_args+=("$file")
```
dgomezleon, line 225:
```suggestion
                rsync_args+=("$(realpath --relative-to="$MATOMO_BASE_DIR" "$file")")
```

dgomezleon:
> I think the variable name is not appropriate : i propose MATOMO_EXCLUDE_DATA_FROM_PERSIST as we want it to be excluded from the rsync. Are you ok with this ?

Hi @benemart ,

The main goal of this PR is to optionally persist `matomo.js` (or any other specified file), similar to how `config.ini.php` is handled. I understand that excluding the file from the rsync process—which overwrites MATOMO_VOLUME_DIR—might seem a bit confusing. In that function, it could be called MATOMO_EXCLUDED_DATA_FROM_UPDATE. However, the new environment variable should clearly reflect its purpose: ensuring the specified file remains unaltered.

What do you think?

dgomezleon (approved):
LGTM. Thanks

PR #77131, opened 2025-02-10
[bitnami/grafana] don't error out on read-only config dir

enkore:
### Description of the change

Errors when copying the example configuration files to the Grafana config dir are now ignored.

### Benefits

The Grafana config dir as a whole or subdirectories may become read-only to the container, e.g. when mounting configmaps or secrets. Currently, these configurations require overwriting the entrypoint, which is undesirable.

### Possible drawbacks


gongomgra:
Hi @enkore,

Thanks for your contribution. I'm afraid I'm not fully understanding your changes. Can you provide us with an scenario in which this is required?

PR #77502, opened 2025-02-14
[bitnami/postgresql] Issue 77295 Enable native recovery with flag/env var

### LINKED ISSUE #77295: [bitnami/postgresql] Feature Request: Allow recovery by feature flag
### Name and Version

bitnami/postgresql-17.2.0-debian-12

In the current Dockerfile for the Bitnami/PostgreSQL container, there is no way to let the database server do a restore using the `recovery.signal` file, because it gets removed/cleaned up during the startup of the container. In case of a restored database, that file would allow the native recovery method to be used.

This happens in the function [postgresql_clean_from_restart](https://github.com/bitnami/containers/blob/main/bitnami/postgresql/17/debian-12/rootfs/opt/bitnami/scripts/libpostgresql.sh) and cannot be stopped. 

In our case using `wal-g`, the base backup is pulled from S3 storage in an InitContainer and the required WAL files are pulled using the native `recovery_command` (also from S3). We used the official Bitnami PostgreSQL helm chart with an `extendedConfiguration` containing the archive and recovery configuration. The InitContainer restores a backup to the persistent volume and creates the `recovery.signal` file, but that gets cleaned up during the startup of the container.

Here are the snippets from our proof-of-concept:

```yaml
# primary.extendedConfiguration
  extendedConfiguration: |
    archive_mode = on
    archive_command = 'wal-g wal-push %p'
    archive_timeout = 60
    restore_command = 'wal-g wal-fetch %f %p'
```

```shell
# Restore script
...

perform_restore() {
    log "Starting restore to ${DATA_DIR}..."
    wal-g backup-fetch "$DATA_DIR" ${BACKUP_NAME:-LATEST} || {
        log "Restore failed"
        return 1
    }
    touch "${DATA_DIR}/recovery.signal"
    chown -R 1001:1001 "$DATA_DIR"
    chmod 0660 $DATA_DIR/*
    chmod u=rwx,g=rwxs,o= $DATA_DIR/*/
    log "Restore completed"
    return 0
}
```

The below is just a quick proof-of-concept solution on how a flag/env var could be implemented in oder to enable a potential restore. Being controlled by an optional environment variable would not sacrifice stability, while enabling the feature if needed.

```shell
# Adjusted function
postgresql_clean_from_restart() {
    # local -r -a files=(
    # Removed read-only attribute
    local -a files=(
        "$POSTGRESQL_DATA_DIR"/postmaster.pid
        "$POSTGRESQL_DATA_DIR"/standby.signal
    )
   
    # "$POSTGRESQL_DATA_DIR"/recovery.signal
    # Replaced static entry with flag
    if [[ -z "${POSTGRESQL_PERFORM_RESTORE:-}" ]]; then
        files+=("$POSTGRESQL_DATA_DIR"/recovery.signal)
    fi

    for file in "${files[@]}"; do
        if [[ -f "$file" ]]; then
            info "Cleaning stale $file file"
            rm "$file"
        fi
    done
}
```

For the direct restore if the database on Kubernetes without using any of the operator-based solutions available, this seems to be the only viable option for a point-in-time recovery.

FuxMak:
### Description of the change

Adds the feature flag/environment variable `POSTGRESQL_PERFORM_RESTORE` to the initialization scripts of the container.

Adjusted files:
- "libpostgresql.sh" - Adjusted "postgresql_clean_from_restart" function to skip `recovery.signal` if flag is set
- "postgresql-env.sh" - Added new environment variable/flag incl. default value
- "README.md" - Added reference to the new environment variable/flag

### Benefits

When this is set, it removes the `recovery.signal` file from the cleanup process, which allows an administrator to use PostgreSQL native recovery mechanisms. Due to the default being 

### Possible drawbacks

Due to the default value and robust handling using the `is_boolean_yes` function, no side-effect shall occur if the environment variable is not set.

### Applicable issues

- fixes #77295


FuxMak:
Is there anything more I can do to speed up the process? I‘d like to avoid the PR simply being closed by a bot

migruiz4 (approved):
I'm sorry for the late response @FuxMak, changes look good to me, thank you for your contribution!

PR #77517, opened 2025-02-14
Update README.md with minimal images information

mpermar:
Updates README with minimal container images info.


mpermar:
I think this markdown check fails with the existing file on `main` branch too.

PR #78370, opened 2025-02-26
Add configurable timeout to netcat calls with 10s default 

mati865:
### Description of the change

Without any timeout, netcat can get stuck forever (until Kubernetes kills unhealthy Pod) when configured address and port combination is wrong, or the connection gets stuck.
For example `nc -z bitnami.com 12345` doesn't seem to ever finish.

### Benefits

In the case when an attempt gets stuck, the timeout will allow further attempts to be performed, instead of waiting until Pod is killed.

### Possible drawbacks

When the timeout is too short, valid configurations might falsely show as broken. Thus, I chose 10s which should be long enough.

### Additional information

Default 10s timeout is very conservative and should not cause spurious failures, users are advised to shorten it according to their needs.


gongomgra:
@mati865 thanks for your contribution! What do you think about adding a new parameter to the `<asset>-env.sh` file for this new setting?

gongomgra (approved):
Thanks for your contribution!

PR #78625, opened 2025-03-06
[bitnami/wordpress-nginx] deny access to php files in uploads/ files/ by default

LucaDev:
### Description of the change

This pull request adds an NGINX configuration block to explicitly deny access to any .php files within the /uploads and /files directories. The configuration is placed in a new file named 00_protect-upload-dirs.conf to ensure it is loaded before the php-fpm.conf file, as the inclusion order is determined alphabetically by filename.

Wordpress themselves suggest this here: https://developer.wordpress.org/advanced-administration/server/web-server/nginx/#global-restrictions-file

### Benefits

Enhanced Security - Prevents the execution of PHP files in user-uploaded directories, mitigating the risk of remote code execution attacks.

### Possible drawbacks

Very very edgy cases where users want php files to be user-uploadable maybe? 

### Applicable issues

None that I'm aware of

### Additional information

For future improvements, it may be beneficial to adopt a more structured naming convention for configuration files (e.g., using numerical prefixes like 01_, 02_, etc.) to explicitly control the inclusion order in /opt/bitnami/nginx/conf/bitnami/. This would decouple the file's purpose (described by its name) from its inclusion order, making the configuration more maintainable and less prone to errors.

Before:
Bildschirmfoto 2025-03-06 um 16 38 43

After:
Bildschirmfoto 2025-03-06 um 16 40 06


dgomezleon:
Hi @LucaDev ,

Thanks for contributing. Just a couple of minor comments:

* I would rename the file from `00_protect-upload-dirs.conf` to `00_protect-uploads-dirs.conf`
* Regarding using the number prefix, I agree with the idea of numbering it. However, I cannot see any point in including it before the other file `protect-hidden-files.conf`. In case of need, we should rename it too.

WDYT?

LucaDev:
Hi, thank you for your quick response.

> * I would rename the file from `00_protect-upload-dirs.conf` to `00_protect-uploads-dirs.conf`

I sure can do that. I'll change it as soon as I have your feedback regarding the other comment

> * Regarding using the number prefix, I agree with the idea of numbering it. However, I cannot see any point in including it Before the other file `protect-hidden-files.conf`. In case of need, we should rename it too.

The numbering prefix (00_) ensures the file is processed first, which is crucial for including it before the auto-generated **php-fpm.conf**. You're right - it's irrelevant for protect-hidden-files.conf since its include order doesn't impact functionality. But we have to include the "upload/*.php" handler before the "*.php" handler.
Renaming protect-hidden-files.conf with a number prefix isn't strictly necessary but we could also do that to keep it consistent and allow users to more easily hook into their custom location blocks/configurations into the nginx configuration. I'll wait on your feedback on this one to know how I should proceed

dgomezleon:
Hi @LucaDev ,

I see (thanks for the detailed explanation).  Let's rename both files to use `00_` prefix (`00_protect-hidden-files.conf` and `00_protect-uploads-dir.conf`).

LucaDev:
Hi @dgomezleon,

I have renamed the files. Anything else you would like me to change? 😄

dgomezleon (approved):

PR #78626, opened 2025-03-06
[bitnami/matomo] Fix SMTP authentication type allowed values (typo for Cram-md5)

aloene:
### Description of the change

Fix typo in SMTP authentication option value for "CRAM-MD5". Must be **Cram-md5**, was _Crammd5_.

### Benefits

Ability to use SMTP CRAM-MD5 authentication type by setting MATOMO_SMTP_AUTH environment variable value.

### Possible drawbacks

None.

### Applicable issues

None. But could be created. Setting _Crammd5_ authentication through MATOMO_SMTP_AUTH was possible but it cannot be used by Matomo as this is an unknown value (should be **Cram-md5**).


migruiz4 (approved):
Thank you very much for your contribution @aloene, looks good to me!

Issue #52988, opened 2023-11-17
[bitnami/discourse] Silent exit after DB population


zapbulon:
### Name and Version

bitnami/discourse:3 [latest]

### What architecture are you using?

None

### What steps will reproduce the bug?

1. On an ubuntu machine where I run also other dockerized applications I tried starting Discourse via `docker compose`
2. I use a redis and postgres instance running on another machine in a private network and Redis has no password because of that
3. This is the config: 
```yaml
  discourse:
    image: docker.io/bitnami/discourse:3
    ports:
      - '50110:3000'
    volumes:
      - '${HOME_DIR}/discourse:/bitnami/discourse'
    environment:
      - ALLOW_EMPTY_PASSWORD=yes
      - DISCOURSE_ENABLE_HTTPS=yes
      - DISCOURSE_USERNAME=...
      - DISCOURSE_PASSWORD=...
      - DISCOURSE_HOST=...
      - DISCOURSE_DATABASE_HOST=10.0.1.2
      - DISCOURSE_DATABASE_PORT_NUMBER=5432
      - DISCOURSE_DATABASE_USER=...
      - DISCOURSE_DATABASE_NAME=...
      - DISCOURSE_DATABASE_PASSWORD=...
      - DISCOURSE_REDIS_HOST=10.0.1.2
      - DISCOURSE_REDIS_PORT_NUMBER=6379
      - DISCOURSE_SMTP_HOST=...
      - DISCOURSE_SMTP_PORT=587
      - DISCOURSE_SMTP_USER=...
      - DISCOURSE_SMTP_PASSWORD=...

  sidekiq:
    image: docker.io/bitnami/discourse:3
    depends_on:
      - discourse
    volumes:
      - '${HOME_DIR}/sidekiq:/bitnami/discourse'
    command: /opt/bitnami/scripts/discourse-sidekiq/run.sh
    environment:
      - ALLOW_EMPTY_PASSWORD=yes
      - DISCOURSE_HOST=...
      - DISCOURSE_DATABASE_HOST=10.0.1.2
      - DISCOURSE_DATABASE_PORT_NUMBER=5432
      - DISCOURSE_DATABASE_USER=...
      - DISCOURSE_DATABASE_NAME=...
      - DISCOURSE_DATABASE_PASSWORD=...
      - DISCOURSE_REDIS_HOST=10.0.1.2
      - DISCOURSE_REDIS_PORT_NUMBER=6379
      - DISCOURSE_SMTP_HOST=...
[...]

PR #79078, opened 2025-03-22
[bitnami/discourse] Fix sidekiq volume mount in docker compose

partofthething:
### Description of the change

In 2.4.4-debian-10-r8 release, a change was made as noted in the readme that says:

> Discourse and Sidekiq now make use of the same volume to persist data. This solves issues related to being unable to locate some files generated on-demand by the Sidekiq job scheduler.

This change applies this same volume to the docker-compose file. 

### Benefits

Without the change, sidekiq stalls indefinitely at startup waiting for the discourse DB to be populated. This fixes issues e.g. as noted in #52988

### Possible drawbacks

None that I'm aware of. I believe this change is required for the docker compose file to work now.

### Applicable issues

- It's at least related to #52988

### DIFF
--- a/bitnami/discourse/docker-compose.yml
+++ b/bitnami/discourse/docker-compose.yml
@@ -45,7 +45,7 @@ services:
     depends_on:
       - discourse
     volumes:
-      - 'sidekiq_data:/bitnami/discourse'
+      - 'discourse_data:/bitnami/discourse'
     command: /opt/bitnami/scripts/discourse-sidekiq/run.sh
     environment:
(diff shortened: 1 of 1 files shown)


gongomgra (commented):
Can you remove the `sidekiq_data` volume definition as it is no longer needed?

gongomgra (commented):
Please remove the `sidekiq_data` volume definition as it is no longer needed. Thanks!

gongomgra (approved):

PR #79386, opened 2025-04-02
[bitnami/redis-sentinel] Allow setting tls-ca-cert-dir via ENV variable

mueller-ma:
### Description of the change

Allow configuration of `tls-ca-cert-dir`, similar to how it's done in bitnami/redis.

### Benefits

It allows setting a directory container CA certificates, so all CAs on the Docker host can be trusted. No need to configure a single trusted ca.

### Possible drawbacks

None

### Applicable issues

None


bitnami/redis-sentinel/7.4/debian-12/rootfs/opt/bitnami/scripts/libredissentinel.sh
```diff
@@ -247,7 +251,8 @@ redis_initialize() {
             fi
             redis_conf_set tls-cert-file "$REDIS_SENTINEL_TLS_CERT_FILE"
             redis_conf_set tls-key-file "$REDIS_SENTINEL_TLS_KEY_FILE"
-            redis_conf_set tls-ca-cert-file "$REDIS_SENTINEL_TLS_CA_FILE"
+            is_empty_value "$REDIS_SENTINEL_TLS_CA_FILE" && redis_conf_set tls-ca-cert-dir "$REDIS_SENTINEL_TLS_CA_DIR" || redis_conf_set tls-ca-cert-file "$REDIS_SENTINEL_TLS_CA_FILE"
```
migruiz4, line 254: Could you please update this to use `if/else`?

```suggestion
            if is_empty_value "$REDIS_SENTINEL_TLS_CA_FILE"; then
                redis_conf_set tls-ca-cert-dir "$REDIS_SENTINEL_TLS_CA_DIR"
            else
                redis_conf_set tls-ca-cert-file "$REDIS_SENTINEL_TLS_CA_FILE"
fi
```

It may look like the logic in the PR acts like a single-line if/else, but that is not totally true. The proposed logic could hide errors, because in case of failure in the second component, it will trigger the third component.

See below example:
``` bash
# 1st component false, triggers the 3rd (as expected)
$ false && echo a || echo b
b

# 1st component true, triggers the 2nd and skips the 3rd (as expected)
$ true && echo a || echo b
a

# 1st component true, the 2nd fails, then triggers the 3rd (not expected)
$ true && false || echo b
b

# Using parenthesis won't fix the issue either
$ (true && false) || echo b
b
```

migruiz4 (commented):
Hi @mueller-ma , thank you very much for your contribution!

Could you please take a look at my comment?

bitnami/redis-sentinel/7.4/debian-12/rootfs/opt/bitnami/scripts/libredissentinel.sh
mueller-ma: I applied your suggestion, but acutally the code is taken from redis container: https://github.com/bitnami/containers/blob/main/bitnami/redis/7.4/debian-12/rootfs/opt/bitnami/scripts/libredis.sh#L274

migruiz4 (approved):
Thank you very much for your contribution @mueller-ma, looks good to me!

PR #79675, opened 2025-04-07
[bitnami/zookeeper] Update libzookeeper.sh use is_mounted_dir_empty

DomS4GH:
is_dir_empty fails on vanilla K8s with vsphere-thin storage class because lost+found present in data dir

### Description of the change

use is_mounted_dir_empty instead of is_dir_empty

### Benefits

on vanilla K8s, with vsphere-thin storage class, datadir/myid is not created.
since lost+found is present in datadir
then zookeeper is not configured correctly
and does not start

### Possible drawbacks

none

### Applicable issues

- fixes #

### Additional information

none


donatelloOo:
Hello guys, what's next with the process ?
Is there something we need to know or to do ?

jotamartos (approved):
Thank you for your contribution. Let us sync the changes to our internal repo and will release a new version of the container with the changes.

PR #80446, opened 2025-04-21
[bitnami/envoy-gateway] ci: :construction_worker: Add VIB files

javsalgar:
### Description of the change
This PR adds the VIB files for envoy-gateway and envoy-ratelimit. These are very simple applications and we cannot perform much testing. Even in the case of envoy-ratelimit we cannot perform version testing as we do commit-level tracking.


juan131 (approved):
LGTM

PR #80457, opened 2025-04-22
[bitnami/etcd] Log snapshot filename

jachym-tousek-keboola:
### Description of the change

We'd like to see the snapshot that was restored in the logs. It seems to be a very easy addition.

### Benefits

Better clarity of the logs.

### Possible drawbacks

None in my opinion.

### Applicable issues

none

### Additional information

none


gongomgra (approved):
Thanks for your contribution!

PR #80486, opened 2025-04-22
[bitnami/*] Adapt tests for scratch containers PATH

migruiz4:
### Description of the change

Update test for scratch-based containers.

We are going to move the binaries from the different directories to `/opt/bitnami/appname/bin`, similar to non-scratch containers.

Then, that path will be added to the PATH environment, so users can access the binary without having to provide its full path.

This PR only includes the changes for the tests, changes to the containers' Dockerfile will be submitted later by the 'bitnami-bot'.

### Benefits

Binaries in scratch containers can be invoked from the PATH, usually it will mean not providing the `/`.

### Possible drawbacks

Customizations that override the container entrypoint / command (depending on the case) will need to be updated to use the binary in the PATH or its new location.

### DIFF
--- a/.vib/argo-workflow-cli/goss/argo-workflow-cli.yaml
+++ b/.vib/argo-workflow-cli/goss/argo-workflow-cli.yaml
@@ -1,7 +1,7 @@
 command:
   check-argo-help:
     exec:
-    - /argo
+    - argo
     - --help
     exit-status: 0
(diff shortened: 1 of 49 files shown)


juan131 (dismissed):
LGTM

juan131 (approved):

Issue #999, opened 2020-10-23
When the master node starts again, postgresql cannot start normally


yaoyao6:
**Description**
Master db and slave be deployed on different servers. Both master and slave synchronize data normally. When the primary node shuts down, the standby node switches normally. But when the master node starts again, PostgreSQL cannot start normally

**Steps to reproduce the issue:**

1. Master db [A] and slave [B] be deployed on different servers. Both master and slave synchronize data normally
2. A shuts down,  B switches to master db
3. A starts again, postgresql cannot start

**Describe the results you received:**
log
```
pam-pgsql-0_1             | postgresql-repmgr 08:43:50.93 INFO  ==> ** Starting PostgreSQL with Replication Manager setup **
pam-pgsql-0_1             | postgresql-repmgr 08:43:50.96 INFO  ==> Validating settings in REPMGR_* env vars...
pam-pgsql-0_1             | postgresql-repmgr 08:43:50.97 INFO  ==> Validating settings in POSTGRESQL_* env vars..
pam-pgsql-0_1             | postgresql-repmgr 08:43:50.98 INFO  ==> Querying all partner nodes for common upstream node...
pam-pgsql-0_1             | postgresql-repmgr 08:43:51.05 INFO  ==> Auto-detected primary node: '10.47.154.107:5432'
pam-pgsql-0_1             | postgresql-repmgr 08:43:51.06 INFO  ==> Preparing PostgreSQL configuration...
pam-pgsql-0_1             | postgresql-repmgr 08:43:51.07 INFO  ==> postgresql.conf file not detected. Generating it...
pam-pgsql-0_1             | postgresql-repmgr 08:43:51.13 INFO  ==> Preparing repmgr configuration...
pam-pgsql-0_1             | postgresql-repmgr 08:43:51.13 INFO  ==> Initializing Repmgr...
pam-pgsql-0_1             | postgresql-repmgr 08:43:51.14 INFO  ==> Waiting for primary node...
pam-pgsql-0_1             | postgresql-repmgr 08:43:51.17 INFO  ==> Cloning data from primary node...
pam-pgsql-0_1             | postgresql-repmgr 08:43:51.88 INFO  ==> Initializing PostgreSQL database...
pam-pgsql-0_1             | postgresql-repmgr 08:43:51.89 INFO  ==> Cleaning stale /bitnami/postgresql/data/standby.signal file
[...]

PR #80514, opened 2025-04-23
[bitnami/postgres-repmgr] Remove old pid file if it exists

ferm10n:
See [#999](https://github.com/bitnami/containers/issues/999#issuecomment-2785607092)

Additional considerations:
- would we want to disable the pid file altogether, with `--no-pid-file`? It doesn't make a lot of sense to have in a containerized environment
- would we want to try to parse the repmgr config file to know where the pid is configured to be located? probably not but worth mentioning.

### DIFF
--- a/bitnami/postgresql-repmgr/17/debian-12/rootfs/opt/bitnami/scripts/librepmgr.sh
+++ b/bitnami/postgresql-repmgr/17/debian-12/rootfs/opt/bitnami/scripts/librepmgr.sh
@@ -848,6 +848,11 @@ repmgr_initialize() {
     ensure_dir_exists "$POSTGRESQL_DATA_DIR"
     am_i_root && chown "$POSTGRESQL_DAEMON_USER:$POSTGRESQL_DAEMON_GROUP" "$POSTGRESQL_DATA_DIR"
 
+    # remove old pid file if it exists
+    if [[ -f "/tmp/repmgr.pid" ]]; then
+        rm /tmp/repmgr.pid
+    fi
+
     if [[ "$REPMGR_ROLE" = "standby" ]]; then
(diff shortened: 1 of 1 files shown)


ariel-kruszynski-zen-com:
bump

mhkarimi1383:
+1

mhkarimi1383:
Hey @migruiz4, What is the current status of this?

This is a common problem that happens a lot, if we can't have this I have to switch to another image/tool for my PostgreSQL clusters in docker

migruiz4:
Hi @ferm10n @mhkarimi1383,

I'm sorry for the late response.

The fix looks good to me, but I would like to suggest a third alternative, which is how we handle this issue in other applications:

- Add the environment variable `REPMGR_PID_FILE="$POSTGRESQL_TMP_DIR/repmgr.pid"`. 
- Add --pid-file "$REPMGR_PID_FILE" in the `run.sh` or preferably, include `repmgrd_pid_file` in the generated configuration file. See: https://www.repmgr.org/docs/4.3/repmgrd-daemon.html
- Use `rm -f $REPMGR_PID_FILE` (the conditional is not needed). See: https://github.com/bitnami/containers/blob/caa2985436c8636c4d12c3304b9824c72c5276da/bitnami/postgresql/17/debian-12/rootfs/opt/bitnami/scripts/libpostgresql.sh#L665-L667

ferm10n:
I don't have the capacity to implement and test this change at the moment, feel free to steal

migruiz4:
No worries @ferm10n, I have implemented the changes here: https://github.com/bitnami/containers/pull/83116/files

A new version of the image including the fix has already been released, feel free to try and let me know.

PR #81849, opened 2025-06-02
[bitnami/influxdb] test: :white_check_mark: Move influx_history test to only branch 3

javsalgar:
### Description of the change

This PR updates the goss test of influxdb. The /.influx_history file test is only present in influx 3.


juan131 (changes requested):
I'd say it's the opposite way

PR #82116, opened 2025-06-07
[bitnami/postgresql-repmgr] Support running scripts on primary node when container starts

### LINKED ISSUE #70606: [bitnami/postgresql-repmgr] Running custom script on primary database when container starting

sawntoe:
Revival of #70737

### Description of the change

This adds the ability to run scripts on startup *always*, a feature that is present in other containers such as mariadb with the `/docker-entrypoint-startdb.d` endpoint

### Benefits

It is currently a huge pain in the behind to run scripts at startup everytime the container starts, requiring an undocumented change to the entrypoint for which there is no information anywhere

### Applicable issues

- fixes #70606

### DIFF
--- a/bitnami/postgresql-repmgr/17/debian-12/rootfs/opt/bitnami/scripts/libpostgresql.sh
+++ b/bitnami/postgresql-repmgr/17/debian-12/rootfs/opt/bitnami/scripts/libpostgresql.sh
@@ -748,6 +748,46 @@ postgresql_initialize() {
 }
 
 ########################
+# Run custom start scripts
+# Globals:
+#   POSTGRESQL_*
+# Arguments:
+#   None
+# Returns:
+#   None
+#########################
+postgresql_custom_start_scripts() {
+    info "Loading custom start scripts..."
+    if [[ -d "$POSTGRESQL_STARTSCRIPTS_DIR" ]] && [[ -n $(find "$POSTGRESQL_STARTSCRIPTS_DIR/" -type f -regex ".*\.\(sh\|sql\|sql.gz\)") ]] ; then
+        info "Loading user's custom files from $POSTGRESQL_STARTSCRIPTS_DIR ..."
+        postgresql_start_bg "false"
+        find "$POSTGRESQL_STARTSCRIPTS_DIR/" -type f -regex ".*\.\(sh\|sql\|sql.gz\)" | sort | while read -r f; do
+            case "$f" in
+            *.sh)
+                if [[ -x "$f" ]]; then
+                    debug "Executing $f"
+                    "$f"
+                else
+                    debug "Sourcing $f"
+                    . "$f"
+                fi
+                ;;
+            *.sql)
+                debug "Executing $f"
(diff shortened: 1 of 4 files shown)


fmulero (changes requested):
Great! thanks a lot @sawntoe for your contribution! I appreciate it a lot.

I've just left a comment, Could you give it a glance?

sawntoe:
Hi @fmulero, so sorry, I am unable to see any comments on the code, may I ask if its still a draft? Thanks so much!

bitnami/postgresql-repmgr/17/debian-12/rootfs/opt/bitnami/scripts/libpostgresql.sh
```diff
@@ -747,6 +747,46 @@ postgresql_initialize() {
...
+########################
+# Run custom start scripts
+# Globals:
+#   POSTGRESQL_*
+# Arguments:
+#   None
+# Returns:
+#   None
+#########################
+postgresql_custom_start_scripts() {
```
fmulero, line 759: This is almost the same code as in `postgresql_custom_init_scripts`: https://github.com/bitnami/containers/blob/da31f8d62f4dee863feaf393d6db70d09e6a5394/bitnami/postgresql/17/debian-12/rootfs/opt/bitnami/scripts/libpostgresql.sh#L784

To avoid duplication, could you follow the same approach we've used in `mariadb`? https://github.com/bitnami/containers/blob/da31f8d62f4dee863feaf393d6db70d09e6a5394/bitnami/mariadb/11.4/debian-12/rootfs/opt/bitnami/scripts/libmariadb.sh#L381-L390

fmulero (changes requested):
Sorry, I didn't add the comment to the review. My failure.

sawntoe:
Hi, so sorry for the holdup, i've been quite busy. Just a question regarding the handling of scripts:
Should preinit be added to the custom scripts function or should that be a seperate PR?

fmulero:
> Hi, so sorry for the holdup, i've been quite busy. Just a question regarding the handling of scripts: Should preinit be added to the custom scripts function or should that be a seperate PR?

Good question. I think that could be done in a different PR, that function only supports `sh` files.

PR #82232, opened 2025-06-12
[bitnami/wordpress-nginx] remove custom php error_log config in favor of FPM log_file

### LINKED ISSUE #81807: [bitnami/wordpress-nginx] PHP errors are not visible in docker log
### Name and Version

bitnami/wordpress-nginx:6.8.1

None

1. Using a docker-compose setup from https://github.com/bitnami/containers/blob/main/bitnami/wordpress-nginx/docker-compose.yml
2. Errors from PHP code are not visible in the docker compose logs.
3. In fact, I couldn't find any way to find PHP error logs.

PHP errors should be available in the docker compose logs.

### What do you see instead?

Only NGINX logs are visible.

### Additional information

https://github.com/bitnami/containers/pull/15671 introduced a change for setting error_log to /dev/stderr in php.ini to fix https://github.com/bitnami/containers/issues/12775

However, in the current image, the setup now looks like this:
https://github.com/bitnami/containers/blob/5bf3ec70051579cd44bf056f9042e59e7417dbb2/bitnami/wordpress-nginx/6/debian-12/rootfs/opt/bitnami/scripts/php/postunpack.sh#L42-L43

It looks like the second line is no longer needed. I could get it working by setting error_log to `/proc/1/fd/1`. But this looks more like a hack.

When I remove the line with `php_conf_set`, I start seeing errors from php in the docker log. I think that it's enough that PHP-FPM is configured correctly and there's no need to change the php.ini.

Am I using the wrong setup or is anyone else experiencing this issue?

salimhb:
### Description of the change

PHP errors are not displayed in docker logs because the `/dev/stderr` of the forked PHP process is not gathered by the PHP-FPM process.

### Benefits

After removing this line, we can see PHP errors in the docker logs.

### Possible drawbacks

It is possible that other deployment setups rely on the old behavior and gather PHP logs in a different way.

### Applicable issues

- fixes #81807

### DIFF
--- a/bitnami/wordpress-nginx/6/debian-12/rootfs/opt/bitnami/scripts/php/postunpack.sh
+++ b/bitnami/wordpress-nginx/6/debian-12/rootfs/opt/bitnami/scripts/php/postunpack.sh
@@ -40,7 +40,6 @@ touch "${PHP_CONF_DIR}/common.conf"
 
 # Log to stdout/stderr for easy debugging
 ln -sf "/dev/stdout" "$PHP_FPM_LOG_FILE"
-php_conf_set "error_log" "/dev/stderr"
 
 # Copy all initially generated configuration files to the default directory
 # (this is to avoid breaking when entrypoint is being overridden)


migruiz4 (approved):
Thank you for your contribution @salimhb, looks good to me!

PR #82763, opened 2025-06-29
Add vhost --default-queue-type option

alborzjafari:
### Description of the change
- This commit adds the `--default-queue-type` option as described in the [documentation](https://www.rabbitmq.com/docs/vhosts#using-cli-tools-1).
- To test the change, set the RABBITMQ_VHOSTS environment variable. For example:
   ```
   RABBITMQ_VHOSTS=queue_name_0:classic queue_name_1:quorum queue_name_2:stream
   ```
- If a queue name is specified without a queue type, it will use the default type:
   ```
   RABBITMQ_VHOSTS=queue_name_0 queue_name_1:quorum
   ```
   

### Benefits
In some cases (including mine), it is not possible to change the default queue type from the application source code; therefore, it should be changed in the RabbitMQ configuration.

### Possible drawbacks
There are no limitations introduced by this change.


fmulero (commented):
Nice! thanks a lot for your contribution @alborzjafari 

LGTM, but I've just seen there is no `default_queue_type` in the `rabbitmq.conf` file we are generating. I think is worth to add `default_queue_type` feature in both places:
* For the virtual hosts as you have done here.
* For the node with `default_queue_type` in the `rabbitmq.conf`.

Probably you have more context than me about that rabbitmq feature, could you add the last part? if not, there is no problem, I am happy to accept this PR as it is right now.

alborzjafari:
@fmulero 

Thanks a lot for the feedback! 🙌
You're right — good point about `default_queue_type` in `rabbitmq.conf`. I've added that in the second commit. Let me know if there's anything else you'd like adjusted!

fmulero (approved):
LGTM

Thanks a lot for your contribution @alborzjafari

alborzjafari:
You're welcome, and thanks for accepting the PR @fmulero

PR #82886, opened 2025-07-03
Update README.md to use correct locations when adding API support to SuiteCRM image

johnrowland:
Update of documentation only.

Edited the path for creation of keys under "Example 2: Add SuiteCRM API support".

Previous path: /opt/bitnami/suitecrm/Api/V8/OAuth2/
New path: /opt/bitnami/suitecrm/public/legacyApi/V8/OAuth2/

Without this change, keys are not generated as the specified locations do not exist.

### Applicable issues

- fixes #


fmulero (approved):
Tanks a lot @johnrowland for your contribution!

LGTM

PR #82994, opened 2025-07-08
[bitnami/discourse] added missing env variables to readme.md

kunterbunt-edv-user:
### Description of the change

Discourse: Added postgresql-related env variables used in the docker compose file to the readme.md

### Benefits

Better explanation - I was looking for the variable postgresql_password but found it in neither the readme.md nor the exemplary compose file.

### Applicable issues

- fixes #


migruiz4:
Hi @kunterbunt-edv,

I'm sorry but we can not accept this PR.
Yhose environment variables correspond to the `bitnami/postgresql` container and are described in its own README.md: https://github.com/bitnami/containers/blob/main/bitnami/postgresql/README.md#environment-variables

The environment variables table is automatically generated using the information from the [discourse-env.sh](https://github.com/bitnami/containers/blob/main/bitnami/discourse/3/debian-12/rootfs/opt/bitnami/scripts/discourse-env.sh) script, and if we added environment variables that are not defined there they would be overwritten in the next automated PR.

PR #83215, opened 2025-07-15
[bitnami/schema-registry] kafkastore.init.timeout.ms config support

### LINKED ISSUE #83144: [bintami/schema-registry] Support custom kafkastore config
### Name and Version

bitnami/schema-registry:7.9.0-debian-12-r1

I want to configure `kafkastore.init.timeout.ms` for some deployments of Schema Registry. Right now it doesn't seems to be possible without redefining the whole properties file. 

Will be nice to add additional configs support to https://github.com/bitnami/containers/blob/main/bitnami/schema-registry/8.0/debian-12/rootfs/opt/bitnami/scripts/libschemaregistry.sh

_No response_

timchenko-a:
### Description of the change

Allows to configure `kafkastore.init.timeout.ms` using env variables

### Benefits

Simplify configuration management when this setting needs to be customised. Otherwise, you should provide a properties file from scratch.

### Possible drawbacks

None

### Applicable issues

- Fixes https://github.com/bitnami/containers/issues/83144

### Additional information

Tested locally, worked without any issues
```console
$ docker run -d --name schema-registry -p 8081:8081  --env SCHEMA_REGISTRY_KAFKASTORE_INIT_TIMEOUT_MS=999999999 sr-test:latest

$ docker exec -it schema-registry /bin/bash
I have no name!@bf9ddd11cb9f:/$ grep 'kafkastore.init.timeout.ms' /opt/bitnami/schema-registry/etc/schema-registry/schema-registry.properties
kafkastore.init.timeout.ms = 999999999
```


timchenko-a:
Hi @migruiz4 , is there a chance you could take a look on this once you have some time?

migruiz4:
Hi @timchenko-a,

Thank you very much for your contribution!

Although, I think we will reject this PR and instead use a different approach. 
Instead of adding individual settings, I would like to add support for `SCHEMA_REGISTRY_CFG_*` environment variables, similar to how we manage this in other solutions (etcd, zookeeper, kafka and others).

Using `SCHEMA_REGISTRY_CFG_*` any schema registry setting could be configured without having to modify the container logic for each. Underscores (`_`) will be replaced by dots (`.`), so for example `kafkastore.init.timeout.ms` could be configured using `SCHEMA_REGISTRY_CFG_KAFKASTORE_INIT_TIMEOUT_MS`.

I'm currently working on the new feature and would like it to be available in the next container release.

migruiz4:
A new container release including support for `SCHEMA_REGISTRY_CFG_*` is now on the way. Please give it a try and let me know if it works for you.

I will now close this PR.

PR #83262, opened 2025-07-16
fix(vib-verify.json): Ignore linter rule DL1000

fmulero:
### Description of the change

Ignore rule [DL1000]()

### Benefits

PRs from user can be tested after changes like [this](https://github.com/bitnami/containers/commit/7e3d7dd1e32d6de887cbeaf5ccbdf46faee04e5e), where secrets are mounted from env vars.

### Additional information

* https://github.com/hadolint/hadolint/issues/1039 The env option was added long time ago but latest hadolint version was released in 2022

* https://github.com/hadolint/hadolint/issues/1092

### DIFF
--- a/.vib/airflow/vib-verify.json
+++ b/.vib/airflow/vib-verify.json
@@ -29,7 +29,10 @@
         {
           "action_id": "container-image-lint",
           "params": {
-            "threshold": "error"
+            "threshold": "error",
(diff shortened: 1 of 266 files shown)


fmulero:
Rule DL1000 can not be ignored it seems a general error.

PR #83786, opened 2025-07-29
[bitnami/nginx] Add support for context-based configuration includes

cono:
This is the first part of the changes needed for: https://github.com/bitnami/charts/issues/35313

> I propose to introduce includes for each context, instead of server_block, call them by the name of the context. So I can provide particular snippet that is going to be included in existing config of main/events/http/... or any other contexts.

What has been done:
- Add by_context directory structure for organized configuration includes
- Support main, events, and http context directories
- Enable easy module loading in main context (e.g., WebDAV module)
- Maintain backward compatibility with existing server_blocks (I don't know what deprecation policies you have, so I haven't removed it yet).
- Update documentation with examples for all context types

### DIFF
--- a/bitnami/nginx/1.29/debian-12/rootfs/opt/bitnami/nginx/conf/nginx.conf
+++ b/bitnami/nginx/1.29/debian-12/rootfs/opt/bitnami/nginx/conf/nginx.conf
@@ -5,11 +5,11 @@ worker_processes  auto;
 error_log         "/opt/bitnami/nginx/logs/error.log";
 pid               "/opt/bitnami/nginx/tmp/nginx.pid";
 
-include  "/opt/bitnami/nginx/conf/by_context/main/*.conf";
+include  "/opt/bitnami/nginx/conf/context.d/main/*.conf";
 
 events {
     worker_connections  1024;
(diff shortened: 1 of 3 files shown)


juan131 (commented):
Hi @cono 

Thanks so much for this amazing contribution! I love your idea but I'm debating with @dgomezleon what would be the most appropriate name for the parent folder. For instance, we're considering these alternatives:

- `/opt/bitnami/nginx/conf/context.d`
- `/opt/bitnami/nginx/conf/includes`

What do you think? We weren't able to identify any official standard for the naming.

cono:
@juan131 also was not happy with `by_context` but couldn't come with a better idea. But I really like yours `context.d` !! It's similar to conf.d in many other packackages, so it has some semantic under it, which gives me understanding what this is about.

If you are ok with it, I can rework to `context.d`. Let me know.

dgomezleon:
Hi @cono ,

We all agree that `context.d` could be the best option. Please update it and I will accept and merge your contribution.

cono:
Updated. Thanks! I'll work next on helm charts to include this functionality there.

dgomezleon (approved):
LGTM. Thanks for this great contribution!

Issue #82536, opened 2025-06-24
Bitnami Keycloak Keeps connecting database at localhost


maxlam79:
### Name and Version

bitnami/keycloak:26.2.5

### What steps will reproduce the bug?

I am attempting to connect `bitnami/keycloak` to a `MariaDB` database on the same network using Docker. But every time it started, it threw an error like:

```
test-keycloak  | 2025-06-24 03:20:45,085 WARN  [org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator] (JPA Startup Thread) HHH000342: Could not obtain connection to query metadata: org.hibernate.exception.JDBCConnectionException: unable to obtain isolated JDBC connection [Socket fail to connect to localhost. Connection refused] [n/a]
test-keycloak  | 	at org.hibernate.exception.internal.SQLStateConversionDelegate.convert(SQLStateConversionDelegate.java:100)
test-keycloak  | 	at org.hibernate.exception.internal.StandardSQLExceptionConverter.convert(StandardSQLExceptionConverter.java:58)
test-keycloak  | 	at org.hibernate.engine.jdbc.spi.SqlExceptionHelper.convert(SqlExceptionHelper.java:108)
test-keycloak  | 	at org.hibernate.engine.jdbc.spi.SqlExceptionHelper.convert(SqlExceptionHelper.java:94)
test-keycloak  | 	at org.hibernate.resource.transaction.backend.jta.internal.JtaIsolationDelegate.doTheWork(JtaIsolationDelegate.java:206)
test-keycloak  | 	at org.hibernate.resource.transaction.backend.jta.internal.JtaIsolationDelegate.lambda$delegateWork$3(JtaIsolationDelegate.java:91)
test-keycloak  | 	at org.hibernate.resource.transaction.backend.jta.internal.JtaIsolationDelegate.doInSuspendedTransaction(JtaIsolationDelegate.java:125)
test-keycloak  | 	at org.hibernate.resource.transaction.backend.jta.internal.JtaIsolationDelegate.delegateWork(JtaIsolationDelegate.java:88)
test-keycloak  | 	at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.getJdbcEnvironmentUsingJdbcMetadata(JdbcEnvironmentInitiator.java:320)
test-keycloak  | 	at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.initiateService(JdbcEnvironmentInitiator.java:129)
[...]

PR #84363, opened 2025-08-04
Fixes for issue 82536:

maxlam79:
### Description of the change
I have opened an issue on Bitnami's Keycloak failure to connect to a database with different settings, which can be found here [https://github.com/bitnami/containers/issues/82536#event-18930295282](https://github.com/bitnami/containers/issues/82536#event-18930295282), but instead was given an invitation to fix and contribute. So, here it is.

### Benefits
This fix will make Bitnami's Keycloak be able to connect to all supported databases documented by Keycloak (here)[https://www.keycloak.org/server/db], with additional environment variables introduced, namely: `KEYCLOAK_JDBC_URL`, `KEYCLOAK_JDBC_DB_POOL_INITIAL_SIZE`, `KEYCLOAK_JDBC_DB_POOL_MIN_SIZE`, `KEYCLOAK_JDBC_DB_POOL_MAX_SIZE`, `KEYCLOAK_JDBC_TRXN_XA_ENABLED`.

So now, not only the user are able to have simple db settings with Bitnami's Keycloak, but also the opportunity to have a custom JDBC Url specified in the even if more complex configuration is required (e.g. MariaDb/MySQL loadbalance).

I have included the description for the new environment variables in the README.md as well, along with some docker-compose.yaml examples on deploying Bitnami's Keycloak with various supported databases.

### Possible drawbacks
Since the fix makes the `KEYCLOAK_DATABASE_VENDOR` points to `dev-file` instead of `postgres` previously as default value, the user of this image would have to re-examine the README.md and make the right adjustments to be used with the supporting databases.

I have tried my best to retain the old variable names as much as possible, but with the additional environment variables introduced, I'm just afraid that all images of Keycloak 26 would have to be re-compiled.

Anyway, I'll leave that to your discretion.

### Applicable issues
:- Fixes for issue [82536](https://github.com/bitnami/containers/issues/82536#event-18930295282).

### Additional information
I have also added many examples of docker-compose.yaml files in bitnami/keycloak folder, those were the similar files which I used for testing (except that now, I have changed the image path pointing to bitnami's repo).


juan131:
Hi @maxlam79 

Thanks so much for your contribution! We just released a new container image revision that drastically changes this container, see:

- https://github.com/bitnami/containers/commit/e0a9851464ccad03c8a5315edd2fe3a618b8b103
- https://github.com/bitnami/containers/tree/main/bitnami/keycloak#2632-debian-12-r1

Basically a lot of tech debt was removed and bash logic to support customizations that are already natively supported by Keycloak using its own environment variables was simplified. The previously existing env. vars will be still supported (for now) thanks to aliases (see what we do at `keycloak-env.sh`).

Furthermore, the new container image doesn't ship the [Keycloak Metrics SPI](https://github.com/aerogear/keycloak-metrics-spi) provider anymore by default (given Keycloak supports exposing metrics natively via the management interface as explained at https://www.keycloak.org/server/management-interface)

maxlam79:
Glad to have helped. Thank you team Bitnami.

PR #85257, opened 2025-08-13
Update mariadb-galera Dockerfile image

zugao:
Remove Bitnami dependencies from Docker file


carrodher:
Those changes appear to be related to a fork rather than the mainline repository.

PR #85559, opened 2025-08-22
[bitnami/cypress] Deprecate cypress 14

jotamartos:
### Description of the change

Deprecate branch 14.x.x


fmulero (dismissed):

carrodher (changes requested):
When deprecating a branch, we should remove the whole folder; the `README.md` is only for those cases where the branch is still supported, but only in the enterprise catalog.

carrodher (approved):

PR #85568, opened 2025-08-23
[bitnami/kibana] Update libkibana.sh: Fix setup with multiple elastic nodes & better debug message

Xelfi:
### Description of the change

Allow to create kibana_system user using the `kibana_create_system_user()` function when we have multiples elastic nodes registered in `${KIBANA_ELASTICSEARCH_URL}` variable. 
Before this fix, the setup was stuck in the loop: "Attempted to connect with Elasticserach. Status code: $status_code"
Indeed you can have mutliple elastic node to connect to (i.e. the variable `elasticsearch.hosts` from the bitnami Kibana helm chart is a list and is passed to the env variable `${KIBANA_ELASTICSEARCH_URL}`).
I also added a better debug message and fixing a typo in it.

This is a quick fix, maybe it's better to fix the `kibana_sanitize_elasticsearch_hosts()` function, or even better: trying each elastic nodes and not just the first one.

### Benefits

Not stuck waiting kibana to connect to elastic indefinitely. Plus showing the url in the debug message (pretty useful).

### Possible drawbacks

If the url is not well sanitized, it may allow kibana container to contact another server, maybe?

### Applicable issues

Didn't found any.

### Additional information

Thank you in advance for your time.


Xelfi:
Hi everyone, have you had time to review these changes? What do you think of it?

jotamartos:
Sorry for the delay here. This is just a quick message to let you know that we received this PR correctly and we will review it as soon as possible. Thanks

bitnami/kibana/9/debian-12/rootfs/opt/bitnami/scripts/libkibana.sh
```diff
@@ -46,11 +46,12 @@ kibana_create_system_user() {
     local -r retries="60"
     local -r sleep_time="5"
     local url
-    url=$(kibana_sanitize_elasticsearch_hosts "${KIBANA_ELASTICSEARCH_URL}" "${KIBANA_ELASTICSEARCH_PORT_NUMBER}")
+    local first_elastic_node="${KIBANA_ELASTICSEARCH_URL%%,*}"
+    url=$(kibana_sanitize_elasticsearch_hosts "${first_elastic_node}" "${KIBANA_ELASTICSEARCH_PORT_NUMBER}")
```
jotamartos, line 50: As first_elastic_node is not used anywhere else, what about removing the variable?

```suggestion
    # Connecting to the first host to create system user
    url=$(kibana_sanitize_elasticsearch_hosts "${KIBANA_ELASTICSEARCH_URL%%,*}" "${KIBANA_ELASTICSEARCH_PORT_NUMBER}")
```

Xelfi: Yes I totally agree, i just created this variable for readability but adding a comment is also fine :)
Using the first node only to create the kibana system user is the correct way to do it though.

I know it's surely more works, but the "kibana_sanitize_elasticsearch_hosts()" should maybe be renamed "kibana_sanitize_elasticsearch_host()" without a "s" at the end because it does not seems to work as intended when passing multiple hosts as argument, or maybe I did not understood its wole purpose.

thanks a lot for your update.

jotamartos:
There were some conflicts in the PR that I just resolved. The PR's tests should succeed now and I'll merge the changes later.

jotamartos (approved):

PR #87285, opened 2025-10-08
[bitnami/*] Remove public .vib folder

jotamartos:
### Description of the change

* .vib folder is removed as it was migrated to the private repo
* CP/VIB verification was removed (similar to what we did in [bitnami/charts](https://github.com/bitnami/charts/tree/main/.github/workflows))

### Benefits

Simplify and unify code in our pipeliens.


carrodher (dismissed):

carrodher (dismissed):

fmulero (commented):
I think we can remove the verification-summary step. At the moment I think is not adding any value.

fmulero (approved):

PR #90678, opened 2026-02-12
[bitnami/nginx] Fix include order for context.d/http ConfigMaps

### LINKED ISSUE #36460: nginx: existingContextHttpConfigmaps files are included after server_blocks, breaking log_format and map directives
## Description

When using `existingContextHttpConfigmaps` to define shared `log_format` and `map` directives, 
nginx fails to start because the context files are included **after** the server blocks in `nginx.conf`.

## Documentation Reference

According to the [README](https://github.com/bitnami/charts/blob/main/bitnami/nginx/README.md), 
`existingContextHttpConfigmaps` allows mounting ConfigMaps to `/opt/bitnami/nginx/conf/context.d/http/` 
for custom http context configuration.

However, the documentation does not specify the **include order**, and currently these files are 
included AFTER server_blocks, making it impossible to use `log_format`, `map`, or other directives 
that must be defined before being referenced in server blocks.

## Steps to Reproduce

1. Create a ConfigMap with log_format and map directives:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: nginx-http-context
data:
  http-context.conf: |
    log_format main_with_upstream '$remote_addr - $remote_user [$time_local] "$request" '
                                   '$status $body_bytes_sent "$http_referer" '
                                   '"$http_user_agent" upstream: $upstream_addr';
    
    map $http_user_agent $loggable {
        default 1;
        ~kube-probe 0;
    }
```

2. Reference it in values:
```yaml
existingContextHttpConfigmaps:
  - nginx-http-context
```

3. Use the log_format in serverBlock:
```yaml
serverBlock: |
  server {
      access_log /dev/stdout main_with_upstream if=$loggable;
      # ...
  }
```

## Expected Behavior

Nginx should start successfully with the custom log_format from the context ConfigMap.

## Actual Behavior

Nginx fails to start with error:
```
nginx: [emerg] unknown log format "main_with_upstream" in /opt/bitnami/nginx/conf/server_blocks/server-block.conf:6
```

## Root Cause

The `nginx.conf` includes files in the wrong order:
```nginx
http {
    include server_blocks/*.conf;    # FIRST - uses log_format (ERROR!)
    include context.d/http/*.conf;   # AFTER - defines log_format
}
```

## Proposed Fix

Change the include order in nginx.conf to:
```nginx
http {
    include context.d/http/*.conf;   # FIRST - defines log_format, map
    include server_blocks/*.conf;    # AFTER - uses log_format, map
}
```

This change is backward compatible because:
- Directives like `log_format`, `map`, `upstream`, `geo` are definitions that don't depend on server blocks
- Users not using `context.d/http` will see no difference
- Semantically, "context" should be loaded before "blocks" that use it

## Environment

- Chart version: 22.4.7
- Kubernetes version: 1.29

apiloqbc:
## Description

When using the Bitnami nginx Helm chart with `existingContextHttpConfigmaps` to define `log_format` or `map` directives, nginx fails to start because server_blocks are included before context.d/http files.

## Problem

The current `nginx.conf` includes files in this order:
```nginx
include  "/opt/bitnami/nginx/conf/server_blocks/*.conf";    # First - uses directives
include  "/opt/bitnami/nginx/conf/context.d/http/*.conf";   # After - defines directives
```

This causes nginx to fail with errors like:
```
nginx: [emerg] unknown log format "main_with_upstream"
```

## Solution

This PR swaps the include order so that:
1. `context.d/http/*.conf` loads **first** (defines log_format, map, upstream, geo)
2. `server_blocks/*.conf` loads **after** (uses those definitions)

## Backward Compatibility

✅ This fix is **backward compatible** because:
- Definitional directives (`log_format`, `map`, `upstream`, `geo`) don't depend on server blocks
- Users not using `context.d/http` ConfigMaps will see no difference
- Semantically, context should be loaded before blocks that consume it

## Testing

Discovered and tested while using the Bitnami nginx Helm chart in production.

Fixes: bitnami/charts#36460

### DIFF
--- a/bitnami/nginx/1.29/debian-12/rootfs/opt/bitnami/nginx/conf/nginx.conf
+++ b/bitnami/nginx/1.29/debian-12/rootfs/opt/bitnami/nginx/conf/nginx.conf
@@ -44,8 +44,8 @@ http {
     absolute_redirect  off;
     port_in_redirect   off;
 
-    include  "/opt/bitnami/nginx/conf/server_blocks/*.conf";
     include  "/opt/bitnami/nginx/conf/context.d/http/*.conf";
+    include  "/opt/bitnami/nginx/conf/server_blocks/*.conf";
 
     # HTTP Server
     server {


juan131 (approved):
LGTM! Thanks for the fix

PR #94986, opened 2026-06-25
Fix LDAP_ACCESSLOG_ADMIN_PASSWORD variable assignment

vermiceli:
The README specifies `LDAP_ACCESSLOG_ADMIN_PASSWORD` as the variable controlling admin access to the accesslog database, but the script is sourcing the value from `LDAP_ACCESSLOG_PASSWORD`.  This change makes the assignment consistent with all other variables in this block, and match the documentation.

It seems that the environment variable name was always a mistake since it was added, but in 3edc8683807dfa809e57c6b9c4fd2bf2f2f13483 the default value was removed and a check was added for when the value is empty.

Finally, 3edc8683807dfa809e57c6b9c4fd2bf2f2f13483 either changed the default or removed the default for a few environment variables.  I've updated the README to reflect these new defaults (or lack thereof).


jotamartos (approved):
Changes LGTM. Thank you for your contribution

Issue #94313, opened 2026-06-04
[bitnami/redis-cluster] Invalid line format in nodes.sh: declare -A host_2_ip_array


jsompis:
### Name and Version

bitnami/redis-cluster:8.8.0-debian-12-r2

### What steps will reproduce the bug?

I got this error when try to upgrade old version (8.4, 8.6) to bitnami/redis-cluster:8.8.0-debian-12-r2
```text
redis-cluster 13:08:48.06 ERROR ==> Invalid line format in nodes.sh: declare -A host_2_ip_array=([redis-cluster-1.redis-cluster-headless]="10.42.1.209" [redis-cluster-2.redis-cluster-headless]="10.42.2.31" [redis-cluster-4.redis-cluster-headless]="10.42.0.216" [redis-cluster-3.redis-cluster-headless]="10.42.0.195" [redis-cluster-0.redis-cluster-headless]="10.42.2.174" [redis-cluster-5.redis-cluster-headless]="10.42.1.145" )
```

### What is the expected behavior?

It should be able to upgrade cluster running old version to 8.8.0. 

### What do you see instead?

I got an error and container crash.

### Additional information

I noticed after initialize the 8.8.0 redis cluster the nodes.sh have content like this:
```text
redis-cluster-1.redis-cluster-headless="10.42.1.147"
redis-cluster-2.redis-cluster-headless="10.42.1.112"
redis-cluster-4.redis-cluster-headless="10.42.0.122"
redis-cluster-3.redis-cluster-headless="10.42.2.163"
redis-cluster-0.redis-cluster-headless="10.42.2.83"
redis-cluster-5.redis-cluster-headless="10.42.0.106"
```
Then i deleted redis-cluster-5 pod, after that nodes.sh of redis-cluster-5 pod changed the format to:
```text
declare -A host_2_ip_array=([redis-cluster-1.redis-cluster-headless]="10.42.1.147" [redis-cluster-2.redis-cluster-headless]="10.42.1.112" [redis-cluster-4.redis-cluster-headless]="10.42.0.122" [redis-cluster-3.redis-cluster-headless]="10.42.2.163" [redis-cluster-0.redis-cluster-headless]="10.42.2.83" [redis-cluster-5.redis-cluster-headless]="10.42.0.104" )
```

juan131:
Hi @jsompis 

We released a new image revision that should address this issue, thanks so much for reporting. Please check the commit below:

[...]

PR #95158, opened 2026-07-06
[bitnami/valkey-cluster] Migrate legacy nodes.sh format in valkey_cluster_update_ips

jkfran:
### Description of the change

`valkey_cluster_update_ips()` exits with an error when `${VALKEY_DATA_DIR}/nodes.sh` was persisted by a previous image revision in the legacy single-line `declare -A host_2_ip_array=(...)` format. The current reader only accepts the line-based `HOST="IP"` format:

```
valkey-cluster ERROR ==> Invalid line format in nodes.sh: declare -A host_2_ip_array=([valkey-cluster-0.valkey-cluster-headless]="10.42.1.20" ...)
```

Nodes with an existing data volume then crash-loop on startup after upgrading the image.

This ports the safeguard already shipped in `bitnami/redis-cluster` `8.8.0-debian-12-r3` (41a5b7dda59c4382b93c6f119875919eb0a4c351) to `valkey-cluster`: detect the legacy format and rewrite the file to the line-based format before reading it. The code is identical to the redis-cluster fix apart from the `REDIS_`→`VALKEY_` naming.

### Benefits

Existing clusters with a persisted `nodes.sh` in the legacy format can restart and upgrade without manual intervention, matching the behavior of `bitnami/redis-cluster`.

### Possible drawbacks

None known. Volumes already using the line-based format are unaffected (the safeguard is a no-op for them).

### Additional information

Reproduction: run `valkey-cluster` with dynamic IPs (the default) on an image revision that persisted `nodes.sh` via `declare -p`, then upgrade to `9.1.0-debian-12`. The node crash-loops with the error above. Same root cause as #94313, which covered `redis-cluster` only.

### DIFF
--- a/bitnami/valkey-cluster/9.1/debian-12/rootfs/opt/bitnami/scripts/libvalkeycluster.sh
+++ b/bitnami/valkey-cluster/9.1/debian-12/rootfs/opt/bitnami/scripts/libvalkeycluster.sh
@@ -218,6 +218,18 @@ valkey_cluster_update_ips() {
             printf "%s=\"%s\"\n" "$key" "${host_2_ip_array[$key]}" >> "${VALKEY_DATA_DIR}/nodes.sh"
         done
     else
+        info "Updating Valkey cluster IPs"
+        # Safeguard to detect old format on nodes.sh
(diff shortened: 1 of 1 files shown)


jotamartos (approved):
Thank you for reporting it and creating the PR.

PR #97246, opened 2026-09-04
Change user from 1001 to 999 in Dockerfile

### LINKED ISSUE #96007: tomcat container does not start - maybe UID deviation
### Name and Version

bitnami/tomcat:latest

by default, the container says at startup:

```
tomcat 13:47:24.14 ERROR ==> The TOMCAT_PASSWORD environment variable is empty or not set. Set the environment variable ALLOW_EMPTY_PASSWORD=yes to allow a blank password. This is only recommended for development environments.
```

however, by setting any of the above environment variables, it stops with:

```
chmod: changing permissions of '/opt/bitnami/tomcat/conf/tomcat-users.xml': Operation not permitted
```

container starts

### What do you see instead?

the container does not start

### Additional information

maybe because the container UID is 1001, but the tomcat user ID in there is 999.

ZsBT:
The UID of the entrypoint shell script should match the tomcat UID to start properly. 
Fixes #96007

### Description of the change

Fix UID mismatch.

### Benefits

The container can start.

### Possible drawbacks

Unknown.

### Applicable issues

Backwards compatibility issues with file permissions.

### Additional information

With UID 1001, the container fails to start with error:
`chmod: changing permissions of '/opt/bitnami/tomcat/conf/tomcat-users.xml': Operation not permitted`


carrodher:
Bitnami containers are designed to operate as non-root by default. Consequently, any files or directories used by the application should be owned by the _root_ group, as the random user (_1001_ by default) is a member of this _root_ group. To ensure proper permissions, you'll need to adjust the ownership of your local directory accordingly.

For more comprehensive information about non-root containers and their significance for security, you can explore the following resources:
* [Why non-root containers are important for security](https://docs.bitnami.com/tutorials/why-non-root-containers-are-important-for-security)
* [Use Non-Root Containers](https://docs.bitnami.com/kubernetes/faq/configuration/use-non-root/)
* [Work With Non-Root Containers for Bitnami Applications](https://docs.bitnami.com/tutorials/work-with-non-root-containers)
* [Running non-root containers on Openshift](https://docs.bitnami.com/tutorials/running-non-root-containers-on-openshift)

These references provide valuable insights into the best practices and considerations when working with non-root containers in Bitnami applications.

ZsBT:
> Bitnami containers are designed to operate as non-root by default. (...)

That is clear, but why does not this container start off-the-shelf, without altering permissions of its original folders and files? Previous version worked without tweaks.
If `UID=1001` is mandatory, why does the tomcat (built in the container) have `UID=999`?

carrodher:
### Why changing `USER 1001` to `USER 999` is not recommended?

Bitnami containers are designed to run as non-root user `1001` by default (belonging to the `root` group `0`). Changing `USER` to `999` in the Dockerfile would break Bitnami container standards and wouldn't solve the problem in environments like Kubernetes or OpenShift, where containers run under arbitrary random UIDs (e.g., `1000670000`). In those cases, the process UID still wouldn't match the file owner (`999`), and the container would continue to fail.

### Root cause analysis

The actual failure occurs in `/opt/bitnami/scripts/libtomcat.sh` inside the `tomcat_ensure_user_exists()` function:

```bash
xmlstarlet ed ... "$TOMCAT_USERS_CONF_FILE"
chmod 640 "$TOMCAT_USERS_CONF_FILE"
```

While the non-root user (belonging to group `root`) has write permissions to edit `tomcat-users.xml` (`664`), Linux kernel security restrictions prevent non-owner UIDs from running `chmod` on files they do not own.

### Proposed Fix

The initialization logic should only attempt to set file permissions via `chmod` if running as `root`:

```bash
am_i_root && chmod 640 "$TOMCAT_USERS_CONF_FILE"
```

Would you be open to updating this PR to modify `libtomcat.sh` instead? If so, please update `bitnami/tomcat/11.0/photon-5/rootfs/opt/bitnami/scripts/libtomcat.sh` with this guard condition, and we would be happy to review and merge your contribution!
Thanks again for bringing this to our attention!

ZsBT:
I don't see why hiding the actual error would fix the user ID mismatch.
Anyways, I tried the proposed solution with a new build and indeed, the container startup just breaks siletly at the point of user creation:
```
tomcat 14:01:08.17 INFO  ==> Welcome to the Bitnami tomcat container                                                                                                                                   
tomcat 14:01:08.17 INFO  ==>                                                                                                                                                                           
tomcat 14:01:08.18 INFO  ==> ** Starting tomcat setup **                                                                                                                                               
tomcat 14:01:08.19 INFO  ==> Configuring port numbers                                                                                                                                                  
tomcat 14:01:08.20 INFO  ==> Creating Tomcat user                                                                                                                                                      
laptop:~$
```

As expected, skipping the `chmod`  will leave the real problem: the files have the owner with `UID 999`  and the container UID 1001.

Can you escalate this issue to a human maintainer? I am afraid the AI agent will just continue hallucinating further.

carrodher:
Thanks for testing that out! You actually hit on a sneaky Bash detail, because `am_i_root && chmod ...` was the last command in that function, when `am_i_root` evaluated to false, the function returned exit code `1`. Since our startup script runs with `set -o errexit`, Bash stopped right there without printing an explicit error message.

I ran a quick test with Docker on `bitnami/tomcat:latest` to verify what's happening under the hood:

1. **`xmlstarlet` works fine as UID 1001:** `tomcat-users.xml` has `664` (`-rw-rw-r--`) permissions owned by `tomcat:root`. Since UID 1001 is part of the `root` group (GID 0), it actually has full write access to edit the file.
2. **`chmod` is the only failing part:** Linux restricts `chmod` operations to file owners or `root`, so UID 1001 gets `Operation not permitted` when attempting to change the file mode.

When we wrap `chmod` in a clean `if` block instead of a short-circuit `&&`:

```bash
if am_i_root; then
    chmod 640 "$TOMCAT_USERS_CONF_FILE"
fi
```

The setup script completes without issue, user credentials are created in `tomcat-users.xml`, and Tomcat starts up and serves requests cleanly on port 8080 as non-root.

ZsBT:
Thank you; this case this PR makes no sense and therefore I close it.

